PR #23, opened 2021-12-13
Add `wandb` logger

ashwinvaidya17:
# Changes

- Add `wandb` logger.

# Known issues

- The docstring in `AnomalibWandbLogger` module is copied from the pytorch lightning wandb logger module. This is because the anomalib one just adds two more functions.
- Another method was added to `visualizer_callback` so that we can log images to wandb in batch rather than sequentially at each `log` call. This is because if images are not provided in batches, they get logged at a different step. So we can't see all the images side by side on the dashboard. Batch operation ensures they are available side by side.

### DIFF
--- a/anomalib/core/callbacks/visualizer_callback.py
+++ b/anomalib/core/callbacks/visualizer_callback.py
@@ -11,6 +11,7 @@ from skimage.segmentation import mark_boundaries
 from anomalib import loggers
 from anomalib.core.model import AnomalyModule
 from anomalib.data.transforms import Denormalize
+from anomalib.loggers.wandb import AnomalibWandbLogger
 from anomalib.utils.post_process import compute_mask, superimpose_anomaly_map
 from anomalib.utils.visualizer import Visualizer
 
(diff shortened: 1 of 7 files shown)


djdameln (approved):

anomalib/core/callbacks/visualizer_callback.py
```diff
@@ -98,3 +99,15 @@ def on_test_batch_end(
             visualizer.add_image(image=vis_img, title="Segmentation Result")
             self._add_images(visualizer, pl_module, Path(filename))
             visualizer.close()
+
+    def on_test_end(self, trainer: pl.Trainer, pl_module: pl.LightningModule) -> None:
+        """Sync logs.
+
+        Some loggers prefer syncing the batch at the end rather than logging at each timestep.
```
samet-akcay, line 106: could you provide more description here. Not sure if this is descriptive enough

anomalib/loggers/wandb.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Note: When logging manually through `wandb.log` or `trainer.logger.experiment.log`,
+    make sure to use `commit=False` so the logging step does not increase.
+
+    See Also:
+        - `Tutorial <https://colab.research.google.com/drive/16d1uctGaw2y9KhGBlINNTsWpmlXdJwRW?usp=sharing>`__
+          on how to use W&B with PyTorch Lightning
+        - `W&B Documentation <https://docs.wandb.ai/integrations/lightning>`__
+
+    """
```
samet-akcay, line 76: Is this from PL? For instance, I'm seeing this.

```
    Example:
        >>> from pytorch_lightning.loggers import WandbLogger
        >>> from pytorch_lightning import Trainer
        >>> wandb_logger = WandbLogger()
        >>> trainer = Trainer(logger=wandb_logger)
```

samet-akcay (commented):
Thanks @ashwinvaidya17. Great addition!
Only a single comment regarding a docstring description.

anomalib/loggers/wandb.py
ashwinvaidya17: My bad. I missed the example (it should be `AnomalibWandbLogger`) but rest is from pytorch lightning. As the only extensions are `save` and `add_image`. I've written it in the docstring above but maybe I can make it more prominent.

samet-akcay (commented):
FYI

samet-akcay (approved):

PR #24, opened 2021-12-13
Address compatibility issues with OTE, that are caused by the legacy code.

samet-akcay:
# Changes

This PR fixes the compatibility issues with OTE SDK. After tagging this version as v0.1.2+, GitHub should be fully compatible

### DIFF
--- a/anomalib/core/model/inference.py
+++ b/anomalib/core/model/inference.py
@@ -137,7 +137,9 @@ class TorchInferencer(Inferencer):
         Returns:
             Tensor: pre-processed image.
         """
-        pre_processor = PreProcessor(config=self.config.transform, to_tensor=True)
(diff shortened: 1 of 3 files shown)


ashwinvaidya17 (approved):
Nice! The config stuff in preprocessor addresses an issue I had in benchmarking script. Maybe we should create a ticket to add inferencer tests.

djdameln:
Did you test this in SC yet?

samet-akcay:
> Did you test this in SC yet?

Tests on OTE Anomalib passes, but gives a warning message. I've shared the log on teams

samet-akcay:
> Nice! The config stuff in preprocessor addresses an issue I had in benchmarking script. Maybe we should create a ticket to add inferencer tests.

@ashwinvaidya17 , I agree, we should definitely have inference tests.

PR #25, opened 2021-12-13
Update OpenVINOInferencer

akorobeinikov:


samet-akcay (changes requested):
My feeling is that anomalib inferencer objects should stick to numpy input numpy output format. In ote-anomalib we could wrap these objects to return ote AnnotationSceneEntity. Perhaps, @ashwinvaidya17 could help here.



It's a good idea to use `OpenvinoAdapter` to load the model.

ashwinvaidya17:
Closing as this is addressed in https://github.com/openvinotoolkit/training_extensions/pull/782

PR #46, opened 2021-12-22
Fix Readmes

ashwinvaidya17:
# Changes

- Modify model specific READMEs
- Add metrics to Main README

# Possible Issues

I have explained the models based on my understanding. Might be wrong. It would be nice to look into that first.


README.md
```diff
@@ -22,26 +22,33 @@ ___
...
+
 ## Getting Started
+
 The repo is thoroughly tested based on the following configuration.
-*  Ubuntu 20.04
-*  NVIDIA GeForce RTX 3090
+
+- Ubuntu 20.04
+
+- NVIDIA GeForce RTX 3090
```
samet-akcay, line 43: Should we still keep this? We could either remove this bit, or add that we also support other platforms.

ashwinvaidya17: Not sure. Maybe we can keep a growing list in docs rather than over here and we can link that page here.

README.md
```diff
@@ -22,26 +22,33 @@ ___
...
 
 After downloading the Anomaly Library, extract the files and navigate to the extracted location.
 
 To perform an installation, run the following:
-```
+
+```bash
 yes | conda create -n anomalib python=3.8
 conda activate anomalib
 pip install -r requirements/base.txt
```
samet-akcay, line 54: Maybe we could also modify this bit. We could have an Installation section, one for pypi and one for editable install. Something like,

# Installation
## PyPI Install
```bash
pip install anomalib
```

## Local Install
For those who would like to edit,
```bash
git clone https://github.com/openvinotoolkit/anomalib.git
cd anomalib
pip install -e .
```

This is of course, quite raw an draft, we could explain it better

README.md
```diff
@@ -51,41 +58,108 @@ pip install -r requirements/base.txt
...
 ```
 
 Training a model on a specific dataset and category requires further configuration. Each model has its own configuration
 file, [`config.yaml`](https://gitlab-icv.inn.intel.com/algo_rnd_team/anomaly/-/blob/development/stfpm/anomalib/models/stfpm/config.yaml)
 , which contains data, model and training configurable parameters. To train a specific model on a specific dataset and
 category, the config file is to be provided:
-```
+
+```bash
 python tools/train.py --model_config_path <path/to/model/config.yaml>
```
samet-akcay, line 72: python tools/train.py --model_config_path anomalib/models/padim/config.yaml

samet-akcay: To show that we have some sample configs for each algo

README.md
```diff
@@ -51,41 +58,108 @@ pip install -r requirements/base.txt
...
 
 ## Inference
 
 Anomalib contains several tools that can be used to perform inference with a trained model. The script in [`tools/inference`](tools/inference.py) contains an example of how the inference tools can be used to generate a prediction for an input image.
 
 The following command can be used to run inference from the command line:
-```
+
+```bash
 python tools/inference.py --model_config_path <path/to/model/config.yaml> --weight_path <path/to/weight/file> --image_path <path/to/image>
```
samet-akcay, line 96: Maybe we could split the arguments to multiple lines since the reader might need to scroll to the right. Something like:

```bash
python tools/inference.py \
    --model_config_path <path/to/model/config.yaml> \
    --weight_path <path/to/weight/file> \
    --image_path <path/to/image>
```

README.md
```diff
@@ -51,41 +58,108 @@ pip install -r requirements/base.txt
...
 ```
+
 If the specified weight path points to a PyTorch Lightning checkpoint file (`.ckpt`), inference will run in PyTorch. If the path points to an ONNX graph (`.onnx`) or OpenVINO IR (`.bin` or `.xml`), inference will run in OpenVINO.
```
samet-akcay, line 100: And here, it would be nice if we sum these up to a single complete example such as,

```bash
# Train a model on MVTec leather category.
python train.py tools/train.py --model padim

# Infer the model on MVTec leather color category.
python tools/inference.py \
    --model_config_path  anomalib/models/stfpm/config.yaml\
    --weight_path results/padim/mvtec/leather/weights/model.ckpt \
    --image_path datasets/MVTec/leather/test/color/000.png
```

samet-akcay (changes requested):
Thanks @ashwinvaidya17! What an improvement! Only minor stuff on the main readme file. It's based on personal experience. When I first try a repo, I just copy paste the example code to see if it works. If not, I got annoyed easily :D

README.md
samet-akcay: Of course the virtualenv stuff could go under the local install. What do you think?

ashwinvaidya17: I agree

README.md
```diff
@@ -51,41 +58,108 @@ pip install -r requirements/base.txt
...
+| STFPM     | Wide ResNet-50     |   0.903   |   0.987   | **0.989** |   0.980   | **0.966** |   0.956   |   0.966   |   0.913   |   0.956   |   0.974   |   0.961   |   0.946   | **0.988** |   0.178    |   0.807    |   0.980   |
+| PatchCore | ResNet-18          |   0.935   |   0.979   |   0.843   |   0.989   |   0.934   |   0.925   |   0.956   |   0.923   |   0.942   |   0.967   |   0.913   |   0.931   |   0.924   |   0.958    |   0.881    |   0.954   |
+| PatchCore | Wide ResNet-50     |   0.955   |   0.988   |   0.903   |   0.990   |   0.957   |   0.936   |   0.972   |   0.950   |   0.968   |   0.974   |   0.960   |   0.948   |   0.917   |   0.969    |   0.913    |   0.976   |
+| PaDiM     | ResNet-18          |   0.968   |   0.984   |   0.918   | **0.994** |   0.934   |   0.947   |   0.983   |   0.965   |   0.984   |   0.978   |   0.970   |   0.957   |   0.978   |   0.988    |   0.968    |   0.979   |
+| **PaDiM** | **Wide ResNet-50** | **0.979** | **0.991** |   0.970   |   0.993   |   0.955   | **0.957** | **0.985** | **0.970** | **0.988** | **0.985** | **0.982** | **0.966** | **0.988** | **0.991**  | **0.976**  | **0.986** |
+
+### Balanced Accuracy Score
```
djdameln, line 133: We don't support the balanced accuracy score anymore since the Github migration. It would be better to leave it out of this list.

samet-akcay (approved):

PR #47, opened 2021-12-23
Added cflow algorithm

blakshma:


samet-akcay:
@blakshma pipeline fails because of missing `einops` module. I guess you need to add it to `requirements/base.txt`. You probably need to add `freia` as well

anomalib/models/cflow/backbone.py
```diff
@@ -0,0 +1,101 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+
+import math
+
+# FrEIA (https://github.com/VLL-HD/FrEIA/)
```
ashwinvaidya17, line 19: Do we need to mention this?

ashwinvaidya17 (approved):
Minor comment but rest looks good.

anomalib/models/cflow/backbone.py
blakshma: Not necessary. Removed the link.

blakshma:
Thank you for updating readme @samet-akcay

blakshma:
@samet-akcay the pipeline fails while testing load model for cflow. Is this something you have encountered before? Any leads?

samet-akcay:
> @samet-akcay the pipeline fails while testing load model for cflow. Is this something you have encountered before? Any leads?

@blakshma , you could look at the details [here](https://github.com/openvinotoolkit/anomalib/runs/4633629579?check_suite_focus=true), which shows why it failed. It's mainly becuase of CUDA memory issue while running C-Flow.

blakshma:
All tests are passed in both of my local system. Running into OOM error with pre-merge check.
image

samet-akcay:
It's quite likely that this is due to hardware differences between your machine and the mlops machine on which we run the CI. We'll investigate whether evaluation the metrics on cpu would remedy the issue.

anomalib/models/cflow/README.md
```diff
@@ -0,0 +1,3 @@
+# Real-Time  Unsupervised Anomaly Detection via Conditional Normalizing Flows
+
+This is the implementation of the [CW-AD](https://arxiv.org/pdf/2107.12571v1.pdf) paper.
```
ashwinvaidya17, line 3: Do you have sample results? It would be nice to include them here for consistency with other tasks. Here is an example https://github.com/openvinotoolkit/anomalib/blob/development/anomalib/models/stfpm/README.md
Otherwise, can you create a ticket for it to address this in a different PR.

samet-akcay (approved):

PR #48, opened 2021-12-23
Bugfix: fix random val/test split issue

djdameln:
This PR solves the problems with incorrect train/val/test splits. The seed from the config is now used when the splits are created. Also the `create_validation_set` argument of the `make_mvtec_dataset` now takes effect when set to True.

### DIFF
--- a/anomalib/data/mvtec.py
+++ b/anomalib/data/mvtec.py
@@ -199,10 +199,10 @@ def make_mvtec_dataset(
     samples.label_index = samples.label_index.astype(int)
 
     if create_validation_set:
-        samples = create_validation_set_from_test_set(samples)
+        samples = create_validation_set_from_test_set(samples, seed=seed)
 
     # Get the data frame for the split.
-    if split is not None and split in ["train", "test"]:
+    if split is not None and split in ["train", "test", "val"]:
         samples = samples[samples.split == split]
         samples = samples.reset_index(drop=True)
 
@@ -217,9 +217,11 @@ class MVTec(VisionDataset):
         root: Union[Path, str],
         category: str,
         pre_process: PreProcessor,
+        split: str,
         task: str = "segmentation",
-        is_train: bool = True,
         download: bool = False,
+        seed: int = 0,
+        create_validation_set: bool = False,
     ) -> None:
         """Mvtec Dataset class.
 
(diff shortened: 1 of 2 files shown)


ashwinvaidya17 (approved):
This was a good catch!

anomalib/data/mvtec.py
```diff
@@ -199,10 +199,10 @@ def make_mvtec_dataset(
     samples.label_index = samples.label_index.astype(int)
 
     if create_validation_set:
-        samples = create_validation_set_from_test_set(samples)
+        samples = create_validation_set_from_test_set(samples, seed=seed)
 
     # Get the data frame for the split.
-    if split is not None and split in ["train", "test"]:
+    if split is not None and split in ["train", "test", "val"]:
```
samet-akcay, line 205: `'train', 'val' or 'test'` as the logical flow

anomalib/data/mvtec.py
```diff
@@ -415,47 +424,45 @@ def __init__(
...
-            root=self.root,
-            category=self.category,
-            pre_process=self.pre_process,
-            is_train=False,
-            download=True,
-        )
+        self.train_data: Dataset
+        self.test_data: Dataset
+        if create_validation_set:
+            self.val_data: Dataset
```
samet-akcay, line 433: Is flake and pylint ok with this? When we don't define the variables in `__init__`, they usually complain.

anomalib/data/mvtec.py
```diff
@@ -464,8 +471,12 @@ def train_dataloader(self) -> DataLoader:
 
     def val_dataloader(self) -> DataLoader:
         """Get validation dataloader."""
-        return DataLoader(self.val_data, shuffle=False, batch_size=self.test_batch_size, num_workers=self.num_workers)
+        if self.create_validation_set:
+            return DataLoader(
+                self.val_data, shuffle=False, batch_size=self.test_batch_size, num_workers=self.num_workers
+            )
+        return DataLoader(self.test_data, shuffle=False, batch_size=self.test_batch_size, num_workers=self.num_workers)
```
samet-akcay, line 478: Maybe we could simplify this as follows:

```
dataset = self.val_data if self.create_validation_set else self.test_data
return DataLoader(dataset=dataset, shuffle=False, batch_size=self.test_batch_size, num_workers=self.num_workers)
```

samet-akcay:
Thanks @djdameln! Nice one! Only minor comments.

samet-akcay (approved):

PR #50, opened 2021-12-24
add distinction between image and pixel threshold in postprocessor

djdameln:
Quick PR to accommodate the latest changes to the OTE normalization branch.

### DIFF
--- a/anomalib/core/model/inference.py
+++ b/anomalib/core/model/inference.py
@@ -292,15 +292,13 @@ class OpenVINOInferencer(Inferencer):
             anomaly_map = np.log(anomaly_map)
             anomaly_map = (anomaly_map - meta_data["pixel_mean"]) / meta_data["pixel_std"]
             anomaly_map -= (meta_data["image_mean"] - meta_data["pixel_mean"]) / meta_data["pixel_std"]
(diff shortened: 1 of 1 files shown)


anomalib/core/model/inference.py
```diff
@@ -292,15 +292,13 @@ def post_process(self, predictions: np.ndarray, meta_data: Optional[Dict] = None
             anomaly_map = np.log(anomaly_map)
             anomaly_map = (anomaly_map - meta_data["pixel_mean"]) / meta_data["pixel_std"]
             anomaly_map -= (meta_data["image_mean"] - meta_data["pixel_mean"]) / meta_data["pixel_std"]
-            if "threshold" in meta_data.keys():
```
samet-akcay, line 295: any reason why you decided to remove the if statement?

djdameln: If the `threshold` key is missing from the meta data, I'd rather have the code fail with an error than return the anomaly map without finishing the conversion to normalized scores (I learned the hard way).

samet-akcay (approved):

ashwinvaidya17 (approved):

PR #53, opened 2021-12-29
Add min-max normalization

djdameln:
# Description
- Adds min-max normalization as an optional method for normalizing the anomaly scores and heatmaps.
- New `MinMaxNormalizationCallback` added
- Add option to change normalization method from the configurable parameters

## Changes

# Checklist:


anomalib/core/callbacks/min_max_normalization.py
```diff
@@ -0,0 +1,65 @@
+"""Anomaly Score Normalization Callback that uses min-max normalization."""
```
ashwinvaidya17, line 1: We will require a license header

anomalib/core/callbacks/__init__.py
```diff
@@ -51,17 +52,22 @@ def get_callbacks(config: Union[ListConfig, DictConfig]) -> List[Callback]:
...
+                raise NotImplementedError("Score Normalization is currently supported for PADIM and STFPM only.")
+        elif config.model.normalization_method == "min_max":
+            callbacks.append(MinMaxNormalizationCallback())
         else:
-            raise NotImplementedError("Score Normalization is currently supported for PADIM and STFPM only.")
+            raise ValueError(f"Normalization method not recognized: {config.model.normalization_method}")
 
     if not config.project.log_images_to == []:
-        callbacks.append(VisualizerCallback(inputs_are_normalized=config.model.normalize_scores))
+        callbacks.append(VisualizerCallback(inputs_are_normalized=config.model.normalization_method == "cdf"))
```
ashwinvaidya17, line 70: Should this be `normalization_method in ("cdf", "min_max")`

anomalib/models/padim/config.yaml
```diff
@@ -35,7 +35,7 @@ model:
 project:
   seed: 42
   path: ./results
-  log_images_to: []
+  log_images_to: [local]
```
ashwinvaidya17, line 38: Maybe we can keep this disabled.

ashwinvaidya17 (approved):
Thanks. Looks good. Only a few very minor comments.

anomalib/models/padim/config.yaml
```diff
@@ -3,7 +3,7 @@ dataset:
   format: mvtec
   path: ./datasets/MVTec
   url: ftp://guest:GU.205dldo@ftp.softronics.ch/mvtec_anomaly_detection/mvtec_anomaly_detection.tar.xz
-  category: leather
+  category: bottle
```
samet-akcay, line 6: Do we need this change? If yes, can we modify the other algos as well?

anomalib/core/callbacks/__init__.py
```diff
@@ -51,17 +52,22 @@ def get_callbacks(config: Union[ListConfig, DictConfig]) -> List[Callback]:
...
 
-    if "normalize_scores" in config.model.keys() and config.model.normalize_scores:
-        if config.model.name in ["padim", "stfpm"]:
-            if not config.optimization.nncf.apply:
-                callbacks.append(AnomalyScoreNormalizationCallback())
+    if "normalization_method" in config.model.keys() and not config.model.normalization_method == "none":
+        if config.model.normalization_method == "cdf":
+            if config.model.name in ["padim", "stfpm"]:
+                if not config.optimization.nncf.apply:
+                    callbacks.append(AnomalyScoreNormalizationCallback())
```
samet-akcay, line 59: I guess this is not super urgent, but we'll have to rename `AnomalyScoreNormalizationCallback` in the future since we now have multiple methods.

anomalib/core/model/anomaly_module.py
```diff
@@ -47,6 +52,7 @@ def __init__(self, params: Union[DictConfig, ListConfig]):
         self.pixel_threshold = AdaptiveThreshold(self.hparams.model.threshold.pixel_default)
 
         self.training_distribution = AnomalyScoreDistribution()
+        self.min_max = MinMax()
```
samet-akcay, line 55: Do we need these two at the same time?

djdameln: Not really, it depends on the normalization method that's being used. But I figured it doesn't hurt to have it either. Same as we also always have the `pixel_threshold` attribute, even when the task type is classification.

anomalib/core/callbacks/normalization.py
```diff
@@ -2,12 +2,12 @@
...
 
 import pytorch_lightning as pl
-import torch
 from pytorch_lightning import Callback, Trainer
 from pytorch_lightning.utilities.types import STEP_OUTPUT
-from torch.distributions import LogNormal, Normal
+from torch.distributions import LogNormal
 
 from anomalib.models import get_model
+from anomalib.utils.normalization.cdf_normalization import normalize, standardize
```
samet-akcay, line 10: do we need to explicitly name the module as `cdf_normalization`. It's already located in `normalisation` package. I think we could rename it as just `cdf`. in this case, it would be like `from anomalib.utils.normalization.cdf import normalize, standardize`. It would also be slightly less verbose.

anomalib/core/callbacks/min_max_normalization.py
```diff
@@ -0,0 +1,83 @@
...
+# See the License for the specific language governing permissions
+# and limitations under the License.
+
+from typing import Any, Dict
+
+import pytorch_lightning as pl
+from pytorch_lightning import Callback
+from pytorch_lightning.utilities.types import STEP_OUTPUT
+
+from anomalib.utils.normalization.min_max_normalization import normalize
```
samet-akcay, line 23: Similar to the comment below:
do we need to explicitly name the module as `min_max_normalization`. It's already located in `normalisation` package. I think we could rename it as just `min_max`. in this case, it would be like `from anomalib.utils.normalization.min_max import normalize, standardize`. It would also be slightly less verbose.

samet-akcay (commented):
Thanks a lot for the effort @djdameln. Such a great improvement! Only minor comments.

samet-akcay (approved):

Issue #56, opened 2022-01-04
Inference Bug - Too many variables to unpack


marvision-ai:
Hello, thank you for the great repo! 

**Describe the bug**
I cannot run inference. 

**To Reproduce**
` conda install openvino-ie4py-ubuntu20 -c intel`
` conda install pytorch==1.8.0 torchvision==0.9.0 torchaudio==0.8.0 cudatoolkit=11.1 -c pytorch -c conda-forge`

```bash
python tools/inference.py \
    --model_config_path anomalib/models/padim/config.yaml \
    --weight_path results/padim/mvtec/leather/weights/model.ckpt \
    --image_path datasets/MVTec/leather/test/color/000.png
```
- If applicable, add screenshots to help explain your problem.

``` bash
Traceback (most recent call last):
  File "tools/inference.py", line 90, in <module>
    infer()
  File "tools/inference.py", line 78, in infer
    output = inference.predict(image=args.image_path, superimpose=True)
  File "anomalib/anomalib/core/model/inference.py", line 90, in predict
    anomaly_map, pred_score = self.post_process(predictions, meta_data=meta_data)
ValueError: too many values to unpack (expected 2)

```

**Hardware and Software Configuration**
- OS: [Ubuntu, 20]
- NVIDIA Driver Version [470.57.02]
- CUDA Version [e.g. 11.4]
- CUDNN Version [e.g. v11.4.120]
- OpenVINO Version [Optional e.g. v2021.4.2]

**Additional context**
I also noticed that when I do inference from a ckpt file it still imports openvino- i expected this not to since it should run purely in pytorch.

samet-akcay:
Thanks for spotting this @marvision-ai! There have been some changes in the `OpenVinoInferencer`. These changes should also be applied to `TorchInferencer` in `post_process` method. Any thoughts @djdameln, @ashwinvaidya17?

ashwinvaidya17:
@samet-akcay I have fixed this in PR https://github.com/openvinotoolkit/anomalib/pull/17
https://github.com/openvinotoolkit/anomalib/blob/feature/ashwin/benchmarking_tools/anomalib/core/model/inference.py#L185

xxl007:
Thanks for the great repo. Reporting the same issue.

martin-inpel:
Thannk you @samet-akcay  and @ashwinvaidya17  for very fast turn around time! 

[...]

PR #60, opened 2022-01-06
Fix Inferencer

ashwinvaidya17:
# Comments

- I have refactored the code a bit and moved the inferencer to `deploy` folder. This is intended to comply with the upcomming refactor
- The place in base `Inferencer` where the meta data is retrieved from the model might require refactor.
- Also, I have changed the default category for `padim` and `stfpm` to bottle as that's the example used in the README.

# Changes

🐞 fix torch inferencer
🛠  Add metadata to inferencer
🧪 Add Inferencer tests

# Tests
image


anomalib/deploy/inference.py
```diff
@@ -105,6 +108,63 @@ def __call__(self, image: np.ndarray) -> np.ndarray:
...
+        meta_data: Union[Dict, DictConfig],
+    ) -> Tuple[Union[np.ndarray, Tensor], float]:
+        """Applies normalization and resizes the image.
+
+        Args:
+            anomaly_maps (Union[Tensor, np.ndarray]): Predicted raw anomaly map.
+            pred_scores (Union[Tensor, np.float32]): Predicted anomaly score
+            meta_data (Dict): Meta data. Post-processing step sometimes requires
+                additional meta data such as image shape. This variable comprises such info.
+                Defaults to {}.
```
samet-akcay, line 124: the default value is not defined above?

anomalib/deploy/inference.py
```diff
@@ -105,6 +108,63 @@ def __call__(self, image: np.ndarray) -> np.ndarray:
...
+            )
+            anomaly_maps = normalize_cdf(anomaly_maps, meta_data["pixel_threshold"])
+
+        # standardize image scores
+        if "image_mean" in meta_data.keys() and "image_std" in meta_data.keys():
+            pred_scores = standardize(pred_scores, meta_data["image_mean"], meta_data["image_std"])
+            pred_scores = normalize_cdf(pred_scores, meta_data["image_threshold"])
+
+        if isinstance(anomaly_maps, Tensor):
+            anomaly_maps = anomaly_maps.numpy()
```
samet-akcay, line 155: are we certain that the tensor is on cpu? Perhaps it would be safer to do `anomaly_maps.cpu().numpy()`

anomalib/deploy/inference.py
```diff
@@ -105,6 +108,63 @@ def __call__(self, image: np.ndarray) -> np.ndarray:
...
+
+        if "image_shape" in meta_data and anomaly_maps.shape != meta_data["image_shape"]:
+            anomaly_maps = cv2.resize(anomaly_maps, meta_data["image_shape"])
+
+        return anomaly_maps, float(pred_scores)
+
+    def _load_meta_data(self, path: Optional[Union[str, Path]] = None) -> Union[DictConfig, Dict]:
+        meta_data = {}
```
samet-akcay, line 163: How does this pass the tox pipeline without the docstring? Because of _ prefix?

anomalib/deploy/inference.py
```diff
@@ -113,15 +173,57 @@ class TorchInferencer(Inferencer):
...
         path (Union[str, Path]): Path to the model ckpt file.
+        meta_data_path (Union[str, Path], optional): Path to metadata file. If none, it tries to load the params
+                from the model state_dict. Defaults to None.
     """
 
-    def __init__(self, config: Union[DictConfig, ListConfig], path: Union[str, Path, AnomalyModule]):
+    def __init__(
+        self,
+        config: Union[DictConfig, ListConfig],
+        path: Union[str, Path, AnomalyModule],
```
samet-akcay, line 183: not sure if `path` is the best name here if it could be `AnomalyModule`

tools/inference.py
```diff
@@ -24,11 +24,7 @@
 import cv2
 
 from anomalib.config import get_configurable_parameters
-from anomalib.core.model.inference import (
-    Inferencer,
-    OpenVINOInferencer,
-    TorchInferencer,
-)
+from anomalib.deploy.inference import Inferencer, OpenVINOInferencer, TorchInferencer
```
samet-akcay, line 27: As per the suggestion in #56, should we import openvino only if the weights are ".onnx", ".bin", ".xml"? This means we would need to import not at the top of the module, but we could perhaps ignore such errors.

tools/inference.py
```diff
@@ -42,6 +38,7 @@ def get_args() -> Namespace:
     parser.add_argument("--weight_path", type=str, required=True, help="Path to a model weights")
     parser.add_argument("--image_path", type=str, required=True, help="Path to an image to infer.")
     parser.add_argument("--save_path", type=str, required=False, help="Path to save the output image.")
+    parser.add_argument("--meta_data", type=str, required=False, help="Path to JSON file containing the metadata.")
```
samet-akcay, line 41: should we use type `Path` here to automatically convert everything to pathlib.Path object to make it easier to work with the path?

tools/inference.py
```diff
@@ -60,10 +57,10 @@ def infer() -> None:
     extension = os.path.splitext(args.weight_path)[-1]
     inference: Inferencer
     if extension in (".ckpt"):
-        inference = TorchInferencer(config=config, path=args.weight_path)
```
samet-akcay, line 63: following the above argument, maybe we could do `from anomalib.deploy.inference TorchInferencer` here?

tools/inference.py
```diff
@@ -60,10 +57,10 @@ def infer() -> None:
...
+        inference = TorchInferencer(config=config, path=args.weight_path, meta_data_path=args.meta_data)
 
     elif extension in (".onnx", ".bin", ".xml"):
-        inference = OpenVINOInferencer(config=config, path=args.weight_path)
```
samet-akcay, line 66: same as above, `from anomalib.deploy.inference import OpenVINOInferencer`?

anomalib/deploy/inference.py
```diff
@@ -80,19 +80,22 @@ def predict(
...
         processed_image = self.pre_process(image)
         predictions = self.forward(processed_image)
-        anomaly_map, pred_score = self.post_process(predictions, meta_data=meta_data)
+        anomaly_map, _ = self.post_process(predictions, meta_data=meta_data)
 
         if superimpose is True:
             anomaly_map = superimpose_anomaly_map(anomaly_map, image)
 
-        return anomaly_map, pred_score
+        return anomaly_map
```
djdameln: I guess this would cause some issues for classification-type algorithms that produce image-level anomaly scores.

djdameln: In OTE, the OpenVINO task expects both the anomaly map and the predicted score to be returned. We'll need to make some changes there to make it compatible with this design.

anomalib/deploy/inference.py
```diff
@@ -105,6 +108,63 @@ def __call__(self, image: np.ndarray) -> np.ndarray:
...
+        # standardize image scores
+        if "image_mean" in meta_data.keys() and "image_std" in meta_data.keys():
+            pred_scores = standardize(pred_scores, meta_data["image_mean"], meta_data["image_std"])
+            pred_scores = normalize_cdf(pred_scores, meta_data["image_threshold"])
+
+        if isinstance(anomaly_maps, Tensor):
+            anomaly_maps = anomaly_maps.numpy()
+
+        if "image_shape" in meta_data and anomaly_maps.shape != meta_data["image_shape"]:
+            anomaly_maps = cv2.resize(anomaly_maps, meta_data["image_shape"])
```
djdameln: I feel that the Numpy conversion and resizing should be done outside of the _normalize function, because these operations are not related to the normalization.

anomalib/deploy/inference.py
```diff
@@ -21,7 +21,7 @@
 import cv2
```
djdameln: Can we rename this file to `inferencer.py`, to better make the distinction between `tools/inference.py`?

anomalib/deploy/inference.py
```diff
@@ -113,15 +173,57 @@ class TorchInferencer(Inferencer):
...
+        Args:
+            path (Optional[Union[str, Path]], optional): Path to metadata file. If none, it tries to load the params
+                from the model state_dict. Defaults to None.
+
+        Returns:
+            Dict: Dictionary containing the meta_data.
+        """
+        meta_data = {}
+        model_params = self.model.state_dict()
+        meta_data_params = {
```
djdameln: Since the model is already loaded here, you could just get the metadata from the model itself:
```
image_threshold = self.model.image_threshold
pixel_threshold = self.model.pixel_threshold
image_mean = self.model.training_distribution.image_mean
...
```

anomalib/deploy/inferencers/base.py
```diff
@@ -0,0 +1,172 @@
...
+                meta_data = getattr(self, "meta_data")
+            else:
+                meta_data = {}
+        if isinstance(image, (str, Path)):
+            image = read_image(image)
+        meta_data["image_shape"] = image.shape[:2]
+
+        processed_image = self.pre_process(image)
+        predictions = self.forward(processed_image)
+        anomaly_map, _ = self.post_process(predictions, meta_data=meta_data)
```
djdameln, line 88: I'm guessing this would cause some issues for classification-type algorithms. Maybe we can think of a design that supports both segmentation and classification algos.

anomalib/deploy/inferencers/base.py
```diff
@@ -0,0 +1,172 @@
...
+
+        if superimpose is True:
+            anomaly_map = superimpose_anomaly_map(anomaly_map, image)
+
+        return anomaly_map
```
djdameln, line 93: The OTE OpenVINO task expects that both the anomaly map and the predicted score are returned here. So we'll need to make some changes there to accommodate this design.

anomalib/deploy/inferencers/base.py
```diff
@@ -0,0 +1,172 @@
...
+        # standardize image scores
+        if "image_mean" in meta_data.keys() and "image_std" in meta_data.keys():
+            pred_scores = standardize(pred_scores, meta_data["image_mean"], meta_data["image_std"])
+            pred_scores = normalize_cdf(pred_scores, meta_data["image_threshold"])
+
+        if isinstance(anomaly_maps, Tensor):
+            anomaly_maps = anomaly_maps.cpu().numpy()
+
+        if "image_shape" in meta_data and anomaly_maps.shape != meta_data["image_shape"]:
+            anomaly_maps = cv2.resize(anomaly_maps, meta_data["image_shape"])
```
djdameln, line 152: I feel that the Numpy conversion and resizing should be taken outside of the normalize method, because both these operations are not related to normalization.

anomalib/deploy/inferencers/torch.py
```diff
@@ -0,0 +1,162 @@
...
+        Args:
+            path (Optional[Union[str, Path]], optional): Path to metadata file. If none, it tries to load the params
+                from the model state_dict. Defaults to None.
+
+        Returns:
+            Dict: Dictionary containing the meta_data.
+        """
+        meta_data: Union[DictConfig, Dict[str, Union[float, np.ndarray, Tensor]]] = {}
+        model_params = self.model.state_dict()
+        meta_data_params = {
```
djdameln, line 69: The model is already loaded here, so we can just get the metadata from the model itself:
```
image_threshold = self.model.image_threshold.value
pixel_threshold = self.model.pixel_threshold.value
image_mean = self.model.training_distribution.image_mean
...
```

anomalib/deploy/inferencers/base.py
```diff
@@ -0,0 +1,172 @@
...
+
+        return anomaly_maps, float(pred_scores)
+
+    def _load_meta_data(
```
djdameln, line 156: It's not fully clear to me how and where these metadata json files are generated

anomalib/core/callbacks/compress.py
```diff
@@ -30,14 +32,10 @@ def on_train_end(self, trainer, pl_module: LightningModule) -> None:  # pylint:
...
         onnx_path = os.path.join(self.dirpath, self.filename + ".onnx")
-        height, width = self.input_size
-        torch.onnx.export(
-            pl_module.model,
-            torch.zeros((1, 3, height, width)).to(pl_module.device),
-            onnx_path,
-            opset_version=11,
-            input_names=["input"],
-            output_names=["output"],
+        pl_module = cast(AnomalyModule, pl_module)
```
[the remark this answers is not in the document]
samet-akcay: In some of the callback methods we define `pl_module: AnomalyModule`, while in some others `pl_module: LightningModule`. If pylint and mypy are happy with `pl_module: AnomalyModule`, then we could maybe update the others as well. Therefore, we wouldn't need to cast it later in the code.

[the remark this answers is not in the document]
samet-akcay: But I guess this is a topic for another PR

anomalib/deploy/inferencers/base.py
```diff
@@ -0,0 +1,172 @@
...
+        """Pre-process."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def forward(self, image: Union[np.ndarray, Tensor]) -> Union[np.ndarray, Tensor]:
+        """Forward-Pass input to model."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def post_process(self, predictions: Union[np.ndarray, Tensor], meta_data: Optional[Dict]) -> np.ndarray:
```
samet-akcay, line 55: base inference task and torch inference tasks return only np.ndarray, while openvino inferencer returns Tuple[np.ndarray, float]. Design-wise, this is not correct, right? The signature should be the same. Is there a reason why only `np.ndarray` is returned here for anomaly_map, and not scores?

tools/inference.py
```diff
@@ -60,10 +59,14 @@ def infer() -> None:
     extension = os.path.splitext(args.weight_path)[-1]
```
samet-akcay, line 59: Now that the paths are of type `Path`, we could maybe use `extension = args.weight_path.suffix` and remove `import os` from the import at the top of the module.

anomalib/deploy/inferencers/base.py
ashwinvaidya17: We don't have metadata json files but this is just to add support for when we add them in the future. We will need this for the OpenVINO inferencer as the model does not contain the metadata.

djdameln: alright. So currently OpenVINO inference does not support normalization

ashwinvaidya17: Yes

samet-akcay (approved):

anomalib/core/callbacks/compress.py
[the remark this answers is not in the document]
ashwinvaidya17: Created Issue https://github.com/openvinotoolkit/anomalib/issues/65

PR #66, opened 2022-01-10
🚀 CI: Nightly Build

ashwinvaidya17:
# Changes

🚀 Add nightly build to CI
🚚 Move PR template to .github root


ashwinvaidya17:
Adding todo here for later reference.
- [ ] Parallelize runs over multiple GPUs

tests/nightly/core/callbacks/compress_callback/test_compress.py
```diff
@@ -1,12 +1,13 @@
...
 import tempfile
 
+import pytest
 import pytorch_lightning as pl
 from pytorch_lightning.callbacks.early_stopping import EarlyStopping
 
 from anomalib.config import get_configurable_parameters
 from anomalib.core.callbacks.compress import CompressModelCallback
-from tests.core.callbacks.compress_callback.dummy_lightning_model import (
+from tests.nightly.core.callbacks.compress_callback.dummy_lightning_model import (
```
samet-akcay, line 10: just to be inline with the recent refactor, can we move callbacks stuff to `tests.nightly.utils`?

tests/nightly/models/test_model_nightly.py
```diff
@@ -0,0 +1,130 @@
...
+from omegaconf import DictConfig, ListConfig, OmegaConf
+from pytorch_lightning import seed_everything
+
+from anomalib.utils.hpo.config import flatten_sweep_params
+from tests.helpers.dataset import get_dataset_path
+from tests.helpers.model import model_load_test, setup
+
+
+def get_model_nncf_cat() -> List:
+    model_support = [
```
samet-akcay, line 34: Could you please add a short docstring here?

tox.ini
```diff
@@ -72,10 +73,28 @@ deps =
     -r{toxinidir}/requirements/openvino.txt
 commands =
     coverage erase
-    coverage run --include=anomalib/* -m pytest -ra
-    coverage report -m --fail-under=90
+    coverage run --include=anomalib/* -m pytest tests/pre_merge/ -ra
+    coverage report -m --fail-under=85
```
samet-akcay, line 77: Should we create a new issue to increase the coverage and link it here? Or is it ok for pre-merge to have 85% coverage?

tests/nightly/models/test_model_nightly.py
```diff
@@ -0,0 +1,130 @@
+"""Test Models on all MVTec Categories."""
```
samet-akcay, line 1: would `test_model.py` be enough since the parent directory already specifies the type of the test?

tests/pre_merge/models/test_model_premerge.py
```diff
@@ -0,0 +1,57 @@
+"""Quick sanity check on models."""
```
samet-akcay, line 1: would test_model.py be enough since the parent directory already specifies the type of the test?

samet-akcay (commented):
What an improvement! Thanks a lot for organising this mess! Only minor comments...

tests/pre_merge/models/test_model_premerge.py
ashwinvaidya17: I get pytest discovery issues if I have keep the names of both the files as `test_model.py`. Hence, I have given both of them a unique (but redundant) name.

tox.ini
ashwinvaidya17: https://github.com/openvinotoolkit/anomalib/issues/94

tests/nightly/core/callbacks/compress_callback/test_compress.py
ashwinvaidya17: I think we should be refactoring all the tests to follow the directory structure of the new refactor design. I've created an issue for this https://github.com/openvinotoolkit/anomalib/issues/95

samet-akcay (approved):

djdameln:
I feel that maybe the callback tests should be executed at pre-merge instead of nightly. We could run them on a dummy dataset so that it does not take a long time to run.

PR #69, opened 2022-01-11
Fixed issue with model loading

blakshma:
Weight files specified as arg is not loaded and defaulted to file specified in the config file. Minor bug in config.py.


samet-akcay (approved):
good catch!

PR #70, opened 2022-01-12
✨ Add GANomaly

ashwinvaidya17:
# Description

Add GANomaly algorithm.

# Checklist:


samet-akcay:
Nice 🎁

anomalib/models/ganomaly/config.yaml
```diff
@@ -0,0 +1,120 @@
...
+  category: bottle
+  task: classification
+  label_format: None
+  tiling:
+    apply: false
+    tile_size: null
+    stride: null
+    remove_border_count: 0
+    use_random_tiling: False
+    random_tile_count: 16
```
samet-akcay, line 15: I think tiling should be on by default. The encoder and decoder network in this model is capable of handling 64x64 images max. So we need to use either 32x32 or 64x64 windows.

samet-akcay (approved):

samet-akcay (changes requested):
I'll probably approve after the tiling changes are added

anomalib/models/ganomaly/torch_model.py
```diff
@@ -0,0 +1,227 @@
...
+
+        assert input_size % 16 == 0, "Input size should be a multiple of 16"  # Why?
+
+        self.model = nn.Sequential()
+
+        self.model.add_module(
+            f"initial-conv-{num_input_channels}-{n_features}",
+            nn.Conv2d(num_input_channels, n_features, kernel_size=4, stride=2, padding=1, bias=False),
+        )
+        self.model.add_module(f"initial-relu-{n_features}", nn.LeakyReLU(0.2, inplace=True))
```
djdameln, line 54: I find this module a bit hard to read. Maybe instead of collecting all the layers in the `self.model` attribute, you could separate them into multiple modules, and then call them in the `forward` method? Something like
```
def __init__():
    self.initial_conv = nn.Conv2d(...)
    self.initial_relu = nn.LeakyReLU(...)
    ...

def forward(x):
    x = self.init_conv(x)
    x = self.init_relu(x)
    ...
```
I always like to be able to understand the data flow in a torch module by just looking at the forward mnethod.

anomalib/models/ganomaly/torch_model.py
```diff
@@ -0,0 +1,227 @@
...
+        exp_factor = int(math.log(input_size // 4, 2)) - 1
+        n_input_features = n_features * (2 ** exp_factor)
+
+        # CNN layer for latent vector input
+        self.model.add_module(
+            f"initial-{latent_vec_size}-{n_input_features}-convt",
+            nn.ConvTranspose2d(latent_vec_size, n_input_features, kernel_size=4, stride=1, padding=0, bias=False),
+        )
+        self.model.add_module(f"initial-{n_input_features}-batchnorm", nn.BatchNorm2d(n_input_features))
+        self.model.add_module(f"initial-{n_input_features}-relu", nn.ReLU(True))
```
djdameln, line 121: Same as in `Encoder`. I think it would be good to split `self.model` up into smaller chunks.

djdameln (commented):
Thanks, overall the PR looks good. Just some comments on the structure of the encoder and decoder modules, but this could be taste. Feel free to ignore them if you disagree.

anomalib/models/ganomaly/torch_model.py
```diff
@@ -0,0 +1,227 @@
+"""Torch models defining encoder, decoder, Generator and Discriminator."""
```
djdameln, line 1: Reminder: We need to check if this code contains any duplication from DCGAN repo. If it does, we should either acknowledge the authors (depending on the license) or write our own implementation.

samet-akcay (approved):
🔥🔥🔥

anomalib/models/ganomaly/torch_model.py
```diff
@@ -0,0 +1,254 @@
...
+        if add_final_conv_layer:
+            self.final_conv_layer = nn.Conv2d(
+                n_features, latent_vec_size, kernel_size=4, stride=1, padding=0, bias=False
+            )
+
+    def forward(self, input_tensor: Tensor):
+        """Return latent vectors."""
+
+        output = self.input_layers(input_tensor)
+        if len(self.extra_layers) > 0:
```
djdameln, line 95: I don't think this check is needed. If the sequential module is empty, calling it will just return the unaltered input.

anomalib/models/ganomaly/torch_model.py
```diff
@@ -0,0 +1,254 @@
...
+            f"final-{n_input_features}-{num_input_channels}-convt",
+            nn.ConvTranspose2d(n_input_features, num_input_channels, kernel_size=4, stride=2, padding=1, bias=False),
+        )
+        self.final_layers.add_module(f"final-{num_input_channels}-tanh", nn.Tanh())
+
+    def forward(self, input_tensor):
+        """Return generated image."""
+        output = self.latent_input(input_tensor)
+        output = self.inverse_pyramid(output)
+        if len(self.extra_layers) > 0:
```
djdameln, line 177: same as above

djdameln (commented):
Thanks for the changes. Just a minor comment

anomalib/models/ganomaly/torch_model.py
ashwinvaidya17: Thanks! I didn't know this. I have updated the PR.

djdameln (approved):

PR #77, opened 2022-01-18
Fix Mix Data type issue on inferencer

### LINKED ISSUE #67: 1 epoch
Is it meant to be only one epoch in training?
Your config files state 1 epoch, is that just as quick example?
I tried to train PADIM for 10 on mvtec leather and wood and the metrics stay the same anyway, so it seems nothing as gained by training more. 
Lightning module also warn that there is no optimizer so I guess train only finds the correct thresholds and that takes 1 epoch.

samet-akcay:
# Description

- As suggested by the community, this PR fixes the mixed data type issue reported in #67

```
patchcore inference:
File "d:\projects\anomalib\anomalib\utils\normalization\min_max.py", line 31, in normalize
normalized = ((targets - threshold) / (max_val - min_val)) + 0.5
TypeError: unsupported operand type(s) for -: 'numpy.ndarray' and 'Tensor'
```

- Fixes #67 

## Changes

# Checklist:

### DIFF
--- a/anomalib/deploy/inferencers/torch.py
+++ b/anomalib/deploy/inferencers/torch.py
@@ -138,7 +138,7 @@ class TorchInferencer(Inferencer):
             anomaly_map = predictions
             pred_score = anomaly_map.reshape(-1).max()
         else:
-            if isinstance(predictions[1],( Tensor)):
+            if isinstance(predictions[1], (Tensor)):
                 anomaly_map, pred_score = predictions
                 pred_score = pred_score.detach()
             else:
--- a/anomalib/deploy/inferencers/torch.py
+++ b/anomalib/deploy/inferencers/torch.py
@@ -138,6 +138,10 @@ class TorchInferencer(Inferencer):
             anomaly_map = predictions
             pred_score = anomaly_map.reshape(-1).max()
         else:
+            # NOTE: Patchcore `forward`` returns heatmap and score.
+            #   We need to add the following check to ensure the variables
+            #   are properly assigned. Without this check, the code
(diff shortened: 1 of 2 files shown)


anomalib/deploy/inferencers/torch.py
```diff
@@ -138,8 +138,12 @@ def post_process(
             anomaly_map = predictions
             pred_score = anomaly_map.reshape(-1).max()
         else:
-            anomaly_map, pred_score = predictions
-            pred_score = pred_score.detach().cpu().numpy()
+            if isinstance(predictions[1], (Tensor)):
+                anomaly_map, pred_score = predictions
+                pred_score = pred_score.detach()
```
ashwinvaidya17, line 143: I don't get this. If scores are of type `Tensor` then we should detach and convert to numpy right? And in the else statement, we shouldn't do anything.

samet-akcay: I haven't added anything myself. this is taken from #67 . Just created this PR so that you could address it, if you think it's not correct:)

ashwinvaidya17: Okay, thanks for creating this. I'll look into it.

samet-akcay: The fix is for patchcore, which returns both anomaly map and score after `forward`. Without this check, pred_score is converted to numpy, but not anomaly_map. That's why we see type mismatch error in normalize function.

I think this would be an enough fix for now. We'll use PL predict stuff anyway for torch inference.

djdameln (approved):

ashwinvaidya17 (approved):

Issue #74, opened 2022-01-17
Diverged metrics in PatchCore: dropping from 0.99 to 0.44 and 0.03 is rather critical?


samet-akcay:
diverged metrics: dropping from 0.99 to 0.44 and 0.03 is rather critical?
log images: nice :) 
Also padim dropped in performance, but not as crazy.
Here's a patchcore result of "good" parts: 
![008](https://user-images.githubusercontent.com/30144796/148973398-5e317afa-b682-4c44-9131-448fb19a64f2.png)

this is padim:
DATALOADER:0 TEST RESULTS
{'image_AUROC': 0.7589669823646545,
 'image_F1': 0.8787878751754761,
 'pixel_AUROC': 0.9781586527824402,
 'pixel_F1': 0.22379672527313232}
![008padim](https://user-images.githubusercontent.com/30144796/148974485-86b96972-a13f-4de2-b977-2036ea9e6f5e.png)

_Originally posted by @sequoiagrove in https://github.com/openvinotoolkit/anomalib/issues/67#issuecomment-1010091154_

samet-akcay:
@sequoiagrove, @blakshma worked on the patchcore results, where he managed to improve the performance to the following:

```
DATALOADER:0 TEST RESULTS
{'image_AUROC': 0.9524492025375366,
 'image_F1': 0.9551020860671997,
 'pixel_AUROC': 0.9894225597381592,
 'pixel_F1': 0.3562552034854889}
```

To reproduce the numbers you could use [this branch](https://github.com/openvinotoolkit/anomalib/tree/fix/barath/patchcore), to be merged to development soon after [this PR](https://github.com/openvinotoolkit/anomalib/pull/73).

Here are the qualitative results for the ones you shared above. (screw/test/good/009/png)
![image](https://user-images.githubusercontent.com/17574420/149715268-ff32518b-87d1-44cb-a900-e832138f831d.png)

sequoiagrove:
I get those numbers too now. but I wonder of the computation of the numbers is also wrong, because when I look through the results images, it only detects about 1/4 to 1/2 of the defects in the different categories. The pixel F1 = 0.35 seems to decribe best the actual performance. but I know that it is low due to correctly detected defects dont have to overlap perfectly to still be a good result.

sequoiagrove:
pixel auroc of 0.97 sound like it is good performance, but looking at the masks it is really not useful in a real system, and the auc is a bad metric for quantifying performance.
[...]

PR #80, opened 2022-01-18
Fixed issue with k_greedy method

blakshma:
#74 
Fixed issue with k_greedy method for core-set sampling
Removed nearest_neighbor and sparse_projection methods


anomalib/core/model/k_center_greedy.py
```diff
@@ -32,31 +30,19 @@ class KCenterGreedy:
         torch.Size([219, 1536])
     """
 
-    def __init__(self, model: SparseRandomProjection, embedding: Tensor, sampling_ratio: float) -> None:
-        self.model = model
+    def __init__(self, embedding: Tensor, sampling_ratio: float) -> None:
         self.embedding = embedding
         self.coreset_size = int(embedding.shape[0] * sampling_ratio)
+        self.model = random_projection.SparseRandomProjection(eps=0.9)
```
samet-akcay, line 36: The reason for using Torch SparseRandomProjection implementation was to be able to export the model to OpenVINO / ONNX. I was wondering if it is possible to do the same thing with sklearn SparseRandomProjection.

blakshma: @samet-akcay, I have verified that the original Torch implementation produces same result as the sklearn one. If the conversion is not possible with sklearn, it might be easier to switch back to the Torch implementation. I made this change just because I thought the implementation is cleaner with less code.

samet-akcay: good to hear, thanks!

samet-akcay: To convert to ONNX/OpenVINO and to make the computation faster on GPU, we could then use Torch implementation. We could move `SparseRandomProjection` to `anomalib.core.model`, which would de-clutter PatchCore implementation itself. We could discuss this further

djdameln (approved):

ashwinvaidya17 (commented):
Thanks. I only have one comment.

anomalib/core/model/k_center_greedy.py
```diff
@@ -109,23 +91,23 @@ def select_coreset_idxs(self, selected_idxs: Optional[List[int]] = None) -> List
...
         if self.embedding.ndim == 2:
+            self.model.fit(self.embedding)
             self.features = self.model.transform(self.embedding)
             self.reset_distances()
         else:
             self.features = self.embedding.reshape(self.embedding.shape[0], -1)
             self.update_distances(cluster_centers=selected_idxs)
 
         selected_coreset_idxs: List[int] = []
+        idx = int(torch.randint(high=self.n_observations, size=(1,)).item())
```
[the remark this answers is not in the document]
blakshma: Thanks for the review @ashwinvaidya17 .

This is assigned for the update distance to be called. Distance gets calculated/updated before idx gets updated again. If we have to remove this we will need a conditional statement in get_idx that gets called every time.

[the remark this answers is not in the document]
ashwinvaidya17: Ah right! I missed that.

ashwinvaidya17 (approved):

PR #84, opened 2022-01-25
Draft: CI: AUPRO metric implementation

blakshma:
Add AUPRO metric for improved segmentation accuracy measurement

### DIFF
--- a/anomalib/core/metrics/__init__.py
+++ b/anomalib/core/metrics/__init__.py
@@ -2,7 +2,8 @@
 from .adaptive_threshold import AdaptiveThreshold
 from .anomaly_score_distribution import AnomalyScoreDistribution
 from .auroc import AUROC
+from .aupro import AUPRO
 from .min_max import MinMax
 from .optimal_f1 import OptimalF1
 
-__all__ = ["AUROC", "OptimalF1", "AdaptiveThreshold", "AnomalyScoreDistribution", "MinMax"]
+__all__ = ["AUROC", "AUPRO", "OptimalF1", "AdaptiveThreshold", "AnomalyScoreDistribution", "MinMax"]
--- a/anomalib/core/metrics/aupro.py
+++ b/anomalib/core/metrics/aupro.py
@@ -1,0 +1,97 @@
+"""Implementation of AUPRO score based on TorchMetrics."""
+from typing import List, Optional, Callable, Any
+
+import torch
+from torchmetrics import Metric
+from torchmetrics.utilities.data import dim_zero_cat
+from skimage.measure import label, regionprops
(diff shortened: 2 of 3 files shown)


blakshma:
@samet-akcay @djdameln  @ashwinvaidya17 
This code is not complete. I have opened this thread to discuss AUROC implementation details. I have two opens:
- Do we need AUROC and F1 for segmentation as it does not add much value? Can we replace it with just AUPRO metric? 
- The original repo (from which this code was borrowed) contains two skimage functions - label, regionprops. Should we implement torch version of these methods to make this metric torch compatible?

samet-akcay:
@blakshma, looks like @djdameln worked on this implementation before. He will see if he could find the torch implementation

djdameln:
> @samet-akcay @djdameln @ashwinvaidya17 This code is not complete. I have opened this thread to discuss AUPRO implementation details. I have two opens:
> 
>     * Do we need AUROC and F1 for segmentation as it does not add much value? Can we replace it with just AUPRO metric?

The main reason why we still report F1 score for segmentation is that we need a threshold dependent metric to show how the models would perform in a real-world setting, where we are interested in the accuracy of the predictions made using a specific threshold. I agree that F1 might not be the best metric to use here, due to the class imbalance (similar for AUROC).

The PRO metric could be a useful addition, but it also has some limitations. My main concern is that it only cares about recall and ignores precision. The PRO score can therefore always be artificially increased by just picking a lower threshold. Reporting AUPRO instead of PRO addresses this, but AUPRO is again a threshold-independent metric, and therefore not a good indication of how the model would perform in a real-world anomaly detection setting.

Because of these considerations I would not be in favor of getting rid of F1 for segmentation at this moment. Though I agree that we should be looking for alternatives. Pixel AUROC is still widely used in literature, so I feel we should keep this to allow verifying our results against other implementations.

>     * The original repo (from which this code was borrowed) contains two skimage functions - label, regionprops. Should we implement torch version of these methods to make this metric torch compatible?

As Samet pointed out, I previously implemented a pytorch-only version of the PRO metric. Unfortunately it seems to have been lost during the Gitlab -> Github migration. I still vaguely remember how it was implemented though, so I could try to re-implement the metric and add it to your PR.

blakshma:
@djdameln I agree with your line of thought and thanks for taking up torch implementation. :-)

djdameln:
I've made a separate PR for the pytorch implementation of PRO and AUPRO: https://github.com/openvinotoolkit/anomalib/pull/96/
Once that one is merged we could close this.

samet-akcay:
Closing this. We could continue to discuss the metric in https://github.com/openvinotoolkit/anomalib/pull/96

Issue #72, opened 2022-01-14
Support python 3.7


VinhLoiIT:
**Is your feature request related to a problem? Please describe.**
I mainly use `anomalib` on Google Colab but it only supports python=3.7 at the moment. There is an error when running the training command:
`python -m tools.train --model_config_path path/to/config.yaml`

The error was:
```
Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/content/anomalib/tools/train.py", line 66, in <module>
    train()
  File "/content/anomalib/tools/train.py", line 55, in train
    model = get_model(config)
  File "/content/anomalib/anomalib/models/__init__.py", line 61, in get_model
    module = import_module(f"anomalib.models.{config.model.name}.model")
  File "/usr/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/content/anomalib/anomalib/models/stfpm/__init__.py", line 17, in <module>
    from .model import (  # noqa # pylint: disable=unused-import
  File "/content/anomalib/anomalib/models/stfpm/model.py", line 31, in <module>
    from anomalib.data.tiler import Tiler
  File "/content/anomalib/anomalib/data/tiler.py", line 19, in <module>
    from typing import Optional, Sequence, SupportsIndex, Tuple, Union
ImportError: cannot import name 'SupportsIndex' from 'typing' (/usr/lib/python3.7/typing.py)
```

**Suggestion**
[...]

PR #90, opened 2022-01-28
Reorder AUC and Python version 3.7

alexriedel1:
https://github.com/openvinotoolkit/anomalib/blob/f1c8a6b553b82a8974ccc9fd27aa5f9b939d5917/anomalib/core/metrics/auroc.py#L17

throws a `ValueError`: `The 'x' tensor is neither increasing or decreasing. Try setting the reorder argument to 'True'.`
It happens for large datasets I guess with only little variance in some validation datapoints?

I set `reorder=True` to resolve the problem.

Also as the development branch is compatible to Python 3.7, it should also be possible to install the package und 3.7
https://github.com/openvinotoolkit/anomalib/issues/72

### DIFF
--- a/anomalib/core/metrics/auroc.py
+++ b/anomalib/core/metrics/auroc.py
@@ -14,4 +14,4 @@ class AUROC(ROC):
             Value of the AUROC metric
         """
         fpr, tpr, _thresholds = super().compute()
-        return auc(fpr, tpr)
+        return auc(fpr, tpr, reorder=True)
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,7 @@ setup(
     license="Copyright (c) Intel - All Rights Reserved. "
     'Licensed under the Apache License, Version 2.0 (the "License")'
     "See LICENSE file for more details.",
-    python_requires=">=3.8",
+    python_requires=">=3.7",
     packages=find_packages("."),
     install_requires=INSTALL_REQUIRES,
     extras_require=EXTRAS_REQUIRE,


samet-akcay:
Hi @alexriedel1, thank you for your contribution to the repo! I've just triggered the Pre-merge checks, which is currently running.

djdameln:
@alexriedel1 Thanks for your PR! I agree that ideally the AUC should be computed with `reorder=True`. However, calling `auc` with this parameter currently leads to an underestimation of the model's performance. The reason for this is that pytorch does not use stable sorting (see https://stackoverflow.com/questions/60366033/torch-sort-and-argsort-sorting-randomly-in-case-of-same-element and https://github.com/pytorch/pytorch/issues/28871). This means that equal elements in the `fpr` array will lose their original ordering, which can lead to unpredictable behavior in the AUC computation.

It seems that stable sorting has been added in pytorch 1.9.0 as an option to `sort`, but for technical reasons we are currently bound to 1.8.1. Besides, even in torch 1.9.0, stable sorting is disabled by default, so just updating pytorch would not solve the problem by itself.

As a temporary solution we could apply the sorting ourselves, using a stable sorting algorithm, inside the `compute` method (before computing the AUC).

djdameln:
@alexriedel1 I created [a new PR](https://github.com/openvinotoolkit/anomalib/pull/91) that should address your problem, but maintain consistent results when the `fpr` is already ordered correctly. Will be closing this one.

PR #102, opened 2022-02-15
Add/Update Callbacks

samet-akcay:
# Description

- This design modifies the callbacks to be compatible with the upcoming PL-CLI

- Fixes # (issue)

## Changes

## Checklist

### DIFF
--- a/anomalib/utils/callbacks/save_to_csv.py
+++ b/anomalib/utils/callbacks/save_to_csv.py
@@ -32,11 +32,11 @@ class SaveToCSVCallback(Callback):
     def __init__(self):
         """Callback to save metrics to CSV."""
 
-    def on_test_epoch_end(self, _trainer: Trainer, pl_module: AnomalyModule) -> None:
+    def on_test_epoch_end(self, trainer: Trainer, pl_module: AnomalyModule) -> None:
         """Save Results at the end of testing.
 
         Args:
-            _trainer (Trainer): Pytorch lightning trainer object (unused)
+            trainer (Trainer): Pytorch lightning trainer object (unused)
             pl_module (LightningModule): Lightning modules derived from BaseAnomalyLightning object.
         """
         results = pl_module.results
@@ -48,4 +48,8 @@ class SaveToCSVCallback(Callback):
                 "wrong_prediction": np.logical_xor(results.true_labels, results.pred_labels).astype(int),
             }
         )
-        data_frame.to_csv(Path(pl_module.hparams.project.path) / "results.csv")
+
+        if trainer.log_dir is not None:
+            data_frame.to_csv(Path(trainer.log_dir) / "results.csv")
+        else:
+            raise ValueError("trainer.log_dir does not exist to save the results.")
--- a/anomalib/utils/callbacks/tiler.py
+++ b/anomalib/utils/callbacks/tiler.py
@@ -1,0 +1,83 @@
+"""Tiler Callback."""
+
+# Copyright (C) 2020 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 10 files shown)


anomalib/utils/callbacks/visualizer_callback.py
```diff
@@ -65,27 +68,30 @@ def _add_images(
...
-                        f"Requested {log_to} logging but logger object is of type: {type(module.logger)}."
-                        f" Skipping logging to {log_to}"
+                        f"Requested {logger} logging but logger object is of type: {type(module.logger)}."
+                        f" Skipping logging to {logger}"
                     )
 
-        if "local" in module.hparams.project.log_images_to:
-            visualizer.save(Path(module.hparams.project.path) / "images" / filename.parent.name / filename.name)
+        if "local" in self.loggers:
+            if trainer.default_root_dir:
```
samet-akcay, line 87: I'm not hundred percent sure here. Should we choose `trainer.default_root_dir` to save/log our stuff or `trainer.log_dir`? What do you think? @ashwinvaidya17, @djdameln

ashwinvaidya17: I guess we need a logger for accessing `log_dir`. If I am not wrong, we can save visualization results even without loggers. So in this case we should stick to `default_root_dir`.

anomalib/utils/callbacks/save_to_csv.py
```diff
@@ -48,4 +48,8 @@ def on_test_epoch_end(self, _trainer: Trainer, pl_module: AnomalyModule) -> None
                 "wrong_prediction": np.logical_xor(results.true_labels, results.pred_labels).astype(int),
             }
         )
-        data_frame.to_csv(Path(pl_module.hparams.project.path) / "results.csv")
+
+        if trainer.log_dir is not None:
+            data_frame.to_csv(Path(trainer.log_dir) / "results.csv")
+        else:
+            raise ValueError("trainer.log_dir does not exist to save the results.")
```
ashwinvaidya17, line 55: Should we maybe make the directory instead of throwing an error? Since this is cone at the end of training, we might have to train again. This might be inconvenient for algorithms like stfpm, cflow.

tests/utils/callbacks/visualizer_callback/test_visualizer.py
```diff
@@ -34,8 +34,10 @@ def test_add_images(dataset):
         )
         logger = get_dummy_logger(config, dir_loc)
         model = get_dummy_module(config)
-        trainer = pl.Trainer(callbacks=model.callbacks, logger=logger, checkpoint_callback=False)
+        callbacks = [VisualizerCallback(loggers=["local", "tensorboard"])]
```
ashwinvaidya17, line 37: In line 51, of `dummy_lightning_module.py` we initialize VisualizerCallback. I feel we should move like 37 in this file to line 51 and use `Trainer(callbacks=model.callbacks..`

ashwinvaidya17 (commented):
Thanks. I only have two comments.

anomalib/utils/callbacks/tiler.py
```diff
@@ -0,0 +1,83 @@
...
+import pytorch_lightning as pl
+from pytorch_lightning.callbacks import Callback
+
+from anomalib.models.components import AnomalyModule
+from anomalib.pre_processing.tiler import Tiler
+
+__all__ = ["TilerCallback"]
+
+
+class TilerCallback(Callback):
```
djdameln, line 29: Should we change this to `TilerConfigurationCallback`, or would it look weird from the CLI? The current name suggests that this callback applies the tiling operations.

djdameln (approved):
Thanks, looks good. Some time ago we discussed a different approach for the tiling, where we use decorators and configure the static attributes of the tiler class from the configuration callback. Now that we have a tiling configuration callback, maybe we can prioritize the implementation of this new tiling design (in a separate PR of course).

anomalib/utils/callbacks/tiler.py
samet-akcay: yes, I guess we could consider this approach after geting the first cli approach ready

ashwinvaidya17 (approved):
Thanks for the change.

anomalib/utils/callbacks/save_to_csv.py
samet-akcay: I'll probably address this in another pr, which will be kind of interconnected

PR #105, opened 2022-02-17
🩹 fix inferencers

ashwinvaidya17:
# Description

- Scores for classification models are unnormalized. We need to address that.

## Changes

- Adds support for classification models in inferencer.

## Checklist

# This test runs on nightly
image

### DIFF
--- a/anomalib/deploy/inferencers/torch.py
+++ b/anomalib/deploy/inferencers/torch.py
@@ -115,7 +115,8 @@ class TorchInferencer(Inferencer):
         Returns:
             Tensor: Output predictions.
         """
-        return self.model(image)
+        with torch.no_grad():
+            return self.model(image)
 
     def post_process(
         self, predictions: Tensor, meta_data: Optional[Union[Dict, DictConfig]] = None
--- a/anomalib/models/dfkde/model.py
+++ b/anomalib/models/dfkde/model.py
@@ -84,6 +84,20 @@ class DfkdeLightning(AnomalyModule):
         embeddings = torch.vstack(self.embeddings)
         self.normality_model.fit(embeddings)
 
+    def forward(self, batch):
+        """Forward method.
+
+        Args:
+            batch (Tensor): Input tensor
+
+        Returns:
+            Tensor: Output tensor
+        """
+        self.feature_extractor.eval()
+        layer_outputs = self.feature_extractor(batch)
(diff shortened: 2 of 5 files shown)


samet-akcay (approved):

djdameln (commented):
Isn't the torch inferencer supposed to run with the PyTorch model, and not the PL model? During inference we only need the raw predictions of the PyTorch model, and none of the training/metric/postprocessing stuff.

samet-akcay:
I think we could start considering PL predict for torch inferencer and deprecate our torch interface

ashwinvaidya17:
image
In the inferencer we create the lightning model object.

djdameln:
> I think we could start considering PL predict for torch inferencer and deprecate our torch interface

That's also possible, and might be a good idea, because that's what the PL `predict` methods are meant for. But then do we need these changes? I'm afraid it would be confusing to add yet another method implementation to the PL modules. Also why only add this for these 3 models and not the others (padim, stfpm etc)?

ashwinvaidya17:
> > I think we could start considering PL predict for torch inferencer and deprecate our torch interface
> 
> That's also possible, and might be a good idea, because that's what the PL `predict` methods are meant for. But then do we need these changes? I'm afraid it would be confusing to add yet another method implementation to the PL modules. Also why only add this for these 3 models and not the others (padim, stfpm etc)?

If we plan to switch then we don't need these changes. This was just to partially address the issues. Before this, inference was supported only for `padim, stfpm and patchcore`. These changes add support for the other models (except cflow).  This also extends the tests in [tests/nightly/deploy/test_inferencer.py](https://github.com/openvinotoolkit/anomalib/pull/105/files#diff-657ec31ec1761a238e937ccce2f9a390df8c99c89f6d3633eeb625d62e16251c)

ashwinvaidya17:
> But how does it work for the other models? For padim for example, we don't implement the `forward` method. Why is it not needed in those models. It feels inconsistent to have this method implemented for some models, but not all.

Good point. In those models were the model name is `model`, the inferencer works well. This is because in the base `AnomalyModule` class we have a forward method which calls `self.model`. So this breaks for all the models which do not use the name `model` in the derived class. So this is a hacky way of overriding the forward method.

ashwinvaidya17:
Closing as `inference.py` is going to be deprecated and most of the stuff here will be addressed in the OpenVINO PR.

Issue #111, opened 2022-02-22
IndexError: tensors used as indices must be long, byte or bool tensors


alexriedel1:
**Describe the bug**
When training PatchCore, I'm getting the following error:
`IndexError: tensors used as indices must be long, byte or bool tensors`
caused by the line https://github.com/openvinotoolkit/anomalib/blob/c33d25a19f1b622a3dd354e81976257d9bac6d7f/anomalib/models/components/dimensionality_reduction/random_projection.py#L85

**To Reproduce**
Run the PatchCore algorithm on CPU (maybe on GPU too)

The problem can be solved by casting to `c_idx` to `c_idx.long()`

innat:
@samet-akcay 
Could you please take a look into this issue? I've encountered this as well. I tried also PADIM, it worked but PatchCore failed with the above-mentioned error. Also please re-check other architecture.

samet-akcay:
Hi @alexriedel1 and @innat . Thanks for spotting this. @alexriedel1 if you prefer, feel free to create a PR for this to become a contributor! If not, we'll try to immediately add a fix

samet-akcay:
Have you guys seen this issue when you train patchcore on GPU

alexriedel1:
> Hi @alexriedel1 and @innat . Thanks for spotting this. @alexriedel1 if you prefer, feel free to create a PR for this to become a contributor! If not, we'll try to immediately add a fix

I did a PR request. I haven't tried to reproduce the issue on GPU

PR #113, opened 2022-02-23
c_idx cast to LongTensor in random sparse projection

alexriedel1:
Fixing https://github.com/openvinotoolkit/anomalib/issues/111

### DIFF
--- a/anomalib/models/components/dimensionality_reduction/random_projection.py
+++ b/anomalib/models/components/dimensionality_reduction/random_projection.py
@@ -78,11 +78,12 @@ class SparseRandomProjection:
                 c_idx = torch.tensor(
                     sample_without_replacement(
                         n_population=n_features, n_samples=nnz_idx, random_state=self.random_state
-                    )
+                    ),
+                    dtype=torch.int64
                 )
                 data = torch.distributions.Binomial(total_count=1, probs=0.5).sample(sample_shape=c_idx.size()) * 2 - 1
                 # assign data to only those columns
-                components[i, c_idx.long()] = data.double()
+                components[i, c_idx] = data.double()
 
             components *= np.sqrt(1 / density) / np.sqrt(self.n_components)
 
--- a/anomalib/models/components/dimensionality_reduction/random_projection.py
+++ b/anomalib/models/components/dimensionality_reduction/random_projection.py
@@ -79,7 +79,7 @@ class SparseRandomProjection:
                     sample_without_replacement(
                         n_population=n_features, n_samples=nnz_idx, random_state=self.random_state
                     ),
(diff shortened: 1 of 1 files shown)


innat:
```python
components[i, c_idx.long()] = data.double()

or 

c_idx = [int(t.item()) for t in c_idx]
components[i, c_idx] = data.double()
```

It's taking really much time to get coreset from embedding. Is it expected in patchcore? I'm running on CPU with 

```python
num_workers: 1 # 36
```

djdameln:
> It's taking really much time to get coreset from embedding. Is it expected in patchcore? I'm running on CPU with

Yes, the coreset computation in the training step is optimized for GPU. Inference should run fine on CPU.

djdameln (changes requested):
Thanks for reporting and addressing this issue. I would suggest constructing `c_idx` as `long` type, like so:
```
c_idx = torch.tensor(
    sample_without_replacement(
        n_population=n_features, n_samples=nnz_idx, random_state=self.random_state
    ),
    dtype=torch.int64
)
```
That way we don't need to cast it afterwards.

djdameln (approved):
Thanks!

ashwinvaidya17:
@alexriedel1 Thanks for your contribution! Let's merge after CI passes.

ashwinvaidya17:
@alexriedel1 There seems to be a minor `black` formatting issue. https://github.com/openvinotoolkit/anomalib/runs/5315789777?check_suite_focus=true Can you address it? I think it would be convenient for you if you install pre-commit hooks. 
```bash
pip install pre-commit
pre-commit install
pre-commit run --all-files
```

alexriedel1:
> @alexriedel1 There seems to be a minor `black` formatting issue. https://github.com/openvinotoolkit/anomalib/runs/5315789777?check_suite_focus=true Can you address it? I think it would be convenient for you if you install pre-commit hooks.
> 
> ```shell
> pip install pre-commit
> pre-commit install
> pre-commit run --all-files
> ```

I guess thats what black wanted to reformat. Unfortunately I wasnt abel to run the pre commit hooks..

ashwinvaidya17 (approved):
Thanks for quickly addressing this!

ashwinvaidya17:
> I guess thats what black wanted to reformat. Unfortunately I wasnt abel to run the pre commit hooks..

Can you share the error? We will try to update the docs to address this.

alexriedel1:
> > I guess thats what black wanted to reformat. Unfortunately I wasnt abel to run the pre commit hooks..
> 
> Can you share the error? We will try to update the docs to address this.

It's an issue on my side. I'm having strange permission issues when trying to run the hooks so dont worry..

PR #122, opened 2022-02-25
🔨 Increase inference + openvino support

### LINKED ISSUE #97: Ganomaly inference.py error
AttributeError: 'GanomalyLightning' object has no attribute 'model'

ashwinvaidya17:
# Description

- Increases inferencer support to all models and openvino support to ganomaly and dfm.

- Fixes #97 

## Known Issues

- Classification models still produce unnormalized scores.

## Changes

## Checklist

# Tests

- Since inference tests run on lightly
image

### DIFF
--- a/anomalib/models/cflow/model.py
+++ b/anomalib/models/cflow/model.py
@@ -165,14 +165,17 @@ class CflowModel(nn.Module):
 
         """
 
-        activation = self.encoder(images)
+        self.encoder.eval()
+        self.decoders.eval()
+        with torch.no_grad():
+            activation = self.encoder(images)
 
-        distribution = [[] for _ in self.pool_layers]
+        distribution = [torch.Tensor(0).to(images.device) for _ in self.pool_layers]
 
         height: List[int] = []
         width: List[int] = []
         for layer_idx, layer in enumerate(self.pool_layers):
-            encoder_activations = activation[layer].detach()  # BxCxHxW
+            encoder_activations = activation[layer]  # BxCxHxW
 
             batch_size, dim_feature_vector, im_height, im_width = encoder_activations.size()
             image_size = im_height * im_width
@@ -206,13 +209,15 @@ class CflowModel(nn.Module):
                 c_p = c_r[idx]  # NxP
                 e_p = e_r[idx]  # NxC
                 # decoder returns the transformed variable z and the log Jacobian determinant
-                p_u, log_jac_det = decoder(e_p, [c_p])
+                with torch.no_grad():
+                    p_u, log_jac_det = decoder(e_p, [c_p])
                 #
                 decoder_log_prob = get_logp(dim_feature_vector, p_u, log_jac_det)
                 log_prob = decoder_log_prob / dim_feature_vector  # likelihood per dim
-                distribution[layer_idx] = distribution[layer_idx] + log_prob.detach().tolist()
+                distribution[layer_idx] = torch.cat((distribution[layer_idx], log_prob))
 
         output = self.anomaly_map_generator(distribution=distribution, height=height, width=width)
+        self.decoders.train()
 
         return output.to(images.device)
 
--- a/anomalib/models/dfkde/model.py
+++ b/anomalib/models/dfkde/model.py
@@ -20,11 +20,69 @@ import torch
 import torchvision
 from omegaconf.dictconfig import DictConfig
 from omegaconf.listconfig import ListConfig
-from torch import Tensor
+from torch import Tensor, nn
 
 from anomalib.models.components import AnomalyModule, FeatureExtractor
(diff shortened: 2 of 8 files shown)


anomalib/models/dfkde/model.py
```diff
@@ -20,13 +20,71 @@
...
+        self, backbone: nn.Module, filter_count: int, threshold_steepness: float, threshold_offset: float
+    ) -> None:
+        super().__init__()
+        self.feature_extractor = FeatureExtractor(backbone=backbone(pretrained=True), layers=["avgpool"]).eval()
+
+        self.normality_model = NormalityModel(
+            filter_count=filter_count,
+            threshold_steepness=threshold_steepness,
+            threshold_offset=threshold_offset,
+        )
```
samet-akcay, line 50: This would probably work, but I feel like the model would still need another refactoring. Ideally `DfkdeModel` should be the normality model that contains feature_extractor, pca and kde modules. I guess this would be scope of another PR, but something to keep in mind.

anomalib/models/dfkde/model.py
```diff
@@ -20,13 +20,71 @@
...
+
+        Returns:
+            Tensor: Tensor containing extracted features.
+        """
+        self.feature_extractor.eval()
+        layer_outputs = self.feature_extractor(batch)
+        layer_outputs = torch.cat(list(layer_outputs.values())).detach()
+        return layer_outputs
+
+    def fit_normality_model(self, embeddings: List[Tensor]):
```
samet-akcay, line 66: i would prefer this is called just `fit`. During its usage, it would just be `self.model.fit(embeddings`

anomalib/models/dfkde/model.py
samet-akcay: Alternatively, there could be a better naming than `NormalityModel` if we want to keep `NormalityModel`. For instance, in DFM, there `SingleClassGaussian`. Again, another PR

anomalib/models/cflow/model.py
```diff
@@ -206,13 +209,15 @@ def forward(self, images):
                 c_p = c_r[idx]  # NxP
                 e_p = e_r[idx]  # NxC
                 # decoder returns the transformed variable z and the log Jacobian determinant
-                p_u, log_jac_det = decoder(e_p, [c_p])
+                with torch.no_grad():
+                    p_u, log_jac_det = decoder(e_p, [c_p])
```
samet-akcay, line 213: I thought, only encoder model is fixed. Wouldn't this fix the decoder too? how is the decoder params learn then? Am I missing the point?

ashwinvaidya17: The `training_step` contains a lot of similar code which trains the decoder. The only difference is that the forward function is used for inference. The original code had a few detach statements which broke the onnx graph. I just replaced the detach with `no_grad`.

samet-akcay: ah thanks for the clarification.

anomalib/models/dfkde/model.py
ashwinvaidya17: Created a ticket for it https://github.com/openvinotoolkit/anomalib/issues/129

anomalib/models/dfkde/model.py
```diff
@@ -20,13 +20,71 @@
...
+
+        Returns:
+            Tensor: Predictions
+        """
+        feature_vector = self.get_features(batch)
+        return self.normality_model.predict(feature_vector.view(feature_vector.shape[:2]))
+
+
 class DfkdeLightning(AnomalyModule):
     """DFKDE: Deep Featured Kernel Density Estimation.
```
djdameln, line 89: typo: Deep Feature

anomalib/models/dfkde/model.py
```diff
@@ -36,17 +94,14 @@ class DfkdeLightning(AnomalyModule):
...
+        threshold_offset = 12
 
-        self.backbone = getattr(torchvision.models, hparams.model.backbone)
-        self.feature_extractor = FeatureExtractor(backbone=self.backbone(pretrained=True), layers=["avgpool"]).eval()
-
-        self.normality_model = NormalityModel(
-            filter_count=hparams.model.max_training_points,
-            threshold_steepness=self.threshold_steepness,
-            threshold_offset=self.threshold_offset,
+        backbone = getattr(torchvision.models, hparams.model.backbone)
```
djdameln, line 100: Maybe you could create the backbone inside `DfkdeModel` and just pass the `hparams.model.backbone` to the constructor of that class. That would be more in line with other models (see Padim for example).

djdameln (changes requested):

samet-akcay (approved):

PR #126, opened 2022-02-28
Update Nightly

ashwinvaidya17:
# Changes

- Add badges to readme
- Reduce thresholds for ganomaly and stfpm
- Increase tolerance
- Disable stfpm nncf tests
- Remove gpu restriction for nightly job

# Last run
- https://github.com/ashwinvaidya17/anomalib/actions/workflows/nightly.yml


tests/nightly/models/performance_thresholds.yaml
```diff
@@ -114,7 +114,7 @@ stfpm:
     image_AUROC: 0.857
     pixel_AUROC: 0.962
     nncf:
-      image_AUROC: 0.979
+      image_AUROC: 0.902
```
djdameln, line 117: In the future we could maybe replace this with an average value and ± tolerance. But for now this should work.

djdameln (approved):

tests/nightly/models/performance_thresholds.yaml
samet-akcay: Yes! This is an ideal approach for papers too

samet-akcay (approved):
Thank a lot for this work! Great effort!

PR #141, opened 2022-03-09
optimize compute_anomaly_score by using torch native funcrtions

alexriedel1:
Hey,
I realized that the `compute_anomy_score` method https://github.com/openvinotoolkit/anomalib/blob/ee8807b48d7c5ccfe9f5ff976fb6d6dac0247f0a/anomalib/models/patchcore/model.py#L69 can be optimized by using torch native functions that might vectorize better.

Using a GPU, the execution time reduces from `0.17330350000000116s` to `0.00022549999999910142s` (~700x faster)

### DIFF
--- a/anomalib/models/patchcore/model.py
+++ b/anomalib/models/patchcore/model.py
@@ -74,9 +74,10 @@ class AnomalyMapGenerator:
         Returns:
             torch.Tensor: Image-level anomaly scores
         """
-        confidence = patch_scores[torch.argmax(patch_scores[:, 0])]
+        max_scores = torch.argmax(patch_scores[:, 0])
+        confidence = torch.index_select(patch_scores, 0, max_scores)
         weights = 1 - (torch.max(torch.exp(confidence)) / torch.sum(torch.exp(confidence)))
-        score = weights * max(patch_scores[:, 0])
+        score = weights * torch.max(patch_scores[:, 0])
         return score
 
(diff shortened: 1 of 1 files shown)


samet-akcay (approved):
Thanks @alexriedel1! That's a good catch! We'd be able to merge it after @djdameln approves the PR

alexriedel1:
> Thanks @alexriedel1! That's a good catch! We'd be able to merge it after @djdameln approves the PR

Theres one more thing I'd like to adress: `torch.cdist()` is slow (https://github.com/pytorch/pytorch/issues/15253)
https://github.com/openvinotoolkit/anomalib/blob/5a5e552cb4f3dbcdc8d9bff788b9cbcd445f9496/anomalib/models/patchcore/model.py#L239

This cdist implementation is fast: https://github.com/pytorch/pytorch/issues/15253#issuecomment-491467128

Maybe you can think about using the version mentioned above. I can create another pull request for this.

djdameln (approved):
Thanks for addressing this @alexriedel1!

samet-akcay:
@alexriedel1, would you like to address that in a separate PR, or in this one? If the former, I could merge this one.

alexriedel1:
> @alexriedel1, would you like to address that in a separate PR, or in this one? If the former, I could merge this one.

Let me do another PR so you can merge this one

PR #146, opened 2022-03-14
Fix IndexError in adaptive threshold computation

djdameln:
# Description

- This PR addresses a bug causing validation to fail when computing the adaptive threshold. In rare cases (e.g. few anomalous images in validation set), it is possible that the recall is 1.0 for all threshold values included in TorchMetrics precision-recall curve computation. In this case, the `thresholds` tensor gets reduced to a scalar value, and trying to retrieve the first item produces an `IndexError`. In this case we can safely return the scalar threshold value instead.

## Changes

## Checklist


samet-akcay (approved):
thanks for spotting this!

PR #154, opened 2022-03-22
Add Custom Dataset Training Support

### LINKED ISSUE #147: Support training with custom MVTec like dataset but without masks (ground truths)
Hello,

I would like to be able to train on custom datasets without the need to have ground truth masks since most anomaly detection methods do not use ground truth for training. The format would be the same as any MVTec dataset just without the ground_truth directory.

Edit: I changed data/mvtec.py:253  `if label_index == 0 or not Path(mask_path).exists():
                    mask = np.zeros(shape=image.shape[:2])`
An empty mask is created if mask_path doesn't exist and it seemed to do the trick

samet-akcay:
# Description

- This PR adds custom dataset support.

- Fixes #147 

## Changes

## Checklist


README.md
```diff
@@ -102,6 +102,33 @@ where the currently available models are:
...
+```yaml
+dataset:
+  name: custom
+  path: <path/to/custom/dataset>
+  normal: normal # name of the folder containing normal images.
+  abnormal: abnormal # name of the folder containing abnormal images.
+  task: segmentation # classification or segmentation
+  mask: <path/to/mask/annotations> #optional
+  extensions: null
+  split_ratio: 0.2
```
djdameln, line 116: Maybe add some comments here to the parameters that may be hard to understand. e.g.
`split_ratio: 0.2  # ratio of the normal images that will be used to create a test split`

README.md
```diff
@@ -102,6 +102,33 @@ where the currently available models are:
 - [DFKDE](anomalib/models/dfkde)
 - [GANomaly](anomalib/models/ganomaly)
 
+### Custom Dataset
+It is also possible to train on a custom dataset. To do so, `data` section in `config.yaml` is to be modified as follows:
+```yaml
+dataset:
+  name: custom
```
djdameln, line 109: Maybe we should use `format` here instead of `name`. For MVTec we also have a `format` field in addition to `name`. The way I see it, `format` determines which dataset class is used under the hood, while `name` can be anything that identifies the specific dataset that is used.

anomalib/config/config.py
```diff
@@ -177,7 +177,8 @@ def get_configurable_parameters(
     config = update_input_size_config(config)
 
     # Project Configs
-    project_path = Path(config.project.path) / config.model.name / config.dataset.name / config.dataset.category
+    category = config.dataset.category if "category" in config.dataset.keys() else ""
```
djdameln, line 180: Maybe it would be a bit more clear if we check the dataset type here, and only add the category to the path if the type is MVTec.

anomalib/data/custom.py
```diff
@@ -0,0 +1,490 @@
...
+
+    # Get the data frame for the split.
+    if split is not None and split in ["train", "val", "test"]:
+        samples = samples[samples.split == split]
+        samples = samples.reset_index(drop=True)
+
+    return samples
+
+
+class CustomDataset(Dataset):
```
djdameln, line 154: I'm not sure about the naming. Maybe `FolderDataset` would be more appropriate? Custom sounds a bit like users can choose their own 'custom' format. But this class represents a dataset that follows a fixed format based on the folder structure of the data.

anomalib/data/custom.py
```diff
@@ -0,0 +1,490 @@
...
+
+            >>> i, test_data = next(enumerate(datamodule.test_dataloader()))
+            >>> test_data.keys()
+            dict_keys(['image'])
+
+            We could also create a Custom DataModule for datasets containing mask annotations.
+            The dataset expects that mask annotation filenames must be same as the original filename.
+            To show an example, we therefore need to modify the mask filenames in MVTec dataset.
+
+            >>> # Rename MVTec mask annotations so that they are the same as image filanames
```
djdameln, line 338: I'm afraid the example in the docstring might cause some confusion with the users (why use the custom dataset class for MVTec if there is a dataset class specific for mvtec). Maybe we could keep it simple and start the example with the assumption that the user has a folder of normal images and a folder of abnormal images, and explicitly state this at the beginning of the example.

djdameln (commented):
Thanks, great addition! I didn't manually test the custom dataset format yet, but I'll do that and will post here if I run into any issues.

djdameln (approved):
Thanks!

Issue #150, opened 2022-03-20
Requirements incompatible with M1 Mac


dreaquil:
**Describe the bug**

Currently when installing the requirements on an M1 Mac, there are certain packages (e.g. torch, numpy, scikit-learn) for which the current version cannot be found or does not work.

**To Reproduce**
1. `pip install requirements/base.txt`

**Expected behavior**
- A clear and concise description of what you expected to happen.

**Screenshots**
- If applicable, add screenshots to help explain your problem.

**Hardware and Software Configuration**
- OS: MacOs
- NVIDIA Driver Version [470.57.02]: NA
- CUDA Version [e.g. 11.4] :NA
- CUDNN Version [e.g. v11.4.120]: NA
- OpenVINO Version [Optional e.g. v2021.4.2]

**Additional context**
- Add any other context about the problem here.

samet-akcay:
Hi @dreaquil, can you paste the exact error you're facing so we could reproduce the problem? Thanks!

dreaquil:
Hi @samet-akcay, there's a few different errors that one needs to work through before getting a functional install. I have started a branch which has the changes I needed to make [here](https://github.com/dreaquil/anomalib/tree/150-m1-mac).
https://github.com/openvinotoolkit/anomalib/pull/151
So far I have found issues in the versions specified of:
- torch (fixed with upgrade)
- torchvision (fixed with upgrade)
- numpy (fixed with upgrade)
- scikit-learn (fixed with upgrade)
- pandas (missing altogether?)
- openvino (not fixed)
- nncf (not fixed)

dreaquil:
hi @samet-akcay, just a follow up on this morning's reply. I think the two main issues are nncf and openvino for the M1 mac. The problem is that their dependencies are pinned to exact versions. A lot of packages (like numpy) have only started working on the M1 mac over the past 6 months. This means that I need to upgrade numpy to make it work but this causes a dependency conflict with nncf. Any thoughts on this?

```
The conflict is caused by:
    anomalib 0.2.4 depends on numpy>=1.22.3
    albumentations 1.1.0 depends on numpy>=1.11.1
    matplotlib 3.4.3 depends on numpy>=1.16
    nncf 2.1.0 depends on numpy~=1.19.2
```

dreaquil:
Hi @samet-akcay,

[...]

PR #158, opened 2022-03-22
WIP Fix/m1 mac

dreaquil:
# Description

- Fixes # (issue)
- https://github.com/openvinotoolkit/anomalib/issues/150

## Changes

## Checklist


samet-akcay:
@dreaquil, we'll need to move nncf and openvino related requirements to `requirements/openvino.txt`. 

For the openvino imports in the inferencers, we'll first need to check if openvino is installed before importing the package. Alternatively we could import openvino in the corresponding part of the code rather than at the top of the module. 

For the package version numbers we need some through testing internally to ensure that all the components work fine, and there is no performance degradation in the algorithms. After the check we could use these version numbers.

samet-akcay:
@dreaquil, looks like the branch has conflicts to be solved. If you want to get the pr merged, would you be able to resolve the conflicts please? Thanks

dreaquil:
Hi @samet-akcay! Just checking in that the CI is still broken right?

samet-akcay:
@dreaquil, almost fixed. we'll try to get it up and running before the end of today.

anomalib/deploy/inferencers/openvino.py
```diff
@@ -20,7 +20,8 @@
 import cv2
 import numpy as np
 from omegaconf import DictConfig, ListConfig
-from openvino.inference_engine import IECore  # pylint: disable=no-name-in-module
+
+# from openvino.inference_engine import IECore  # pylint: disable=no-name-in-module
```
samet-akcay, line 24: It's unfortunately not possible to comment out this import since openvino stuff depends on this. Perhaps, we could add a check if openvino is installed or not.

requirements/base.txt
```diff
@@ -7,16 +7,17 @@ kornia==0.5.6
 lxml==4.6.5
 matplotlib==3.4.3
 networkx~=2.5
-numpy~=1.19.5
+# nncf==2.1.0
```
samet-akcay, line 10: This line is not needed it was moved to `openvino` requirements now.

samet-akcay, line 10: This line is not needed it was moved to `openvino` requirements now.

requirements/base.txt
```diff
@@ -7,16 +7,17 @@ kornia==0.5.6
...
-torch==1.8.1
-torchvision==0.9.1
-torchtext==0.9.1
+pytorch-lightning==1.6.0
+torch==1.11.0
+torchvision==0.12.0
+torchtext==0.12.0
 scikit-image>=0.17.2
-scikit-learn==0.24.2
+scikit-learn==1.0.2
```
samet-akcay, line 22: it might be an idea to make these dependencies a bit more flexible. Rather than asking for the exact version, we could maybe do something like `scikit-learn>=0.24.2`. In this case would you be able to download the version you want? The reason is that there could be some usecases where another library depends on an earlier package version, but anomalib has the latest version as the dependency, which ultimately would causes hard package dependency conflicts.

samet-akcay, line 22: it might be an idea to make these dependencies a bit more flexible. Rather than asking for the exact version, we could maybe do something like `scikit-learn>=0.24.2`. In this case would you be able to download the version you want? The reason is that there could be some usecases where another library depends on an earlier package version, but anomalib has the latest version as the dependency, which ultimately would causes hard package dependency conflicts.

requirements/base.txt
```diff
@@ -7,16 +7,17 @@ kornia==0.5.6
...
+pandas==1.4.1
 opencv-python>=4.5.3.56
 opencv-contrib-python==4.5.5.62
-pandas~=1.1.5
 pillow==9.0.0
-pytorch-lightning==1.5.9
-torch==1.8.1
-torchvision==0.9.1
-torchtext==0.9.1
+pytorch-lightning==1.6.0
```
samet-akcay, line 17: pytorch-lightning 1.6.0 modified some of the annotations under the hood. CI is failing due to these changes. 
[CI-PR-158.log](https://github.com/openvinotoolkit/anomalib/files/8427463/CI-PR-158.log)

samet-akcay, line 17: pytorch-lightning 1.6.0 modified some of the annotations under the hood. CI is failing due to these changes. 
[CI-PR-158.log](https://github.com/openvinotoolkit/anomalib/files/8427463/CI-PR-158.log)

samet-akcay (changes requested):
@dreaquil, these changes causes the CI to fail due linting and mypy issues. I'll try to check if we could find a common ground to install anomalib on M1 and pass the tests. 

I've attached the CI logs here, showing where/why the CI fails.
[CI-PR-158.log](https://github.com/openvinotoolkit/anomalib/files/8427483/CI-PR-158.log)

requirements/base.txt
dreaquil: This is a problem because I cannot install torchtext 0.9.1 but it does not seem that pytorch lightning 1.5.9 works with 0.12.0.

anomalib/deploy/inferencers/openvino.py
dreaquil: This is no longer an issue on mac.

requirements/base.txt
samet-akcay: can you try the below requirements file to see if it works on you mac?
https://github.com/openvinotoolkit/anomalib/blob/fix/sa/requirements/requirements/base.txt

samet-akcay: With those requirements, anomalib tests pass. They shoudl be working fine with your mac too.

dreaquil: kornia needs to be tied down to 0.5.6 otherwise it throws this error ImportError: cannot import name 'gaussian_blur2d' from 'kornia'

matplotlib seems to be missing from the requirements somehow. I changed my base to match yours and added matplotlib and fixed the kornia dependency.

samet-akcay: regarding the kornia, looks like we need to modify the import as follows:

~~from kornia import gaussian_blur2d~~
from kornia.filters import gaussian_blur2d
 in `anomalib/models/patchcore/model.py` and `anomalib/models/padim/model.py`

dreaquil: Yep that's right. I also checked that import with the 0.5.6 and it seems to work too. Pushed the change. I can't seem to get the test suite to autodiscover on my machine as it's throwing errors in albumentations but I can give it a try tomorrow. Are tests passing on yours?

samet-akcay: When I use the base requirements that I shared with you, tests pass. I used #193 as well, though. I'll be merging it, so you might want to pull the latest develop branch after I merge it. I could trigger your tests afterwards.

samet-akcay: Just merged PR #193 now.

dreaquil: Merged here.

samet-akcay: Thanks, I'll run the CI now.

samet-akcay: CI failed due to sorting imports. Can you run 

```
isort  --df anomalib --profile=black
```

from your terminal and push the changes?

dreaquil: Ah sorry, left an extra line. Done!

dreaquil: @samet-akcay, saw the CI failed. Pushed a fix for tox. Would you mind running the workflow again please?

requirements/base.txt
```diff
@@ -1,21 +1,11 @@
...
 pytorch-lightning>=1.6.0
-torch==1.8.1
-torchvision==0.9.1
-torchtext==0.9.1
-scikit-image>=0.17.2
-scikit-learn==0.24.2
-wandb==0.12.9
+torchvision>=0.9.1
+torchtext>=0.9.1
+wandb>=0.12.9
```
samet-akcay, line 10: @dreaquil, this also needs to be reverted back to `wandb==0.12.9` for now. For some reason, CI fails due to this. We could address it later.

requirements/base.txt
samet-akcay: @dreaquil, as I mentioned above, if you could set `wandb==0.12.9`, I could re-run the CI, which would hopefully pass this time.

dreaquil: Haha, okay... this should be it! @samet-akcay

samet-akcay:
@dreaquil, looks like black also fails. Can you run `black -l 120 anomalib` from your terminal, and hopefully this will be the last change. Thanks :)

if `black` is not installed in your virtualenv, you could run `pip install black`

dreaquil:
> @dreaquil, looks like black also fails. Can you run `black -l 120 anomalib` from your terminal, and hopefully this will be the last change. Thanks :)
> 
> if `black` is not installed in your virtualenv, you could run `pip install black`

Done! here's hoping :)

samet-akcay:
CI has passed finally 🎉. I'm gonna be merging the PR now

samet-akcay (approved):

Issue #171, opened 2022-03-29
Dockerfile improvements


LukasBommes:
The provided Dockerfile does not work as expected. I had to change it as follows:

```
#########################################################
## Python Environment with CUDA
#########################################################

FROM nvidia/cuda:11.4-devel-ubuntu20.04 AS python_base_cuda
LABEL MAINTAINER="Anomalib Development Team"

# Setup Proxies
#ENV http_proxy=http://proxy-dmz.intel.com:912
#ENV https_proxy=http://proxy-dmz.intel.com:912
#ENV ftp_proxy=http://proxy-dmz.intel.com:912

# Update system and install wget
RUN apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get install -y wget ffmpeg libpython3.8 git sudo

# Install Conda
RUN wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh --quiet && \
    bash ~/miniconda.sh -b -p /opt/conda
ENV PATH "/opt/conda/bin:${PATH}"
RUN conda install python=3.8

#########################################################
## Anomalib Development Env
#########################################################

FROM python_base_cuda as anomalib_development_env

# Get MVTec Dataset
# cache datasets first as changes to requirements do no affect this stage
RUN wget ftp://guest:GU.205dldo@ftp.softronics.ch/mvtec_anomaly_detection/mvtec_anomaly_detection.tar.xz --quiet && \
    mkdir -p /tmp/anomalib/datasets/MVTec && \
    tar -xf mvtec_anomaly_detection.tar.xz -C /tmp/anomalib/datasets/MVTec

# Install all anomalib requirements
COPY ./requirements/base.txt /tmp/anomalib/requirements/base.txt
RUN pip install -r /tmp/anomalib/requirements/base.txt

COPY ./requirements/openvino.txt /tmp/anomalib/requirements/openvino.txt
RUN pip install -r /tmp/anomalib/requirements/openvino.txt

# Install other requirements related to development
COPY ./requirements/dev.txt /tmp/anomalib/requirements/dev.txt
RUN pip install -r /tmp/anomalib/requirements/dev.txt
```
I.e. setup of proxies is not needed for end users and the line `ENV PATH "/opt/conda/bin:${PATH}" && conda install python=3.8` in the Dockerfile lead to a corrupted path.

[...]

PR #172, opened 2022-03-29
Dockerfile enhancements

LukasBommes:
# Description

Improves the Dockerfile. This fixes issue #171.

I removed the caching of the MVTec dataset, since it is anyway downloaded upon calling `tools/train.py`. Furthermore, anomalib is installed inside the image.

## Changes

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,18 +5,14 @@
 FROM nvidia/cuda:11.4.0-devel-ubuntu20.04 AS python_base_cuda
 LABEL MAINTAINER="Anomalib Development Team"
 
-# Setup Proxies
-ENV http_proxy=http://proxy-dmz.intel.com:912
-ENV https_proxy=http://proxy-dmz.intel.com:912
-ENV ftp_proxy=http://proxy-dmz.intel.com:912
-
 # Update system and install wget
 RUN apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get install -y wget ffmpeg libpython3.8 git sudo
 
 # Install Conda
 RUN wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh --quiet && \
     bash ~/miniconda.sh -b -p /opt/conda
-ENV PATH "/opt/conda/bin:${PATH}" && conda install python=3.8
+ENV PATH "/opt/conda/bin:${PATH}"
+RUN conda install python=3.8
 
 
 #########################################################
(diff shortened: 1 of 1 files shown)


LukasBommes:
Maybe, we could also add a brief usage description in the readme. It could state that the image is built as follows:
`docker build . --tag=anomalib`
and can be run from the project's root directory with
`docker run -it --ipc=host --env="DISPLAY" --gpus=all -v /tmp/.X11-unix:/tmp/.X11-unix:rw -v "$(pwd)":/anomalib anomalib /bin/bash`
It may also require disabling access control to the X11 server by running `xhost +` on the host. This is needed to forward graphical output of the container to the host.

samet-akcay:
> Maybe, we could also add a brief usage description in the readme. It could state that the image is built as follows:
docker build . --tag=anomalib
and can be run from the project's root directory with
docker run -it --ipc=host --env="DISPLAY" --gpus=all -v /tmp/.X11-unix:/tmp/.X11-unix:rw -v "$(pwd)":/anomalib anomalib /bin/bash
It may also require disabling access control to the X11 server by running xhost + on the host. This is needed to forward graphical output of the container to the host.

@LukasBommes, we have a [section](https://github.com/openvinotoolkit/anomalib/blob/development/docs/source/guides/developing_on_docker.rst) in the documentation explaining how to develop on docker. Would it be an idea to include these descriptions as well as linking the dataset as you mentioned in the issue? What do you think?

LukasBommes:
The "Developing on Docker" section seems a good place. I will add the usage comment there.

The dataset is downloaded when running `tools/train.py` for the first time, right? Hence, I would just leave it to the python script to perform the download. When the container is run with the `-v "$(pwd)":/anomalib` argument will map all directory contents (including the dowanloaded dataset) into the container. Does that sound like a good solution? Otherwise, there would be redundant code for the download (once in the python code and once in the Docker entrypoint script).

LukasBommes:
Actually, the "Developing on Docker" section could be cleaned up a bit. I n my opinion it is unecessary to describe the use of the VSCode Remote-Containers extension. I'd just put the bare minimum of shell commands for building and running the container there. If someone wants to use the VSCode extension, then he could refer to the respective doumentation in VSCode.
What do you think?

ashwinvaidya17:
closes #171

ashwinvaidya17 (approved):
Thanks for looking into this! The original image was created for the gitlab runner but it has been deprecated. It is a good idea to decouple bare minimum development environment and the rest of the code and dataset. Mapping directory makes sense at this point. Also, I like the idea of forwarding X11 to host machine. Can you add it to the documentation/create a ticket for it?

PR #175, opened 2022-03-30
Fix `KeyError:'label'` in classification folder dataset

### LINKED ISSUE #173: Issues with heatmap scaling when training without masks
I ran into an issue when training on a dataset (subset of MVTec), where I set all ground truth masks to zero (to simulate training on a dataset for which I have no ground truth masks). When training with the actual ground truth masks, the model produces heatmaps as expected as in the first image below (produced with `tools/inference.py`). However, when training with the zero masks, the heatmaps seem to be scaled differently as in the second image below. The confidence score seems unaffected.

ground_truth_masks

zero_masks

This behaviour is the same for both PADIM and PatchCore. I haven't tested the other models.

This is my model config for PADIM
```
dataset:
  name: mvtec_test
  format: folder
  path: ./datasets/mvtec_test/images
  normal: normal
  abnormal: abnormal
  task: segmentation
  mask: ./datasets/mvtec_test/masks_orig
  extensions: null
  seed: 0  
  image_size: 224
  train_batch_size: 32
  test_batch_size: 1
  num_workers: 16
  transform_config: null
  split_ratio: 0.2
  create_validation_set: true
  tiling:
    apply: false
    tile_size: null
    stride: null
    remove_border_count: 0
    use_random_tiling: False
    random_tile_count: 16

model:
  name: padim
  backbone: resnet18
  layers:
    - layer1
    - layer2
    - layer3
  metric: auc
  normalization_method: min_max # options: [none, min_max, cdf]
  threshold:
    image_default: 3
    pixel_default: 3
    adaptive: true

project:
  seed: 42
  path: ./results
  log_images_to: ["local"]
  logger: false
  save_to_csv: false

optimization:
  openvino:
    apply: false

# PL Trainer Args. Don't add extra parameter here.
trainer:
  accelerator: null
  accumulate_grad_batches: 1
  amp_backend: native
  auto_lr_find: false
  auto_scale_batch_size: false
  auto_select_gpus: false
  benchmark: false
  check_val_every_n_epoch: 1 # Don't validate before extracting features.
  checkpoint_callback: true
  default_root_dir: null
  deterministic: false
  fast_dev_run: false
  gpus: 1
  gradient_clip_val: 0
  limit_predict_batches: 1.0
  limit_test_batches: 1.0
  limit_train_batches: 1.0
  limit_val_batches: 1.0
  log_every_n_steps: 50
  log_gpu_memory: null
  max_epochs: 1
  max_steps: -1
  min_epochs: null
  min_steps: null
  move_metrics_to_cpu: false
  multiple_trainloader_mode: max_size_cycle
  num_nodes: 1
  num_processes: 1
  num_sanity_val_steps: 0
  overfit_batches: 0.0
  plugins: null
  precision: 32
  prepare_data_per_node: true
  process_position: 0
  profiler: null
  progress_bar_refresh_rate: null
  replace_sampler_ddp: true
  stochastic_weight_avg: false
  sync_batchnorm: false
  terminate_on_nan: false
  tpu_cores: null
  track_grad_norm: -1
  val_check_interval: 1.0 # Don't validate before extracting features.
  weights_save_path: null
  weights_summary: top
```

samet-akcay:
# Description

- This hotfix proposes a fix to return `label` in `folder` dataset when `classification` task is chosen. Currently the `folder` dataset doesn't return `label` in `classification`, which ultimately causes a failed validation.
- PR also bumps up the black version due to a version dependency, resulting in some changes in some modules.

- Fixes #173

## Changes

## Checklist

### DIFF
--- a/anomalib/utils/callbacks/visualizer_callback.py
+++ b/anomalib/utils/callbacks/visualizer_callback.py
@@ -111,25 +111,26 @@ class VisualizerCallback(Callback):
             normalize = True  # raw anomaly maps. Still need to normalize
         threshold = pl_module.pixel_metrics.F1.threshold
 
-        for (filename, image, true_mask, anomaly_map) in zip(
-            outputs["image_path"], outputs["image"], outputs["mask"], outputs["anomaly_maps"]
-        ):
-            image = Denormalize()(image.cpu())
-            true_mask = true_mask.cpu().numpy()
-            anomaly_map = anomaly_map.cpu().numpy()
+        if isinstance(outputs, dict) and "mask" in outputs.keys():
+            for (filename, image, true_mask, anomaly_map) in zip(
+                outputs["image_path"], outputs["image"], outputs["mask"], outputs["anomaly_maps"]
+            ):
+                image = Denormalize()(image.cpu())
+                true_mask = true_mask.cpu().numpy()
+                anomaly_map = anomaly_map.cpu().numpy()
 
-            heat_map = superimpose_anomaly_map(anomaly_map, image, normalize=normalize)
-            pred_mask = compute_mask(anomaly_map, threshold)
-            vis_img = mark_boundaries(image, pred_mask, color=(1, 0, 0), mode="thick")
+                heat_map = superimpose_anomaly_map(anomaly_map, image, normalize=normalize)
+                pred_mask = compute_mask(anomaly_map, threshold)
+                vis_img = mark_boundaries(image, pred_mask, color=(1, 0, 0), mode="thick")
 
-            visualizer = Visualizer(num_rows=1, num_cols=5, figure_size=(12, 3))
-            visualizer.add_image(image=image, title="Image")
-            visualizer.add_image(image=true_mask, color_map="gray", title="Ground Truth")
-            visualizer.add_image(image=heat_map, title="Predicted Heat Map")
-            visualizer.add_image(image=pred_mask, color_map="gray", title="Predicted Mask")
-            visualizer.add_image(image=vis_img, title="Segmentation Result")
-            self._add_images(visualizer, pl_module, Path(filename))
-            visualizer.close()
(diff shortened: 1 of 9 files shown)


alexriedel1:
Hi,
do you think it would make sense to also visualize in case of classification, but then calling something like the `add_label` function? https://github.com/openvinotoolkit/anomalib/blob/8c1a04fc7fbd95fae1dac6cb0641a7381ec8e5d4/tools/inference.py#L48

anomalib/utils/callbacks/visualizer_callback.py
```diff
@@ -111,25 +111,26 @@ def on_test_batch_end(
...
-
-            visualizer = Visualizer(num_rows=1, num_cols=5, figure_size=(12, 3))
-            visualizer.add_image(image=image, title="Image")
-            visualizer.add_image(image=true_mask, color_map="gray", title="Ground Truth")
-            visualizer.add_image(image=heat_map, title="Predicted Heat Map")
-            visualizer.add_image(image=pred_mask, color_map="gray", title="Predicted Mask")
-            visualizer.add_image(image=vis_img, title="Segmentation Result")
-            self._add_images(visualizer, pl_module, Path(filename))
-            visualizer.close()
+        if isinstance(outputs, dict) and "mask" in outputs.keys():
```
djdameln, line 114: It would be good if we could also visualize the results when no masks are not available, but we could address this separately.

djdameln (approved):
Thanks, looks good to me!

samet-akcay:
Hi @alexriedel1, yes, that would definitely make sense. I've created this PR as a hotfix to address this bug. Not sure if I have time to address the classification visualization issue today, but it's been on our list for a while. Perhaps, we could address that in a separate PR?

anomalib/utils/callbacks/visualizer_callback.py
samet-akcay: Exactly, similar to what @alexriedel1 proposed. A separate PR would address this

PR #200, opened 2022-04-06
Fix docs workflow

ashwinvaidya17:
# Description

- Run: https://github.com/openvinotoolkit/anomalib/runs/5849349970?check_suite_focus=true

- Fixes Docs push issue. It also uses github token to avoid ssh.


samet-akcay:
PR 200 🎉🥳🎊

samet-akcay (approved):

PR #204, opened 2022-04-07
🐞fix support for non-square images

### LINKED ISSUE #136: GANomaly breaks when changing image size
**Describe the bug**
- GANomaly training fails when changing the image size in the config.

**To Reproduce**
1. Change the `image_size` parameter in the Patchcore config, to for example `[256, 234]`
2. Run training.
3. Training will fail with a runtime error.

**Expected behavior**
- Training should succeed, or notify the user that the specified image size is invalid.

- OS: Ubuntu
- NVIDIA Driver Version [470.103.01]
- CUDA Version [e.g. 11.1]

ashwinvaidya17:
# Description

- Computes padding based on the image size. Might not be the best approach.

- Fixes #136 

## Changes

## Checklist

### DIFF
--- a/anomalib/data/utils/image.py
+++ b/anomalib/data/utils/image.py
@@ -14,11 +14,14 @@
 # See the License for the specific language governing permissions
 # and limitations under the License.
 
+import math
 from pathlib import Path
 from typing import List, Union
 
 import cv2
 import numpy as np
+import torch.nn.functional as F
+from torch import Tensor
 from torchvision.datasets.folder import IMG_EXTENSIONS
 
 
@@ -66,3 +69,23 @@ def read_image(path: Union[str, Path]) -> np.ndarray:
     image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
 
     return image
+
+
+def pad_nextpow2(batch: Tensor) -> Tensor:
+    """Compute required padding from input size and return padded images.
+
+    Finds the largest dimension and computes a square image to pass into the model. In case the image dimension
+    is odd, it returns the image with an extra padding on one side.
+
+    Args:
+        batch (Tensor): Input images
+
+    Returns:
+        batch: Padded batch
+    """
+    # find the largest dimension
+    l_dim = 2 ** math.ceil(math.log(max(*batch.shape[-2:]), 2))
+    padding_w = [math.ceil((l_dim - batch.shape[-2]) / 2), math.floor((l_dim - batch.shape[-2]) / 2)]
+    padding_h = [math.ceil((l_dim - batch.shape[-1]) / 2), math.floor((l_dim - batch.shape[-1]) / 2)]
+    padded_batch = F.pad(batch, pad=[*padding_h, *padding_w])
+    return padded_batch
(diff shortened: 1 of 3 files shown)


anomalib/models/ganomaly/torch_model.py
```diff
@@ -40,13 +42,14 @@ def __init__(
     ):
         super().__init__()
 
-        assert input_size % 16 == 0, "Input size should be a multiple of 16"
+        assert input_size[0] % 16 == 0 and input_size[1] % 16 == 0, "Input size should be a multiple of 16"
```
djdameln, line 45: I guess this is no longer needed now that we use padding to make sure that the network can process any image size.

anomalib/models/ganomaly/torch_model.py
```diff
@@ -74,14 +77,33 @@ def __init__(
...
+
+        Args:
+            input_size (Tuple[int, int]): Input size
+
+        Returns:
+            Tuple[int, int]: Padding for each dimension
+        """
+        # find the largest dimension
+        l_dim = 2 ** math.ceil(math.log(max(*input_size), 2))
+        padding = math.ceil((l_dim - input_size[0]) // 2 + 1), math.ceil((l_dim - input_size[1]) // 2 + 1)
```
djdameln, line 104: This won't work for odd image size, e.g. 221

djdameln (changes requested):

anomalib/models/ganomaly/torch_model.py
```diff
@@ -363,6 +394,26 @@ def forward(self, batch: Tensor) -> Tensor:
         Returns:
             Tensor: Regeneration scores.
         """
+        padded_batch = self.get_padded_tensor(batch)
         self.generator.eval()
-        _, latent_i, latent_o = self.generator(batch)
+        _, latent_i, latent_o = self.generator(padded_batch)
         return torch.mean(torch.pow((latent_i - latent_o), 2), dim=1).view(-1)  # convert nx1x1 to n
+
+    def get_padded_tensor(self, batch: Tensor) -> Tensor:
```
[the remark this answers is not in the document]
ashwinvaidya17: I'll update it in a bit. Static is a good idea. In that case would it make sense to move it to `data/utils/image.py`?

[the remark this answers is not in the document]
djdameln: Yes, I think it would make sense to move it to a shared location so that other modules could access it as well. In that case we should maybe name it `pad_nextpow2` or similar

anomalib/data/utils/image.py
```diff
@@ -66,3 +69,23 @@ def read_image(path: Union[str, Path]) -> np.ndarray:
     image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
 
     return image
+
+
+def pad_nextpow2(batch: Tensor) -> Tensor:
+    """Compute required padding from input size and return padded images.
+
+    Finds the largest dimension and computes a square image to pass into the model. In case the image dimension
```
djdameln, line 77: Maybe mention in the description that the function pads the images so that both sides are a power of 2.

djdameln (changes requested):
Thanks! Just one minor comment left

djdameln (approved):

PR #210, opened 2022-04-08
✨ Directory streaming

### LINKED ISSUE #182: Inference on a directory of images
It would be handy if the inference script could handle not only a single image, but an entire folder of images. Since looking at only a single example does not provide a complete picture of the algorithm's performance. 

I know, this could be achieved with a bash script. But it is easier for newcomers to have this functionality built into the inference script.

I would automatically disable graphical output (imshow) when performing inference on an entire folder (requring specifiction of the `save_path` in that case). The directory and its subdirectories could be recursively traversed and the outputs could be stored in the same directory structure at the location indicated by the `save_path`.

I could make a PR with this. What do you think?

ashwinvaidya17:
# Description

- Add stream function to go over all images in a directory if it is provided as the image path.

- Fixes #182 

## Changes

## Checklist

### DIFF
--- a/tools/inference.py
+++ b/tools/inference.py
@@ -21,6 +21,7 @@ command line, and show the visualization results.
 from argparse import ArgumentParser, Namespace
 from importlib import import_module
 from pathlib import Path
+from typing import Optional
 
 import cv2
 import numpy as np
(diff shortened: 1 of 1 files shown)


tools/inference.py
```diff
@@ -108,11 +134,17 @@ def infer() -> None:
     # Show or save the output image, depending on what's provided as
     # the command line argument.
     output = cv2.cvtColor(output, cv2.COLOR_RGB2BGR)
```
djdameln, line 136: Maybe we should also show the segmentation predictions here, but I guess we could address it separately

djdameln (approved):

tools/inference.py
ashwinvaidya17: I didn't understand this. Aren't we overlaying anomaly masks on line 127?

samet-akcay (approved):

tools/inference.py
djdameln: This is what I get when I run the inferencer. It just shows the heatmap and the confidence score, no segmentation predictions.
Screenshot from 2022-04-11 12-37-36

PR #217, opened 2022-04-11
Updated documentation for development on Docker

LukasBommes:
# Description

Updates the [Developing on Docker](https://github.com/openvinotoolkit/anomalib/blob/development/docs/source/guides/developing_on_docker.rst) guide following the enhancements of the Dockerfile as per #172.

Since VSCode is not a necessity for building and running the Docker container, I removed all details regarding the use of VSCode from the docs.

### DIFF
--- a/docs/source/guides/developing_on_docker.rst
+++ b/docs/source/guides/developing_on_docker.rst
@@ -3,21 +3,39 @@
 Developing on Docker
 ======================
 
-.. note:: Assumes you have GPU on your system with the correct drivers installed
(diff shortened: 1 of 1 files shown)


ashwinvaidya17 (commented):
Thanks for updating this doc with a more thorough guide. However, I still feel we should have a section for those who want to use VS code remote container. We can just keep the section below.

LukasBommes:
Sure. I added the description on VSCode back into a subsection.

ashwinvaidya17 (approved):
Thanks!

Issue #180, opened 2022-03-31
Are masks needed if we use classification models like dfm ?


MhdKAT:
I tried to run dfm to classify my images as anomalous/non anomalous. I made my dataset in MVTec hierarchy but I get an error when i run the train script with the error being that the trainer can't access segmentation masks. Are they mandatory when the task is to classify? Below is a the error I get.   
File "C:\Users\hp\anomalib\env\lib\site-packages\torch\utils\data\_utils\fetch.py", line 44, in <listcomp>
    data = [self.dataset[idx] for idx in possibly_batched_index]
  File "c:\users\hp\anomalib\anomalib\data\mvtec.py", line 256, in __getitem__
    mask = cv2.imread(mask_path, flags=0) / 255.0
TypeError: unsupported operand type(s) for /: 'NoneType' and 'float'

samet-akcay:
@MhdKAT, it's the same issue that was found in `folder` dataset. Need to propose the same fixes to address the issue.
@LukasBommes, yes that's on our todo list, but didn't have a chance to work on it this week unfortunately.

MhdKAT:
I see ! Thanks for the clarification @LukasBommes,  @samet-akcay.

PR #220, opened 2022-04-11
assign test split for folder dataset

alexriedel1:
This PR addresses https://github.com/openvinotoolkit/anomalib/issues/180#issuecomment-1084564180

If you provide a `config.dataset.normal_test` to the datamodule, you can now define test data in a folder datamodule.


samet-akcay (changes requested):
Thanks for adding this @alexriedel1! It would also be great to cover `normal_dir` in the tests, but I guess we might need some refactoring in the tests to do so. We could potentially address this in a separate PR.

samet-akcay:
Thanks @alexriedel1, just in time! We're planning to release v0.2.6 today and upload to pypi. This will be a nice addition!

PR #221, opened 2022-04-11
Fix inconsistent benchmarking throughput/time

### LINKED ISSUE #185: Error while running benchmark.py on GPU
Running benchmark.py on Nvidia A100 SXM4  produces an error:

Exception: Error occurred while computing benchmark on device File "tools/benchmarking/benchmark.py", line 177, in distribute
    raise Exception(f"Error occurred while computing benchmark on device {job}") from exception
Exception: Error occurred while computing benchmark on device <Future at 0x7f05c18ac310 state=finished raised Exception>

Note: In my setup this error does not occur when running on CPU only. It does not occur either when running on CPU and GPU for patchcore model. It occurs when running on GPU for padim or cflow

ashwinvaidya17:
# Description

- Supress outputs from model and print values from benchmarking script

- Fixes #185 

## Changes

## Checklist

### DIFF
--- a/tools/benchmarking/benchmark.py
+++ b/tools/benchmarking/benchmark.py
@@ -15,10 +15,14 @@
 # and limitations under the License.
 
 
+import functools
+import io
 import logging
 import math
 import multiprocessing
+import sys
 import time
+import warnings
 from concurrent.futures import ProcessPoolExecutor, as_completed
 from pathlib import Path
 from tempfile import TemporaryDirectory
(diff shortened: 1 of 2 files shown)


tools/benchmarking/benchmark.py
```diff
@@ -218,6 +256,8 @@ def sweep(run_config: Union[DictConfig, ListConfig], device: int = 0, seed: int
 
     # Run benchmarking for current config
     model_metrics = get_single_model_metrics(model_config=model_config, openvino_metrics=convert_openvino)
+    print(model_config.model.name, model_config.dataset.category)
```
samet-akcay, line 259: What if we want to run `benchmark.py` for custom datasets that don't have `category` field?

tools/benchmarking/benchmark.py
```diff
@@ -41,9 +45,42 @@
     set_in_nested_config,
 )
 
+warnings.filterwarnings("ignore")
+
 logger = logging.getLogger(__file__)
+logging.getLogger("pytorch_lightning").setLevel(logging.ERROR)
+logging.getLogger("torchmetrics").setLevel(logging.ERROR)
+logging.getLogger("os").setLevel(logging.ERROR)
```
samet-akcay, line 53: maybe we could set the level in a for loop to avoid the duplicated lines

samet-akcay (changes requested):
Thanks, just minor stuff

djdameln (commented):
This works, but there will still be inconsistency in the throughput numbers between a normal training run and the benchmarking script. Maybe to avoid confusion, you could also add the batch size to the output in the timer callback Something like `Throughput (batch_size=16): 12 FPS`

djdameln (approved):

PR #234, opened 2022-04-14
Fix configs to properly use pytorch-lightning==1.6 with GPU

### LINKED ISSUE #232: GPU available but not used
Executing CFLOW model on Colab, PyTorch returns this

```
GPU available: True, used: False
TPU available: False, using: 0 TPU cores
IPU available: False, using: 0 IPUs
HPU available: False, using: 0 HPUs

GPU available but not used. Set "accelerator" and "devices" using Trainer(accelerator='gpu', devices=1).
```

Even if i'm setting gpus=1 on the config.yml of cflow.

 
**To Reproduce**
`train.py --model cflow
`

**GPU Model**
GPU 0: Tesla T4

**Versions**
```
pytorch-lightning             1.6.1
torch                         1.10.0+cu111
torchaudio                    0.10.0+cu111
torchmetrics                  0.7.3
torchsummary                  1.5.1
torchtext                     0.11.0
torchvision                   0.11.1+cu111
```

samet-akcay:
# Description

- Fix configs to properly use `pytorch-lightning==1.6.*` and train with GPU

- Fixes #232 

## Changes

## Checklist

### DIFF
--- a/anomalib/models/cflow/config.yaml
+++ b/anomalib/models/cflow/config.yaml
@@ -85,7 +85,6 @@ trainer:
   replace_sampler_ddp: true
   stochastic_weight_avg: false
   sync_batchnorm: false
-  terminate_on_nan: false
   tpu_cores: null
   track_grad_norm: -1
   val_check_interval: 1.0
--- a/anomalib/models/cflow/config.yaml
+++ b/anomalib/models/cflow/config.yaml
@@ -53,12 +53,16 @@ trainer:
   auto_scale_batch_size: false
   auto_select_gpus: false
   benchmark: false
+  callbacks: null
   check_val_every_n_epoch: 1
   default_root_dir: null
+  detect_anomaly: false
   deterministic: false
-  device: 1
+  device: null # Set automatically
   enable_checkpointing: true
+  enable_progress_bar: false
   fast_dev_run: false
+  gpus: null # Set automatically
   gradient_clip_val: 0
   limit_predict_batches: 1.0
   limit_test_batches: 1.0
@@ -78,15 +82,9 @@ trainer:
   overfit_batches: 0.0
   plugins: null
   precision: 32
-  prepare_data_per_node: true
-  process_position: 0
   profiler: null
-  progress_bar_refresh_rate: null
   replace_sampler_ddp: true
-  stochastic_weight_avg: false
   sync_batchnorm: false
   tpu_cores: null
   track_grad_norm: -1
   val_check_interval: 1.0
-  weights_save_path: null
-  weights_summary: top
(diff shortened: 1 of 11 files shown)


anomalib/models/cflow/config.yaml
```diff
@@ -55,10 +55,12 @@ trainer:
   benchmark: false
   check_val_every_n_epoch: 1
   default_root_dir: null
+  detect_anomaly: false
   deterministic: false
   enable_checkpointing: true
+  enable_progress_bar: true
   fast_dev_run: false
-  gpus: 1
+  gpus: null # Set automatically
```
ashwinvaidya17, line 63: Does it mean that pytorch lightning will distribute job across all available GPUs? In which case, we might have to look at the performance.

ashwinvaidya17 (approved):
Thanks! One minor comment but rest looks good

anomalib/models/cflow/config.yaml
samet-akcay: Not sure what's the best approach here. Setting it to 1 or automatically assigning it?
@ashwinvaidya17, @djdameln?

ashwinvaidya17: I would prefer using only a single gpu by default. I feel users should use distribute only if they know what they are doing. If we distribute the training then we will have to look at the learning rate and experiments might not be reproducible across different number of GPUs.

djdameln: What if no GPUs are available. Will it break or just switch to CPU training?

samet-akcay: > What if no GPUs are available. Will it break or just switch to CPU training?

It switches to CPU.

> I would prefer using only a single gpu by default. I feel users should use distribute only if they know what they are doing. If we distribute the training then we will have to look at the learning rate and experiments might not be reproducible across different number of GPUs.

As far as I understand, this doesn't set it to distributed training because `strategy` is still `null`. It automatically uses a single GPU.

ashwinvaidya17: In that case I am fine with how it is now.

ashwinvaidya17:
@aj2563 This is a known issue and is addressed in a different open PR. For now you can use `pip install torchmetrics==0.6.0` to make it work. Here is a colab link which uses this branch for training. https://colab.research.google.com/drive/18NciqtQwlrUIiuwWn8ld_Tp_5K6kgRPH?usp=sharing The first 7 cells are relevant in your case.

djdameln (approved):

ING3TP2:
Hello, I am having the error: Cannot import name 'configure_logger'  from 'anomalib.utils.loggers' while trying to run the following command : 
`python tools/train.py --config anomalib/models/padim/config.yaml`

image

PR #241, opened 2022-04-20
Add console logger

### LINKED ISSUE #54: Add proper logging and remove `print` statements with `logger` via `logger = logging.getLogger("pytorch_lightning")`
**Describe the bug**
- Currently `print` statements are used to log to console. Instead, `logger` should be used.
- Polish out pl loggings.

**Expected behavior**
- Modules should get access to the pytorch lightning logger via `logger = logging.getLogger("pytorch_lightning")` and log to the console.

- OS: [Ubuntu, OD] - any
- NVIDIA Driver Version [470.57.02] - any
- CUDA Version [e.g. 11.4] - any
- CUDNN Version [e.g. v11.4.120] - any
- OpenVINO Version [Optional e.g. v2021.4.2] - any

**Additional context**
- N/A

samet-akcay:
# Description

- Add python console logger and replace print

- Fixes #54 

## Changes

## Checklist


anomalib/utils/loggers/console.py
```diff
@@ -0,0 +1,110 @@
+"""Console Logger."""
+
+# Original Copyright (c) OpenMMLab.
```
ashwinvaidya17, line 3: I think this should say Anomalib. Also, should we change the date to 2022?

tools/train.py
```diff
@@ -39,6 +40,7 @@ def get_args() -> Namespace:
     parser = ArgumentParser()
     parser.add_argument("--model", type=str, default="padim", help="Name of the algorithm to train/test")
     parser.add_argument("--model_config_path", type=str, required=False, help="Path to a model config file")
+    parser.add_argument("--log-level", type=str, help="<DEBUG, INFO, WARNING, ERROR>")
```
ashwinvaidya17, line 43: Is this `log-level` passed to the console logger somewhere?

ashwinvaidya17 (commented):
Looks good! Just two minor comments.

anomalib/utils/loggers/console.py
samet-akcay: no, the code has been modified based on OpenMMLab, so we give the credit.

tools/train.py
samet-akcay: Not yet, it's a bit more involved than we thought. It's something I was going to bring up. This could be either addressed in a separate PR, or this one needs to be modified a little bit.

anomalib/utils/loggers/console.py
```diff
@@ -0,0 +1,110 @@
+"""Console Logger."""
```
djdameln, line 1: I'm not sure I completely understand why this file is needed. Can't we just specify the format using 
`logging.basicConfig(format="%(asctime)s - %(name)s - %(levelname)s - %(message)s", level=logging.INFO)`
in the entrypoint scripts, and create the individual loggers in each file with `logging.getLogger(__name__)`?

anomalib/utils/loggers/__init__.py
```diff
@@ -25,15 +26,49 @@
...
+]
 AVAILABLE_LOGGERS = ["tensorboard", "wandb", "csv"]
 
 
 class UnknownLogger(Exception):
     """This is raised when the logger option in `config.yaml` file is set incorrectly."""
 
 
-def get_logger(
+def get_console_logger(name: Optional[str] = None, level: Union[int, str] = logging.INFO) -> Logger:
```
djdameln, line 42: Maybe we could rename this `configure_logger` and instead of having it return the logger, just define `logger = logging.getLogger("anomalib")` in `train.py` and the other entrypoitns. That way the logging in those entrypoint files would be more in line with the other submodules.

djdameln (changes requested):

anomalib/utils/loggers/__init__.py
samet-akcay: This would work if the existing `logger` were renamed to something like `experiment_logger` or similar

samet-akcay: Do you mean something like the following in `train.py` and other entrypoints?

```
...
configure_logger()
logger = logging.getLogger("anomalib")
...
```

If yes, this would cost us two lines for the sake of following the standard logging definition.

We could maybe rename this to `get_logger` instead, which would be a one liner in this case. And the experiment logger would also be renamed accordingly.
```
...
logger = get_logger("anomalib", level=logging.INFO)
...
experiment_logger = get_experiment_logger(config)
```

djdameln: > If yes, this would cost us two lines for the sake of following the standard logging definition.

It would add a single line `logger = logging.getLogger("anomalib")`, but this line would be placed at the top of the file, right under the imports. So it may be an additional line, but it does not clutter the `train` function. 

Also this same line would be removed from the `get_logger`/`configure_logger` function, which would end up a lot cleaner, because it would also only require a single argument (`log_level`), no logic for setting the logger name, and no return type.

djdameln (approved):
Thanks!

Issue #240, opened 2022-04-19
`log_graph = True` for tensorboard logger doesn't show model graph.


innat:
**Describe the bug**
I wanted to show the computational graph of the `patchcore` model and for that I enable `project.logger: tensorboard` and set `log_graph = True` 

```python
logger = AnomalibTensorBoardLogger(
            name="Tensorboard Logs",
            save_dir=os.path.join(config.project.path, "logs"),
            log_graph=True
        )
```

Now, when I finished training, I run the tensorboard and under the `GRAPHS` tag, it showed nothing. 

![image](https://user-images.githubusercontent.com/17668390/164081646-1f8d32d4-13d4-4636-a2e0-57ab9fd91eb2.png)

**To Reproduce**
1. Go to the `patchcore` config file and set `logger: tensorboard'
2. Next, `working/anomaly_detection_engine/anomalib/anomalib/utils/loggers/__init__.py` and set `log_graph=True` 
3. Run the anomalib with patchcore config
4. After training, run the event log in tensorboard.

**Expected behavior**
- It should give a computational graph, like https://github.com/PyTorchLightning/pytorch-lightning/issues/2915

**Screenshots**
- (Mention above)

**Hardware and Software Configuration**
- OS: [Ubuntu, OD] Ubuntu
- PyTorch-Lighting: '1.5.9'
- NVIDIA Driver Version [470.57.02]
- CUDA Version [e.g. 11.4] 10.2
- CUDNN Version [e.g. v11.4.120] 7605
- OpenVINO Version [Optional e.g. v2021.4.2]

**Additional context**
- I also found that the [`add_image`](https://github.com/openvinotoolkit/anomalib/blob/4910e186925637ae3724c7c23f200494521552cc/anomalib/utils/loggers/tensorboard.py#L84) function of the tensorboard callback, doesn't write any image on the tensorboard either.
- Is there any way, I can write the **Histogram** in tensorboard in pytorch-lighting. Histograms are made for weights and bias matrices in the network. They tell us about the distribution of weights and biases among themselves. Is there any convenient way to add it to anomalib' models?

innat:
@samet-akcay could you please take a look at this issue? 
I tried this [fix](https://github.com/openvinotoolkit/anomalib/pull/233) but ended up with this [issue](https://github.com/openvinotoolkit/anomalib/issues/237).

innat:
[...]

Issue #227, opened 2022-04-13
Create a section for Logging and HPO in `README.md` file.


samet-akcay:
**Is your feature request related to a problem? Please describe.**
  - Currently `README.md` file lacks the instructions of how to run tensorboard and wandb logging as well as HPO runs. 

**Describe the solution you'd like**
  - Adding the CLI commands to run these and screenshot for each would be really useful for the newcomers.

PR #280, opened 2022-04-26
📃Update documentation

ashwinvaidya17:
# Description

Update documentation

- [x] Readme
- [x] HPO - need to add screenshot
- [x] Logger
- [x] Benchmarking
- [ ] Other artefacts in logger
- [ ] Possibly activation maps?
- [x] Proposed new design for logging graph

# Fixes
- #240
- #227 
- #260


docs/source/guides/benchmarking.md
```diff
@@ -0,0 +1,33 @@
...
+      - capsule
+      - carpet
+    image_size: [224]
+  model_name:
+    - padim
+    - patchcore
+```
+
+This configuration computes the throughput and performance metrics on CPU and GPU for four categories of the MVTec dataset for Padim and PatchCore models. The dataset can be configured in the respective model configuration files. By default, `compute_openvino` is set to `False` to support instances where OpenVINO requirements are not installed in the environment. Once installed, this flag can be set to True to get throughput on OpenVINO optimized models. The `writer` parameter is optional and can be set to `writer: []` in case the user only requires a csv file without logging to each respective logger. It is a good practice to set a value of seed to ensure reproducibility across runs and thus, is set to a non-zero value by default.
+
```
samet-akcay, line 28: This config shows the benchmarking on certain categories on MVTec. A user might want to try it on a custom dataset. In this case, it would be good to mention how the benchmarking could be used with a folder dataset.

docs/source/guides/logging.md
```diff
@@ -0,0 +1,65 @@
+# Logging
+
+Anomalib offers various mechanisms for logging metrics and predicted masks.
+
+## Enabling Logging
+
+These can be enabled using the `logger` parameter in `project` section of each model configuration file. The available options are `tensorboard`, `wandb` and `csv`.
+For example, to log to TensorBoard:
```
samet-akcay, line 8: this PR changes this right? `logger` parameter will be in `logging` section of the config file?

docs/source/guides/logging.md
```diff
@@ -0,0 +1,65 @@
...
+```yaml
+logger: "tensorboard"
+```
+
+ You can also pass a list of loggers to enable multiple loggers. For example:
+```yaml
+project:
```
samet-akcay, line 15: `logging`?

docs/source/guides/logging.md
```diff
@@ -0,0 +1,65 @@
...
+  logger:
+    - tensorboard
+    - wandb
+    - csv
+```
+
+## Logging Images
+
+Anomalib allows you to save predictions to the disc by setting `log_images_to: local`. As of the current version, Anomalib also supports TensorBoard and WandB loggers for logging images. These loggers extend upon the base loggers by providing a common interface for logging images. You can access the required logger from `trainer.loggers`. Then you can use `logger.add_image` method to log images. For a complete overview of this method. Refer to our [API documentation](https://openvinotoolkit.github.io/anomalib/api/anomalib/utils/loggers/index.html).
```
samet-akcay, line 24: ~~method. Refer~~ ✍️. for a complete overview of this method, refer to our ...

anomalib/models/cflow/config.yaml
```diff
@@ -49,8 +49,11 @@ metrics:
 project:
   seed: 0
   path: ./results
-  log_images_to: [local]
-  logger: false # options: [tensorboard, wandb, csv] or combinations.
+
+logging:
+  log_images_to: ["local"] # options: [wandb, tensorboard, local]. Make sure you also set logger with using wandb or tensorboard.
```
samet-akcay, line 54: I think we should decide how to handle `log_images_to` parameter before merging this PR. It could be annoying to the user if we decide to change it again in the future. @djdameln, @ashwinvaidya17

samet-akcay (changes requested):

anomalib/models/cflow/config.yaml
djdameln: I think it will take some time to change this. We'll need a design first and then implement the changes. So I would suggest to merge this first to prevent delays.

docs/source/guides/logging.md
ashwinvaidya17: Yeah. I'll update the text if you guys are satisfied with the new design.

docs/source/guides/benchmarking.md
```diff
@@ -0,0 +1,33 @@
...
+      - cable
+      - capsule
+      - carpet
+    image_size: [224]
+  model_name:
+    - padim
+    - patchcore
+```
+
+This configuration computes the throughput and performance metrics on CPU and GPU for four categories of the MVTec dataset for Padim and PatchCore models. The dataset can be configured in the respective model configuration files. By default, `compute_openvino` is set to `False` to support instances where OpenVINO requirements are not installed in the environment. Once installed, this flag can be set to True to get throughput on OpenVINO optimized models. The `writer` parameter is optional and can be set to `writer: []` in case the user only requires a csv file without logging to each respective logger. It is a good practice to set a value of seed to ensure reproducibility across runs and thus, is set to a non-zero value by default.
```
djdameln, line 27: > The dataset can be configured in the respective model configuration files.

This may be a bit confusing. What if there's a difference in the dataset sections of the different model configs. It would probably be better to set the dataset settings from the benchmarking config (this would of course require a separate PR).

docs/source/guides/hyperparameter_optimization.md
```diff
@@ -0,0 +1,48 @@
+# Hyperparameter Optimization
```
djdameln, line 1: Maybe you could add a link to the wandb documentation on HPO somewhere on this page for further reference

anomalib/utils/callbacks/graph.py
```diff
@@ -0,0 +1,52 @@
...
+
+        for logger in trainer.loggers:
+            if isinstance(logger, AnomalibWandbLogger):
+                # NOTE: log graph gets populated only after one backward pass. This won't work for models which do not
+                # require training such as Padim
+                logger.watch(pl_module, log_graph=True, log="all")
+                break
+
+    def on_train_end(self, trainer: Trainer, pl_module: LightningModule) -> None:
+        """Unwatch model if configured for wandb.
```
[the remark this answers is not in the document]
ashwinvaidya17: I forgot to add the `logger.unwatch` line for wandb logger here. We have to unwatch model for wandb. For models such as patchcore, it needs be trained first before the complete graph is generated, and the user needs to call `logger.log_graph` from the code, I decided we could combine both in a callback which handles logging and unwatch after training completes. While we can add `logger.watch` when we are initializing the wandb logger I am not sure which place would be ideal for `unwatch` and `log_graph`.

docs/source/guides/hyperparameter_optimization.md
```diff
@@ -0,0 +1,47 @@
...
+## Running HPO
+
+:::{Note}
+You will need to have logged into a wandb account to use HPO search and view the results.
+:::
+
+To run the hyperparameter optimization, use the following command:
+
+```bash
+python tools/hpo/wandb_sweep.py --model padim --model_config ./path_to_config.yaml --sweep_config tools/hpo/sweep.yaml
```
samet-akcay, line 35: would be good to break it down to multiple lines for smaller screens.

```bash
python tools/hpo/wandb_sweep.py \
    --model padim --model_config ./path_to_config.yaml \
    --sweep_config tools/hpo/sweep.yaml
```

samet-akcay (changes requested):

samet-akcay (commented):
I think it's in a good shape to go. one final question would be whether it would be an idea to use gifs to show tensorboard and wandb stuff. What do you think?

ashwinvaidya17:
> > I think it's in a good shape to go. one final question would be whether it would be an idea to use gifs to show tensorboard and wandb stuff. What do you think?
> 
> gifs are a good idea. I'll see if I can create gifs from a few clips of the videos

I'll leave this for a different PR

djdameln (approved):
Thanks. All my comments have been addressed

samet-akcay (approved):

Issue #275, opened 2022-04-25
Add Gradio Support


julien-blanchon:
Gradio App is an open source machine learning interface running on top of Python : https://gradio.app/

It's quite easy to implement and very hackable.
It's also compatible with Hugging Face Space (https://huggingface.co, https://huggingface.co/spaces).

I may help user and researcher to:

- Quickly debug their model (infer without any line of code, provide web drawing tool ...)
- Showcase their work
- Share model with the community (ie Hugging Face Space ...)
- Flagging unexpected behavior

![gradio](https://user-images.githubusercontent.com/11278197/165093206-ccb80241-690a-4ada-a218-ed86edb18856.png)
![gradio_draw1](https://user-images.githubusercontent.com/11278197/165093211-8c595ecb-3431-4b55-854e-52c4744d95d7.png)
![gradio_draw2](https://user-images.githubusercontent.com/11278197/165093212-8558d2e2-2855-4f9c-b63f-d0bea6fe9f19.png)

julien-blanchon:
I've implemented a quick example with my EuroSat Dataset here: https://github.com/julien-blanchon/anomalib/tree/gradio
It may for now only work with `segmentation`

ashwinvaidya17:
Great stuff! It was on our todo list.

julien-blanchon:
I don't have the right to create a new branch here, I don't feel comfortable enough to request a merge with current project branch https://github.com/julien-blanchon/anomalib/tree/gradio

ashwinvaidya17:
@julien-blanchon feel free to create a PR from your fork. We squash commits before merging in case that's what you are concerned about. We can always polish the rest in the PR review.

PR #283, opened 2022-04-26
Add Gradio

julien-blanchon:
# Add Simple Gladio support

Add new tool `inference_gladio.py` for gladio integration

```raw
optional arguments:
  -h, --help            show this help message and exit
  --config CONFIG       Path to a model config file
  --weight_path WEIGHT_PATH
                        Path to a model weights
  --threshold THRESHOLD
                        Value to threshold anomaly scores into 0-100 range
```

## Example

`python tools/inference_gradio.py --config ./anomalib/models/padim/config.yaml --weight_path ./results/padim/mvtec/bottle/weights/model.ckpt`

## Changes

- [x] Add Gladio Support
- [x] Add gradio>=2.9.4 dependencies

## Checklist


julien-blanchon:
https://github.com/openvinotoolkit/anomalib/issues/275

tools/inference_gradio.py
```diff
@@ -0,0 +1,126 @@
...
+
+from anomalib.config import get_configurable_parameters
+from anomalib.deploy.inferencers.base import Inferencer
+from anomalib.post_processing import compute_mask, superimpose_anomaly_map
+
+
+def infer(
+    image: np.ndarray, inferencer: Inferencer, threshold: float = 50.0
+) -> Tuple[np.ndarray, float, np.ndarray, np.ndarray, np.ndarray]:
+    """Inference fonction, return anomaly map, score, heat map, prediction mask ans visualisation.
```
ashwinvaidya17, line 26: Minor issue but there is a typo at `function`.
Also, we are using Google format as our documentation style. So instead of 
```
:param image: image
:type image: np.ndarray
```
It should be
```
image: (np.ndarray): Image
```

ashwinvaidya17 (changes requested):
Thanks for this contribution! I have a minor comment. This would help with the checks: pre-commit https://openvinotoolkit.github.io/anomalib/guides/using_pre_commit.html

tools/inference_gradio.py
```diff
@@ -0,0 +1,126 @@
...
+    )
+
+    parser.add_argument("--share", type=bool, required=False, default=False, help="Share Gradio `share_url`")
+
+    args = parser.parse_args()
+
+    return args
+
+
+def get_inferencer(gladio_args: Union[DictConfig, ListConfig]) -> Inferencer:
```
samet-akcay, line 72: I guess `gladio_args` should be `gradio_args`?

samet-akcay: I like the idea to create `get_inferencer` function. It would be nice if the function args would be explicit rather than passing args all together. For example, something like:

```
def get_inferencer(config_path: Path, weight_path: Path, meta_data_path: Path) -> Inferencer:
```

This way, we could even use this for `inference.py` (In a separate PR, of course :) )

tools/inference_gradio.py
```diff
@@ -0,0 +1,126 @@
...
+            f"OpenVINO Inferencer expects either .onnx, .bin or .xml file. Got {extension}"
+        )
+
+    return inferencer
+
+
+if __name__ == "__main__":
+    session_args = get_args()
+
+    gladio_inferencer = get_inferencer(session_args)
```
samet-akcay, line 105: `gladio_inferencer` to `gradio_inferencer`?

tools/inference_gradio.py
```diff
@@ -0,0 +1,126 @@
...
+
+    iface = gr.Interface(
```
samet-akcay, line 107: Would you mind renaming `iface` to `interface`?

samet-akcay (changes requested):
Thanks @julien-blanchon! This is a great addition! 🔥

I only have a few comments, hopefully wouldn't require so much of your time. If yes, we could do those changes as well. Thanks!

tools/inference_gradio.py
julien-blanchon: Yes gradio of course.

About the `get_inferencer`, yes if used out of the scope of gradio it may be usefull

tools/inference_gradio.py
julien-blanchon: :+1:

tools/inference_gradio.py
julien-blanchon: `iface` is the default interface name of the gradio doc, I don't know how strong this convention is but yes interface is more explicit

julien-blanchon:
I will make a new commit this tomorrow with some patch

julien-blanchon:
I've just push a typo fix, feel free to reformat `get_inferencer` and to move it as an utils.

julien-blanchon:
The `infer` function may also be moved outside and reused.

Maybe a better description and interface could be done.

samet-akcay:
> The infer function may also be moved outside and reused.
Maybe a better description and interface could be done.

@julien-blanchon thanks a lot! Yeah, I agree, we could do another round of refactoring to re-use some of these across other inference stuff. This is the scope of another PR that we could create in the future, though.

Really appreciate your efforts here. Thanks again!

samet-akcay (approved):

ashwinvaidya17 (approved):

PR #315, opened 2022-05-14
Refactor `AnomalyModule` and `LightningModules` to explicitly define class arguments.

samet-akcay:
# Description

- This PR refactors `AnomalyModule` and `LightningModules` to explicitly define class arguments. Current design depends on `hparams` variable, which is a `DictConfig`. Despite being convenient, it is not explicit such that the user may not now what exactly needed to construct a model class. The proposed implementation addresses this explicitly construct the model classes.

## Changes

## Checklist


anomalib/utils/callbacks/metrics.py
```diff
@@ -0,0 +1,67 @@
+"""Metrics Configuration Callback."""
+
+# Copyright (C) 2020 Intel Corporation
```
ashwinvaidya17, line 3: Should we make it 2022?

anomalib/utils/callbacks/tiler.py
```diff
@@ -0,0 +1,83 @@
...
+__all__ = ["TilerCallback"]
+
+
+class TilerCallback(Callback):
+    """Tiler Configuration Callback."""
+
+    def __init__(
+        self,
+        enable: bool,
+        tile_size: Union[int, Sequence],
```
ashwinvaidya17, line 35: I think we should have a default value here for the cases when enable is false. We wouldn't have to care about passing a parameter here.

anomalib/models/cflow/lightning_model.py
```diff
@@ -159,3 +166,54 @@ def validation_step(self, batch, _):  # pylint: disable=arguments-differ
...
+        """
+        decoders_parameters = []
+        for decoder_idx in range(len(self.model.pool_layers)):
+            decoders_parameters.extend(list(self.model.decoders[decoder_idx].parameters()))
+
+        optimizer = optim.Adam(
+            params=decoders_parameters,
+            lr=self.hparams.model.lr,
+        )
+        return optimizer
```
ashwinvaidya17, line 219: Since the idea is to deprecate this class with hparams in the future, I think the member functions should be in the base class. That way we can just delete the init function.

anomalib/models/ganomaly/lightning_model.py
```diff
@@ -183,3 +185,56 @@ def _normalize(self, scores: Tensor) -> Tensor:
...
+            extra_layers=hparams.model.extra_layers,
+            add_final_conv_layer=hparams.model.add_final_conv,
+            wadv=hparams.model.wadv,
+            wcon=hparams.model.wcon,
+            wenc=hparams.model.wenc,
+        )
+        self.hparams: Union[DictConfig, ListConfig]  # type: ignore
+        self.save_hyperparameters(hparams)
+
+    def configure_callbacks(self):
```
ashwinvaidya17, line 215: Same here. Can we move this to the base class?

ashwinvaidya17 (commented):
Thanks for the efforts! I have a few minor comments.

anomalib/models/cflow/lightning_model.py
djdameln: I agree. There seems to be some inconsistency as well. Some classes have the methods in the base class, others in the Lightning class, yet others in a mixture of both.

anomalib/models/components/base/anomaly_module.py
```diff
@@ -36,30 +35,41 @@ class AnomalyModule(pl.LightningModule, ABC):
...
+            Defaults to None.
     """
 
-    def __init__(self, params: Union[DictConfig, ListConfig]):
-
+    def __init__(
+        self,
+        adaptive_threshold: Optional[bool] = None,
+        default_image_threshold: Optional[float] = None,
+        default_pixel_threshold: Optional[float] = None,
```
djdameln, line 50: Would it be an idea to initialize these in the metrics callback instead of here?

anomalib/models/components/base/anomaly_module.py
```diff
@@ -36,30 +35,41 @@ class AnomalyModule(pl.LightningModule, ABC):
...
         self.callbacks: List[Callback]
 
-        self.image_threshold = AdaptiveThreshold(self.hparams.model.threshold.image_default).cpu()
-        self.pixel_threshold = AdaptiveThreshold(self.hparams.model.threshold.pixel_default).cpu()
+        self.adaptive_threshold = False if adaptive_threshold is None else adaptive_threshold
+        self.default_image_threshold = 0.0 if default_image_threshold is None else default_image_threshold
+        self.default_pixel_threshold = 0.0 if default_pixel_threshold is None else default_pixel_threshold
+
+        self.image_threshold = AdaptiveThreshold(self.default_image_threshold).cpu()
+        self.pixel_threshold = AdaptiveThreshold(self.default_pixel_threshold).cpu()
```
djdameln, line 63: So these would be moved to the callback, and here in the base class we just create placeholders like we do for the MetricCollections below.

anomalib/utils/callbacks/cdf_normalization.py
```diff
@@ -38,8 +38,11 @@ def __init__(self):
 
     def on_test_start(self, _trainer: pl.Trainer, pl_module: AnomalyModule) -> None:
         """Called when the test begins."""
-        pl_module.image_metrics.set_threshold(0.5)
-        pl_module.pixel_metrics.set_threshold(0.5)
+        if isinstance(pl_module, AnomalyModule):
```
djdameln, line 41: Why is this needed? `pl_module` will always be an instance of AnomalyModule no?

anomalib/models/dfkde/config.yaml
```diff
@@ -20,6 +20,8 @@ model:
   confidence_threshold: 0.5
   pre_processing: scale
   n_components: 16
+  threshold_steepness: 0.05
+  threshold_offset: 12
```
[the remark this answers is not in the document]
djdameln: https://github.com/openvinotoolkit/anomalib/issues/325

anomalib/models/cflow/lightning_model.py
djdameln: Ah, that explains it then. Thanks for clarifying :)

Maybe you could add this info to the docstring of the to-be-deprecated lightning classes to avoid confusion?

anomalib/models/components/base/anomaly_module.py
samet-akcay: would be good. With a quick try, it gave some issues when loading models. Unfortunately, I don't think I will enough time today to investigate what's wrong

ashwinvaidya17 (approved):
Thanks! I am fine with the changes.

samet-akcay:
@djdameln, if you are happy with the final changes we could merge this PR today

djdameln (approved):
Great work, thanks!

PR #318, opened 2022-05-16
🐞 Increase tolerance + nightly path fix

ashwinvaidya17:
# Description

- Increase tolerance for nightly tests
- Fix path for benchmarking tests


djdameln (approved):
Instead of allowing a tolerance, we could also just lower the threshold. This would be a bit more readable (just an idea, don't need to change it now).

samet-akcay (approved):

PR #332, opened 2022-05-24
🐞 Fix inferencer in Gradio

### LINKED ISSUE #296: Gradio Error
running :
python tools/inference_gradio.py --config ./anomalib/models/patchcore/config.yaml --weight_path ./results/patchcore/mvtec/zipper/weights/model.ckp

error:
  File "tools/inference_gradio.py", line 105, in <module>
    gradio_inferencer = get_inferencer(session_args.config, session_args.weight_path, session_args.meta_data)
  File "tools/inference_gradio.py", line 74, in get_inferencer
    config = get_configurable_parameters(config_path)
  File "/home/rumanto/AI_Development/anomalib/anomalib/config/config.py", line 142, in get_configurable_parameters
    config = OmegaConf.load(config_path)
  File "/home/rumanto/miniconda3/envs/anomalib_env/lib/python3.8/site-packages/omegaconf/omegaconf.py", line 183, in load
    with io.open(os.path.abspath(file_), "r", encoding="utf-8") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/rumanto/AI_Development/anomalib/anomalib/models/anomalib/models/patchcore/config.yaml/config.yaml'

any one same result?

ashwinvaidya17:
# Description

- Passes correct parameter to Inferencers.

- Fixes #296 

## Changes

## Checklist

### DIFF
--- a/tools/inference_gradio.py
+++ b/tools/inference_gradio.py
@@ -6,13 +6,12 @@ This script provide a gradio web interface
 from argparse import ArgumentParser, Namespace
 from importlib import import_module
 from pathlib import Path
-from typing import Tuple, Union
+from typing import Optional, Tuple, Union
 
 import gradio as gr
 import gradio.inputs
 import gradio.outputs
 import numpy as np
-from omegaconf import DictConfig, ListConfig
 from skimage.segmentation import mark_boundaries
 
 from anomalib.config import get_configurable_parameters
@@ -46,11 +45,18 @@ def infer(
 def get_args() -> Namespace:
     """Get command line arguments.
 
+    Example:
+
+        >>> python tools/inference_gradio.py --model ./anomalib/models/padim/config.yaml --weight_path ./results/padim/mvtec/bottle/weights/model.ckpt
+        >>> python tools/inference_gradio.py --model padim --weight_path ./results/padim/mvtec/bottle/weights/model.ckpt
+
     Returns:
         Namespace: List of arguments.
     """
     parser = ArgumentParser()
-    parser.add_argument("--config", type=Path, required=True, help="Path to a model config file")
+    parser.add_argument(
+        "--model", type=str, required=True, help="Can be either the model name or the path to a model config file"
+    )
     parser.add_argument("--weight_path", type=Path, required=True, help="Path to a model weights")
     parser.add_argument("--meta_data", type=Path, required=False, help="Path to JSON file containing the metadata.")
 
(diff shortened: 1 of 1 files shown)


tools/inference_gradio.py
```diff
@@ -46,11 +45,18 @@ def infer(
...
+    Example:
+
+        >>> python tools/inference_gradio.py --model ./anomalib/models/padim/config.yaml --weight_path ./results/padim/mvtec/bottle/weights/model.ckpt
+        >>> python tools/inference_gradio.py --model padim --weight_path ./results/padim/mvtec/bottle/weights/model.ckpt
+
     Returns:
         Namespace: List of arguments.
     """
     parser = ArgumentParser()
-    parser.add_argument("--config", type=Path, required=True, help="Path to a model config file")
```
samet-akcay, line 53: Maybe we could keep `config` to be consistent with other entrypoints?

tools/inference_gradio.py
```diff
@@ -66,29 +72,44 @@ def get_args() -> Namespace:
...
     # Get the inferencer. We use .ckpt extension for Torch models and (onnx, bin)
     # for the openvino models.
     extension = weight_path.suffix
     inferencer: Inferencer
     if extension in (".ckpt"):
         module = import_module("anomalib.deploy.inferencers.torch")
         TorchInferencer = getattr(module, "TorchInferencer")  # pylint: disable=invalid-name
-        inferencer = TorchInferencer(
-            config=config, model_source=weight_path, meta_data_path=meta_data
-        )
```
samet-akcay, line 84: Is this needed to correct the parameters. If `TorchInferencer` and `OpenVINOInferencer` are explicitly imported, we wouldn't have to disable the pylint, no?

samet-akcay (changes requested):

tools/inference_gradio.py
ashwinvaidya17: I wasn't sure what to call it. We can either pass the model config or the model name now. If I switch back to config should I drop calling inferencer with just model name?

samet-akcay: `--config` would be inline with the new PL CLI, so I would prefer that.

> If I switch back to config should I drop calling inferencer with just model name?

Not sure if I get this part

ashwinvaidya17: What I mean is that initially with `config` parameter we had to pass the `yaml` file to the inferencer. I changed it so that we can either pass `yaml` or only the model name. So, I wasn't sure sure what to call this parameter. If I change it back to `config` then passing just the model name might not match with the parameter name. In which case we can drop passing only the model name and keep `yaml` as the only option. It might be better in some sense as it will force people to ensure that their train config matches with the config they use for inference. Otherwise inferencer might pick up the default config with just the model name which might not match with the path for config that was used to train with.

samet-akcay: I agree. Passing only config would ensure the right config file is passed. Otherwise, it would be the default config, which may not be the same as the one that is used to train the model.

tools/inference_gradio.py
```diff
@@ -47,16 +47,15 @@ def get_args() -> Namespace:
...
     Returns:
         Namespace: List of arguments.
     """
     parser = ArgumentParser()
-    parser.add_argument(
-        "--model", type=str, required=True, help="Can be either the model name or the path to a model config file"
-    )
+    parser.add_argument("--config", type=Path, required=True, help="Path to a model config file")
     parser.add_argument("--weight_path", type=Path, required=True, help="Path to a model weights")
     parser.add_argument("--meta_data", type=Path, required=False, help="Path to JSON file containing the metadata.")
```
samet-akcay, line 60: there is a bit of inconsistency in naming. In fact, this is the case for other entrypoints. I think we should stick to one of the following

```
- config, weights, meta_data
- config_path, weight_path, meta_data_path
```

ashwinvaidya17: That's alright. The whole point of the review is to ensure code quality

djdameln (approved):
I'm fine with these changes, but I guess we'll need to refactor gradio inference once we switch to the PL inferencer in https://github.com/openvinotoolkit/anomalib/pull/298. This should probably be included in https://github.com/openvinotoolkit/anomalib/pull/298 before merging.

PR #340, opened 2022-05-31
Upgrade wandb version

ashwinvaidya17:
# Description

- Upgrade wandb version
- Fix tests to work with new version


samet-akcay (approved):
Thanks, lgtm

Issue #339, opened 2022-05-31
Add DRAEM Model


djdameln:
Add DRAEM model implementation
https://paperswithcode.com/paper/draem-a-discriminatively-trained

PR #344, opened 2022-06-02
Add DRAEM Model

djdameln:
# Description

- Adds model implementation of [DRAEM](https://arxiv.org/pdf/2108.07610v2.pdf).

- Fixes https://github.com/openvinotoolkit/anomalib/issues/339

## Known issues:

- Benchmarking results are not yet available. This algorithm is pretty slow (a few hours per training run), so it will take some time before I manage to collect the numbers for all categories.
- I'm using borrowed code for the Perlin noise generation. This module is a bit messy and will need to be refactored in the future.

## Questions/Discussion:

- I feel that the augmenter class in `anomalib/models/draem/augmenter.py` could be useful for other models as well. We could consider moving it to `components`
- I ran into some problems with the pre-processor, because it assumes that we always want to normalize the input images to the ImageNet stats. For DRAEM this was not the case. It would be good to revisit this.

## Changes

## Checklist


anomalib/models/draem/augmenter.py
```diff
@@ -0,0 +1,150 @@
...
+            iaa.pillike.EnhanceSharpness(),
+            iaa.AddToHueAndSaturation((-50, 50), per_channel=True),
+            iaa.Solarize(0.5, threshold=(32, 128)),
+            iaa.Posterize(),
+            iaa.Invert(),
+            iaa.pillike.Autocontrast(),
+            iaa.pillike.Equalize(),
+            iaa.Affine(rotate=(-45, 45)),
+        ]
+        self.rot = iaa.Sequential([iaa.Affine(rotate=(-90, 90))])
```
samet-akcay, line 53: Is it possible to use `albumentations` library instead of `imgaug`? They are similar libraries, would be good to use one of them for consistency

djdameln: I agree. I tried rewriting this with albumentations, but some of these transforms are not available in the albumentations package. For reproducibility I decided to keep imgaug for now. I don't expect it to lead to significant performance drop if we omit or replace some of these transforms, but it would be good to verify this first.

anomalib/models/draem/augmenter.py
```diff
@@ -0,0 +1,150 @@
...
+
+        # Apply perturbations batch wise
+        beta = torch.rand(batch_size) * 0.8
+        beta = beta.view(batch_size, 1, 1, 1).expand_as(batch).to(batch.device)
+
+        augmented_batch = batch * (1 - masks) + (1 - beta) * perturbations + beta * batch * (masks)
+
+        # for i in range(batch_size):
+        #     cv2.imshow("aug", augmented_batch[i].permute(1, 2, 0).cpu().numpy())
+        #     cv2.waitKey(0)
```
samet-akcay, line 148: Is this a leftover?

anomalib/models/draem/augmenter.py
```diff
@@ -0,0 +1,150 @@
...
+
+    Args:
+        anomaly_source_path (Optional[str]): Path to a folder of images that will be used as source of the anomalous
+        noise. If not specified, random noise will be used instead.
+    """
+
+    def __init__(self, anomaly_source_path: Optional[str] = None):
+
+        if anomaly_source_path is not None:
+            self.anomaly_source_paths = sorted(glob.glob(anomaly_source_path + "/**/*.jpg", recursive=True))
```
samet-akcay, line 37: Not sure if I understand this. Why does this search for jpg images?

anomalib/models/draem/config.yaml
```diff
@@ -0,0 +1,102 @@
...
+    apply: false
+    tile_size: null
+    stride: null
+    remove_border_count: 0
+    use_random_tiling: False
+    random_tile_count: 16
+
+model:
+  name: draem
+  anomaly_source_path: null # optional, e.g. ./datasets/dtd
```
samet-akcay, line 25: What is dtd? Does this make `dataset` section above obsolete?

anomalib/models/draem/lightning_model.py
```diff
@@ -0,0 +1,125 @@
+"""DRÆM – A discriminatively trained reconstruction embedding for surface anomaly detection.
+
+Paper https://arxiv.org/abs/2108.07610
+"""
+
+# Copyright (C) 2020 Intel Corporation
```
samet-akcay, line 6: 2022. Also same as above. i'd prefer using the shorter license header

anomalib/models/draem/lightning_model.py
```diff
@@ -0,0 +1,125 @@
...
+        input_image = batch["image"]
+        # Apply corruption to input image
+        augmented_image, anomaly_mask = self.augmenter.augment_batch(input_image)
+        # Generate model prediction
+        reconstruction, prediction = self.model(augmented_image)
+        # Compute loss
+        l2_loss = self.l2_loss(reconstruction, input_image)
+        ssim_loss = self.ssim_loss(reconstruction, input_image)
+        focal_loss = self.focal_loss(prediction, anomaly_mask)
+        loss = l2_loss + ssim_loss + focal_loss
```
samet-akcay, line 78: Would it be possible to have a Loss class similar to other algos, where we would have all these different losses as attributes?

anomalib/models/draem/loss.py
```diff
@@ -0,0 +1,204 @@
...
+from math import exp
+
+import numpy as np
+import torch
+import torch.nn.functional as F
+from torch import Tensor, nn
+
+
+class FocalLoss(nn.Module):
+    """Copy from: https://github.com/Hsuxu/Loss_ToolBox-PyTorch/blob/master/FocalLoss/FocalLoss.py.
```
samet-akcay, line 23: Please keep in mind that, for each 3rd party component we use, we need to add to `third-party-components.txt` in the main root

anomalib/models/draem/loss.py
```diff
@@ -0,0 +1,204 @@
...
+
+from math import exp
+
+import numpy as np
+import torch
+import torch.nn.functional as F
+from torch import Tensor, nn
+
+
+class FocalLoss(nn.Module):
```
samet-akcay, line 22: Would be good to annotate the arguments here in the methods of this class.

samet-akcay (changes requested):

anomalib/models/draem/loss.py
ashwinvaidya17, line 22: Have you experimented with torchvision's focal loss? While the implementation looks different, the linked repo and torchvision use it for RetinaNet. So replacing it might work. Not completely sure though.

anomalib/models/draem/loss.py
```diff
@@ -0,0 +1,204 @@
...
+        ret = ssim_map.mean()
+    else:
+        ret = ssim_map.mean(1).mean(1).mean(1)
+
+    if full:
+        return ret, cs
+    return ret, ssim_map
+
+
+class SSIM(torch.nn.Module):
```
ashwinvaidya17, line 179: I think we should consider having a folder for losses in components. Also, interestingly kornia has ssim loss https://kornia.readthedocs.io/en/latest/losses.html?highlight=loss#kornia.losses.ssim_loss. Maybe we can experiment using that implementation as we use it as our dependency.

ashwinvaidya17 (changes requested):
Thanks! I have gone through my first round of review.

anomalib/models/draem/augmenter.py
```diff
@@ -0,0 +1,150 @@
...
+        else:  # if no anomaly source is specified, we use the perlin noise as anomalous source
+            anomaly_source_img = np.expand_dims(perlin_noise, 2).repeat(3, 2)
+            anomaly_source_img = (anomaly_source_img * 255).astype(np.uint8)
+
+        # Augment anomaly source image
+        aug = self.rand_augmenter()
+        anomaly_img_augmented = aug(image=anomaly_source_img)
+
+        # Create anomalous perturbation that we will apply to the image
+        perturbation = anomaly_img_augmented.astype(np.float32) * mask / 255.0
```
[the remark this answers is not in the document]
djdameln: We only apply the augmentations during training. The training set only consists of normal images, so we will never augment an anomalous image. During testing/inference we just use the original images, both normal and anomalous, without augmentation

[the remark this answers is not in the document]
ashwinvaidya17: Then anomaly source path might be bit confusing. Also, do we need to pass dataset path? Can't we only stick to augmenting the input batch. The model does this for now 
image

anomalib/models/draem/augmenter.py
djdameln: See my comment in the other thread. I'll make this more flexible by adding different image extensions.

anomalib/models/draem/augmenter.py
djdameln: yep

anomalib/models/draem/config.yaml
djdameln: It's not the most easily digestible algorithm. It would probably be good to add a few sentences explaining this anomaly source dataset in the readme of the draem module.

djdameln: These are some examples of simulated anomalies, clearly showing the carried over textures of the original source images.

Screenshot from 2022-06-03 16-46-15
Screenshot from 2022-06-03 16-45-59
Screenshot from 2022-06-03 16-45-29

samet-akcay: This is a bit similar to our algo in nous. 

https://scholar.google.com/citations?view_op=view_citation&hl=en&user=SVpL2VMAAAAJ&cstart=20&pagesize=80&citation_for_view=SVpL2VMAAAAJ:AHdEip9mkN0C

anomalib/models/draem/loss.py
```diff
@@ -0,0 +1,29 @@
...
+    The total loss consists of the sum of the L2 loss and Focal loss between the reconstructed image and the input
+    image, and the Structural Similarity loss between the predicted and GT anomaly masks.
+    """
+
+    def __init__(self):
+        super().__init__()
+
+        self.l2_loss = nn.modules.loss.MSELoss()
+        self.focal_loss = FocalLoss(alpha=1, reduction="mean")
+        self.ssim_kornia_loss = SSIMLoss(window_size=11)
```
samet-akcay, line 22: Is it needed to emphasize the kornia in the variable? Why not just `self.ssim_loss`?

anomalib/models/draem/LICENSE
```diff
@@ -0,0 +1,29 @@
...
+    The above copyright notice and this permission notice shall be included in all
+    copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+    SOFTWARE.
```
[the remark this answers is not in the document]
djdameln: I believe it is required to include the full text of the license when sublicensing MIT.
"The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software."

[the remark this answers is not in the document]
samet-akcay: Ah I see. My bad

samet-akcay (approved):
Thanks!

anomalib/models/draem/perlin.py
```diff
@@ -0,0 +1,134 @@
+"""Helper functions for generating Perlin noise."""
```
ashwinvaidya17, line 1: Can we move augmenter.py, perlin.py and loss.py to `components/` folder? This way we can keep the root directory clean.

anomalib/models/draem/torch_model.py
```diff
@@ -0,0 +1,495 @@
...
+# Original Code
+# Copyright (c) 2021 VitjanZ
+# https://github.com/VitjanZ/DRAEM.
+# SPDX-License-Identifier: MIT
+#
+# Modified
+# Copyright (C) 2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+# pylint: disable=invalid-name
```
ashwinvaidya17, line 12: Why is this added here?

anomalib/models/draem/torch_model.py
```diff
@@ -0,0 +1,495 @@
...
+        in_channels (int): Number of input channels.
+        out_channels (int): Number of output channels.
+        base_width (int): Base dimensionality of the layers of the autoencoder.
+    """
+
+    def __init__(self, in_channels: int = 3, out_channels: int = 3, base_width: int = 64, out_features: bool = False):
+        super().__init__()
+        self.encoder_segment = EncoderDiscriminative(in_channels, base_width)
+        self.decoder_segment = DecoderDiscriminative(base_width, out_channels=out_channels)
+        # self.segment_act = torch.nn.Sigmoid()
```
ashwinvaidya17, line 87: Leftover?

anomalib/models/draem/perlin.py
djdameln: How about a `utils` directory? It would be more consistent with how we've structured the PatchCore directory. We could then use this convention for other models as well.

anomalib/models/draem/torch_model.py
djdameln: The torch model uses some single letter variable names, e.g. `def forward(x):`. Similar to FastFlow, we could refactor this later.

ashwinvaidya17: Can't we just replace it with batch? Otherwise it might be too many todos for the future

ashwinvaidya17 (approved):
Thanks for the changes!

samet-akcay (approved):

samet-akcay (changes requested):
We need to update
https://github.com/openvinotoolkit/anomalib/blob/development/third-party-programs.txt file

samet-akcay (approved):

hihunjin:
@samet-akcay  Hi. Thanks for your implementation. Did you experiment this model on the whole MVTec AD datasets? If so, please share them.

PR #377, opened 2022-06-16
SFM MVP

ashwinvaidya17:
# Description

The test_step in the callback still needs to be refactored but I've created this for reference.

# Link to results

https://wandb.ai/anomalib/sfm


anomalib/utils/callbacks/selective_feature_model.py
```diff
@@ -0,0 +1,148 @@
...
+            return
+
+        class_names = np.unique(self.class_labels)
+
+        results: Dict[str, list] = {}
+        for class_name in class_names:
+            results[class_name] = []
+        # sorted values and idx for entire feature set
+        max_val, max_idx = torch.sort(self.max_features, descending=True)
+        reduced_range = int(max_val.shape[1] * 0.10)
```
samet-akcay, line 119: the sfm implementation uses `feature_percentage`, but here `reduced_range` is defined by `0.10`?

anomalib/utils/callbacks/selective_feature_model.py
```diff
@@ -0,0 +1,148 @@
...
+            scores = []
+            for class_name in class_names:
+                stats = getattr(self.feature_model, class_name).cpu()
+                score = stats[1][np.isin(stats[0], feature)].sum()
+                scores.append(score)
+            scores[np.where(class_names == "good")[0][0]] = 0
+            if "combined" in class_names:
+                scores[np.where(class_names == "combined")[0][0]] = 0
+            if "thread" in class_names:
+                scores[np.where(class_names == "thread")[0][0]] = 0
```
samet-akcay, line 133: Do you know why we have this sub-category checks only for `good, thread and combined`?

anomalib/models/components/sampling/selective_feature_model.py
```diff
@@ -0,0 +1,62 @@
...
+            max_val, max_idx = torch.sort(class_max_activations, descending=True)
+            reduced_range = int(max_val.shape[1] * self.feature_percentage)
+            # indexes of top 10% FEATURES HAVING MAX VALUE
+            top_max_idx = max_idx[:, 0:reduced_range]
+            # out of sorted top 10, what features are affiliated the most
+            idx, repetitions = torch.unique(top_max_idx, return_counts=True)
+            sorted_repetition, sorted_repetition_idx = torch.sort(repetitions, descending=True)
+            sorted_idx = idx[sorted_repetition_idx]
+
+            sorted_idx_normalized = sorted_repetition / class_max_activations.shape[0]
```
djdameln, line 50: This operation is unnecessary, as we are already normalizing to the L1 norm in the next step

anomalib/data/mvtec.py
```diff
@@ -259,6 +260,7 @@ def __getitem__(self, index: int) -> Dict[str, Union[str, Tensor]]:
             label_index = self.samples.label_index[index]
 
             item["image_path"] = image_path
+            item["class"] = label
```
djdameln, line 263: Maybe instead of "class" we could call this "subclass", "anomalous_subclass", "defect_type" or something similar. This would avoid confusion.

anomalib/data/mvtec.py
```diff
@@ -296,7 +298,7 @@ def __init__(
         transform_config_train: Optional[Union[str, A.Compose]] = None,
         transform_config_val: Optional[Union[str, A.Compose]] = None,
         seed: int = 0,
-        create_validation_set: bool = False,
+        create_validation_set: bool = True,
```
djdameln, line 301: Why this?

anomalib/utils/callbacks/selective_feature_model.py
```diff
@@ -0,0 +1,148 @@
...
+            class_labels = np.hstack(list(outputs["class"]))
+
+            if self.max_features is None:
+                self.max_features = max_val_features
+                self.class_labels = class_labels
+            else:
+                self.max_features = torch.vstack([self.max_features, max_val_features])
+                self.class_labels = np.hstack([self.class_labels, class_labels])
+
+    def on_test_epoch_end(self, _trainer: Trainer, pl_module: LightningModule) -> None:
```
djdameln, line 106: I feel that the contents of this method should be part of the `SelectiveFeatureModel` class. Maybe as a `predict` method, or even better, as the `forward` method (the `fit` method would then also need to be modified of course).

djdameln (changes requested):

anomalib/data/mvtec.py
samet-akcay: maybe `anomaly_type`? I also like `defect_type`, but the content of the task may not always be defect detection.

samet-akcay: or in simple terms, just `sub_class`

PR #379, opened 2022-06-20
Visualizer refactor pt4

djdameln:
# Description

- This PR makes the visualization mode configurable between "full" and "simple".
- "full" visualization is the subplots-style visualization which we previously used for the test set visualization.
  - In segmentation tasks it shows the original image, GT (if available), predicted heatmap, predicted mask and predicted segmentations.
    
  - In classification tasks it shows the original image and the heatmap with the predicted label.
    
- "simple" visualization Only shows the predictions (no GT), and always shows a single frame only.
  - In segmentation tasks it shows both the heatmap and the segmentations, overlayed on the original image.
    
  - In classification tasks it shows the heatmap and the predicted label, overlayed on the original image.
    
- The visualization mode can be configured from `visualization.mode` in the config file, and from the `--visualization_mode` argument of the inference entrypoint script.
- Some refactoring and changes in class responsibility were needed to accommodate the configurability of the visualization mode.
  - A dataclass `ImageResult` was added, which holds all data needed to turn a set of GT items and predictions into a visualized image. This class is used to pass data around easily between different parts of the code.
  - The `Visualizer` class now contains the logic to generate visualizations (numpy images) from the GT and Predictions, depending on the task type and visualization mode. The visualizer can handle both batches and single images.
  - The `Visualizer` class no longer contains the code for showing images on the screen and saving images to the file system. It just generates and returns the images in numpy format. It is the receiver's responsibility to save or display the images.

## Known issues
- Unit tests available yet for the different visualization modes.

## Changes

## Checklist

### DIFF
--- a/anomalib/models/cflow/config.yaml
+++ b/anomalib/models/cflow/config.yaml
@@ -51,6 +51,7 @@ visualization:
   save_images: True # save images to the file system
   log_images: True # log images to the available loggers (if any)
   image_save_path: null # path to which images will be saved
+  mode: simple # options: ["full", "simple"]
 
 project:
   seed: 0
--- a/anomalib/models/padim/config.yaml
+++ b/anomalib/models/padim/config.yaml
@@ -3,7 +3,7 @@ dataset:
   format: mvtec
   path: ./datasets/MVTec
   category: bottle
-  task: segmentation
+  task: classification
   image_size: 256
   train_batch_size: 32
   test_batch_size: 32
@@ -46,6 +46,7 @@ visualization:
   save_images: True # save images to the file system
   log_images: True # log images to the available loggers (if any)
   image_save_path: null # path to which images will be saved
+  mode: simple # options: ["full", "simple"]
 
 project:
   seed: 42
--- a/anomalib/post_processing/visualizer.py
+++ b/anomalib/post_processing/visualizer.py
@@ -14,36 +14,164 @@
 # See the License for the specific language governing permissions
 # and limitations under the License.
 
-from pathlib import Path
-from typing import Dict, List, Optional
+from dataclasses import dataclass, field
+from typing import Dict, Iterator, List, Optional
 
 import cv2
 import matplotlib.figure
 import matplotlib.pyplot as plt
 import numpy as np
+from skimage.segmentation import mark_boundaries
+
+from anomalib.post_processing import (
+    add_anomalous_label,
+    add_normal_label,
+    superimpose_anomaly_map,
+)
+from anomalib.pre_processing.transforms import Denormalize
+
+
+@dataclass
+class ImageResult:
+    """Collection of data needed to visualize the predictions for an image."""
+
+    image: np.ndarray
+    pred_score: float
+    pred_label: str
+    anomaly_map: np.ndarray
+    gt_mask: Optional[np.ndarray] = None
+    pred_mask: Optional[np.ndarray] = None
+
+    heat_map: np.ndarray = field(init=False)
+    segmentations: np.ndarray = field(init=False)
+
+    def __post_init__(self):
+        """Generate heatmap overlay and segmentations, convert masks to images."""
+        self.heat_map = superimpose_anomaly_map(self.anomaly_map, self.image, normalize=False)
+        if self.pred_mask is not None:
+            self.pred_mask *= 255
+            self.segmentations = mark_boundaries(self.image, self.pred_mask, color=(1, 0, 0), mode="thick")
+        if self.gt_mask is not None:
+            self.gt_mask *= 255
 
 
 class Visualizer:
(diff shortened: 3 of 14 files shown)


anomalib/utils/callbacks/visualizer_callback.py
```diff
@@ -61,17 +57,19 @@ def __init__(
         self.save_images = save_images
         self.image_save_path = Path(image_save_path)
```
samet-akcay, line 59: maybe add value check to `task` and `mode` here as well?

anomalib/utils/callbacks/visualizer_callback.py
```diff
@@ -87,41 +85,21 @@ def _add_to_logger(
...
+    @staticmethod
+    def _show_image(title, image):
+        cv2.imshow(title, image)
+        cv2.waitKey(0)
+        cv2.destroyAllWindows()
+
+    @staticmethod
+    def _save_image(file_path, image):
+        file_path.parent.mkdir(parents=True, exist_ok=True)
+        cv2.imwrite(str(file_path), image)
```
samet-akcay, line 102: I feel like it would be better if `show_image` and `save_image` belong to `Visualizer` class

samet-akcay: The reason being that `Visualizer` already have `add_image`. It would be more logical to me if `show_image` is part of `Visualizer` object as well.

tools/inference.py
```diff
@@ -25,6 +25,13 @@ def get_args() -> Namespace:
     parser.add_argument("--config", type=Path, required=True, help="Path to a model config file")
     parser.add_argument("--weight_path", type=Path, required=True, help="Path to a model weights")
     parser.add_argument("--image_path", type=Path, required=True, help="Path to an image to infer.")
+    parser.add_argument(
+        "--visualization_mode",
+        type=str,
+        required=False,
+        default="simple",
+        help="Visualization mode. 'full' or 'simple'",
```
samet-akcay, line 33: Maybe we could add `choices=['simple', 'full']`

samet-akcay (changes requested):
Thanks a ton! Once done, it'll be a lot cleaner!

ashwinvaidya17 (approved):
Thanks for the efforts! I have nothing to add here

anomalib/models/padim/config.yaml
```diff
@@ -3,7 +3,7 @@ dataset:
   format: mvtec
   path: ./datasets/MVTec
   category: bottle
-  task: segmentation
+  task: classification
```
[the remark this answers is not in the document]
djdameln: This is a leftover from debugging

anomalib/models/cflow/config.yaml
```diff
@@ -51,6 +51,7 @@ visualization:
   save_images: True # save images to the file system
   log_images: True # log images to the available loggers (if any)
   image_save_path: null # path to which images will be saved
+  mode: simple # options: ["full", "simple"]
```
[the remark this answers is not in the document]
djdameln: Hmm no this is by mistake. I'll change it to "full"

anomalib/utils/callbacks/visualizer_callback.py
djdameln: I've added the checks, but it clutters the classes a bit. Maybe we can switch to using enums for selectable parameters in the future.

samet-akcay: enums would be good to avoid such cases. Their disadvantage, though, is it couples the code a bit.

anomalib/post_processing/visualizer.py
```diff
@@ -14,36 +14,193 @@
...
+    gt_mask: Optional[np.ndarray] = None
+    pred_mask: Optional[np.ndarray] = None
+
+    heat_map: np.ndarray = field(init=False)
+    segmentations: np.ndarray = field(init=False)
+
+    def __post_init__(self):
+        """Generate heatmap overlay and segmentations, convert masks to images."""
+        self.heat_map = superimpose_anomaly_map(self.anomaly_map, self.image, normalize=False)
+        if self.pred_mask is not None and not np.max(self.pred_mask) == 255:
```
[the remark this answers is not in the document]
djdameln: I'll change the check to `max(pred_mask) <= 1`. In any case it is unlikely that the image is in the [0, 255] range, because the predicted mask is the result of the thresholding mechanism applied to the anomaly maps.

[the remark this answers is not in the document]
samet-akcay: sure, that makes sense. I was just thinking that in case it's used as an API inside or outside anomalib.

samet-akcay (approved):

PR #396, opened 2022-07-01
Change requirements

ashwinvaidya17:
# Description

- Torchvision 0.13.0 ends up installing torch 0.12.0 which leads to segmentation fault in OpenVINO (2022.1.0) inferencer. This PR fixes the versions.


ashwinvaidya17:
Reopening as this adds range to the requirements

requirements/base.txt
```diff
@@ -6,9 +6,9 @@ opencv-python>=4.5.3.56
 pandas>=1.1.0
 pytorch-lightning[extra]>=1.6.0
 timm==0.5.4
-torchmetrics>=0.9.0
-torchvision>=0.9.1
-torchtext>=0.9.1
+torchmetrics>=0.9.0,<=0.9.1
```
samet-akcay, line 9: do we need to restrict `torchmetrics`? It's a different library by the Lightning team,

PR #404, opened 2022-07-05
Fix: data split issue

### LINKED ISSUE #403: When test_split is applied to a custom dataset, test data contamination occurs
**Describe the bug**
- When test_split is applied to a custom dataset, test data contamination occurs. (test data(test data is also used for training).

**To Reproduce**
1. Use custom dataset 
2. Set `split_ratio` to a specific value in config.yaml (without `nomal_test_dir`)
3. See error (test data contamination) 

**Expected behavior**
- I hope that the train and test `samples` are the same because the same seed is used. 
I hope that the train and test results are the same because the same seed is used.
https://github.com/openvinotoolkit/anomalib/blob/c5220c037f1580df5f7a7474ae488483f0f83eae/anomalib/data/folder.py#L164
- However, `samples` are different. So test data contamination occurs. 

**Screenshots**
- If applicable, add screenshots to help explain your problem.

- OS: [ Windows 10 ]
- NVIDIA Driver Version [470.57.02]
- CUDA Version [11.3.1]

**Additional context**
- PatchCore

jeongHwarr:
# Description

- Fixes #403

## Changes

## Checklist

### DIFF
--- a/anomalib/data/utils/split.py
+++ b/anomalib/data/utils/split.py
@@ -49,7 +49,7 @@ def split_normal_images_in_train_set(
         DataFrame: Output dataframe where the part of the training set is assigned to test set.
     """
 
-    if seed > 0:
+    if seed >= 0:
         random.seed(seed)
 
     normal_train_image_indices = samples.index[(samples.split == "train") & (samples.label == normal_label)].to_list()
@@ -74,7 +74,7 @@ def create_validation_set_from_test_set(samples: DataFrame, seed: int = 0, norma
         normal_label (str): Name of the normal label. For MVTec AD, for instance, this is normal_label.
     """
 
-    if seed > 0:
+    if seed >= 0:
         random.seed(seed)
 
     # Split normal images.


ashwinvaidya17:
What's wrong with setting the seed to a value other than 0? If we make this change, we will have to ensure that lines like

```python
if config.project.seed != 0:
        seed_everything(config.project.seed)
```

are changed in other files - `train.py`, etc

jeongHwarr:
@ashwinvaidya17 

If seed is non-zero, there is no problem.
However, if the seed is 0, the test data is also included in the training data. So the performance is abnormally high. 

Does seed 0 means that the seed is not fixed in this project?
If so, the code seems to need a lot of modification.

It seems that `samples` are shared rather than made the same seed.
https://github.com/openvinotoolkit/anomalib/blob/83645772754a1b7d909930b9b49726ed460696d8/anomalib/data/folder.py#L163-L166

Now, the PR code solves the data contamination problem,  but each traininig data and test data are created identically for the seed 0.

ashwinvaidya17:
@jeongHwarr not sure if I quite understand the contamination problem. Currently we treat seed 0 as having no seed. However, even with seed 0 I get no overlap between train split and the test split. Here are the sample lines of code I am using for testing this. This does not raise Assertion Error for multiple runs.

```python
from anomalib.data.folder import make_dataset
from pathlib import Path

dataset = make_dataset(normal_dir=Path("../datasets/bottle/good/"), abnormal_dir=Path('../datasets/bottle/broken_large'), split_ratio=0.33, seed=0)

train_split = dataset[dataset.split == "train"]
test_split = dataset[dataset.split == "test"]
assert train_split["image_path"].values.all() != test_split["image_path"].values.all()
```

Here is another sample using Folder DataModule
```python
folder_dataset = Folder(root=Path("../datasets/bottle"),normal_dir="good", abnormal_dir='broken_large', split_ratio=0.1, seed=0, image_size=256)
folder_dataset.setup()
assert folder_dataset.test_data.samples["image_path"].values.all() != folder_dataset.train_data.samples["image_path"].values.all()
```

jeongHwarr:
@ashwinvaidya17 

There is no overlap in your test code because you called `make_dataset` only once.

However, in the code of this project, `make_dataset` is called multiple times. At this time, if the seed is not fixed, the `sample` will not come out the same.
https://github.com/openvinotoolkit/anomalib/blob/83645772754a1b7d909930b9b49726ed460696d8/anomalib/data/folder.py#L163-L166
To reproduce this situation, set `normal_dir` to `None` in yaml, let the above part be called, and then compare the training data and test data. Then you can find the overlap there.


`make_dataset` is called multiple times through the code below.
https://github.com/openvinotoolkit/anomalib/blob/83645772754a1b7d909930b9b49726ed460696d8/anomalib/data/folder.py#L476-L523

https://github.com/openvinotoolkit/anomalib/blob/83645772754a1b7d909930b9b49726ed460696d8/anomalib/data/folder.py#L246-L256


To solve this problem, `make_datset` is called only once, and the below part in `make_dataset` must be processed in another part not in the `make_datset`.
https://github.com/openvinotoolkit/anomalib/blob/83645772754a1b7d909930b9b49726ed460696d8/anomalib/data/folder.py#L172-L175

ashwinvaidya17:
@jeongHwarr I see what you mean. Thanks for spotting this. I'll look into the issue. But does it make sense to keep seed 0 as no seed or should we refactor the code to make `seed: null` as no seed?

ashwinvaidya17:
@jeongHwarr Thanks again for spotting this issue and creating a PR for addressing it. Is it possible for you to make a few changes to this PR (https://github.com/openvinotoolkit/anomalib/pull/437/files) so that we can merge this and you can be a contributor to this repo 😀. Other option is to always merge this and I can create a new PR to address this.

jeongHwarr:
@ashwinvaidya17 

I think that `None` or `null` is more clear than seed 0.
Until I saw the code, I understood that it was fixed as seed 0.

I've reviewed your PR and nothing seems to change!
Thank you 😀

samet-akcay (approved):

PR #429, opened 2022-07-11
Add metric visualizations

### LINKED ISSUE #420: Add plots and additional metrics
Imo, it would be nice to have:

* Pixel-wise AUPR + PRO until 30% FPR as metrics
* Plots with baselines where applicable for all metrics (ROC-Curve, PR-Curve, PRO-Curve)

  1. Add/implement appropriate torchmetric classes, and add them to the `AnomlibMetricCollection`
  2. Add callbacks which access the underlying scores for metric calculation (i.e. tuple of tpr/fpr for auroc) and which subsequently use the scores to generate and save the plots.
      * Alternatively, one could think of expanding/inheriting the `VisualizerCallback`

Thoughts?

ORippler:
# Description
This PR:
* Adds 2 additional metrics:
  * [x] AUPR
  * [x] AUPRO
* [x] Adds plots for all metrics by refactoring `VisualizerCallback` into `VisualizerCallbackBase` and `VisualizerCallbackimage`, and by introducing a new `VisualizerCallbackMetric` class

- Fixes #420 (issue)

## Changes

## Checklist


anomalib/utils/callbacks/visualizer_base.py
```diff
@@ -0,0 +1,106 @@
...
+        Currently only ``AnomalibWandbLogger.save`` is called from this method.
+        This is because logging as a single batch ensures that all images appear as part of the same step.
+
+        Args:
+            trainer (pl.Trainer): Pytorch Lightning trainer
+            pl_module (AnomalyModule): Anomaly module (unused)
+        """
+        for logger in trainer.loggers:
+            if isinstance(logger, AnomalibWandbLogger):
+                logger.save()
```
ORippler, line 106: Unsure whether it's "bad practice" to call .save on a `AnomalibWandbLogger` multiple times, as is done currently. Any insights?

ashwinvaidya17: It has been a while since I have looked into AnomalibWandbLogger but ideally it should only be called once at the very end. This is explained in the comment above `This is because logging as a single batch ensures that all images appear as part of the same step.`. In case it is being called multiple times then my guess is that the `isinstance` fallsback to the base class and probably logger ends up always being an instance of `AnomalibWandbLogger` (`LightningLogger` more specifically). Maybe a good idea would be to use `type()` but mypy or pylint might complain.

samet-akcay:
@ORippler, thanks a lot for your contribution! Just so know, code quality checks have failed due to number of reasons

1. anomalib/utils/cli/cli.py:21:0: E0611: No name 'VisualizerCallback' in module 'anomalib.utils.callbacks' (no-name-in-module)
2.  notebooks/200_models/201_fastflow.ipynb:cell_1:17:0: E0611: No name 'VisualizerCallback' in module 'anomalib.utils.callbacks' (no-name-in-module)

For more details, you could refer to [this](https://github.com/openvinotoolkit/anomalib/runs/7284425098?check_suite_focus=true).

Will you be able to address this ?

ORippler:
> @ORippler, thanks a lot for your contribution! Just so know, code quality checks have failed due to number of reasons
> 
>     1. anomalib/utils/cli/cli.py:21:0: E0611: No name 'VisualizerCallback' in module 'anomalib.utils.callbacks' (no-name-in-module)
> 
>     2. notebooks/200_models/201_fastflow.ipynb:cell_1:17:0: E0611: No name 'VisualizerCallback' in module 'anomalib.utils.callbacks' (no-name-in-module)
> 
> 
> For more details, you could refer to [this](https://github.com/openvinotoolkit/anomalib/runs/7284425098?check_suite_focus=true).
> 
> Will you be able to address this ?

Sure, once I get all the functionality implemented.

anomalib/utils/callbacks/visualizer_base.py
ORippler: I opted with clearing `self.images` on save. Thus, all predictions are in the first step, and all metrics in the second step

ORippler:
@samet-akcay

I got the core functionality implemented and tested it locally so far. Some points from my side:

* While I confirmed that metric-images are saved to tensorboard via the following snippet:
```
import cv2
import numpy as np
from matplotlib import pyplot as plt
from tensorboard.backend.event_processing.event_accumulator import EventAccumulator

acc = EventAccumulator("PATH")
acc.Reload()

for tag in acc.Tags()["images"]:
    if "image_" in tag or "pixel_" in tag:
        events = acc.Images(tag)

        for index, event in enumerate(events):
            s = np.frombuffer(event.encoded_image_string, dtype=np.uint8)
            image = cv2.imdecode(s, cv2.IMREAD_COLOR)
            image = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
            plt.imshow(image)
            plt.show()
```
they don't necessarily appear when I open a `tensorboard` on the folder.
* I did not test the changes to `CLI`, as I did not manage to call `anomalib test` after training (`project.path` is no longer in the namespace and cannot be provided via `anomalib test --config --project.path`. Therefore, a new time-stamped folder ist generated instead...)
* instead of testing for presence of `generate_figure`, one could potentially generate an `AnomalibMetric` class with an abstractmethod, and let our own metrics inherit from that class (and subsequently test via `isinstance(AnomalibMetric)`).
* The current implementation results in two calls to `Metric._compute` as opposed to the singular call prior. While this is not a bottleneck (at least not on an RTX 3090), one could nonetheless think of storing results from the first call and simply returning them instead.

ashwinvaidya17 (approved):
Thanks for the efforts!

anomalib/utils/callbacks/visualizer_base.py
```diff
@@ -0,0 +1,109 @@
+"""Base Visualizer Callback."""
```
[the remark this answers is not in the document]
ORippler: Done

ORippler:
In addition to bugfixing the PRO-calculation, i passed the mypy checks and clarified wording.

Btw, `mypy` behaves differently when called with `tox` vs via the `pre-commit` hook (which didn't threw the errors I now fixed), at least in the provided `Dockerfile`

samet-akcay (commented):
@ORippler, thanks again for such contribution! I only have a single comment regarding the naming. Other than, that looks great!

ORippler:
Alright so the test for ganomaly should pass now as well (There were a couple of bugs in visualizer for the `classification` task as well as in the `config` for ganomaly, which were unrelated to my PR).

samet-akcay:
> Alright so the test for ganomaly should pass now as well (There were a couple of bugs in visualizer for the `classification` task as well as in the `config` for ganomaly, which were unrelated to my PR).

Thanks, yeah, I also noticed the classification bug in the visualizer yesterday. Will fix it in a separate PR

samet-akcay (approved):
Many thanks for your effort and valuable contribution!

anomalib/post_processing/visualizer.py
```diff
@@ -132,9 +133,9 @@ def _visualize_full(self, image_result: ImageResult):
         elif self.task == "classification":
             visualization.add_image(image_result.image, title="Image")
             if image_result.pred_label:
-                image_classified = add_anomalous_label(image_result.heat_map, image_result.pred_score)
+                image_classified = add_anomalous_label(image_result.image, image_result.pred_score)
             else:
-                image_classified = add_normal_label(image_result.heat_map, 1 - image_result.pred_score)
+                image_classified = add_normal_label(image_result.image, 1 - image_result.pred_score)
```
simon376, line 138: I'm curious why this was changed to show the original image, instead of the heatmap one?
Even if one only wants to classify images, it's interesting to see the heatmap during development.

curiously enough, it's also inconsistent now, since in `_visualize_simple` (L:161) , the heatmap is still used

ORippler: Not all AD methods generate heat_maps (for example, ganomaly does not). Therefore, we need to distinguish between `classification` and `segmentation` tasks also in the `Visualizer`. The behavior for `segmentation` tasks should therefore be left unchanged (we check for task in L:133).

Regarding the second point: I might have missed changing it there/it might actually be a bug, should be investigated further.

PR #453, opened 2022-07-25
Add Torch Inferencer and Update Openvino and Gradio Inferencers.

samet-akcay:
# Description

- Add Torch Inferencer.
- Return all the predictions from `inferencer.predict(image)`. This way, the user will be able to access to any info they need.

- Fixes # 294

## Changes

## Checklist


anomalib/deploy/inferencers/openvino_inference.py
```diff
@@ -16,12 +16,13 @@
```
djdameln, line 16: Can we rename this file to `openvino_inferencer.py`? The naming is a bit confusing now, because the file has the same name as `tools/inference/openvino_inference.py`

tools/inference/torch_inference.py
```diff
@@ -0,0 +1,92 @@
...
+    visualizer = Visualizer(mode=args.visualization_mode, task=args.task)
+
+    filenames = get_image_filenames(path=args.input)
+    for filename in filenames:
+        image = read_image(filename)
+        predictions = inferencer.predict(image=image)
+        output = visualizer.visualize_image(predictions)
+
+        if args.output:
+            file_path = generate_output_image_filename(input_path=args.input, output_path=args.output)
```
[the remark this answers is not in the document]
samet-akcay: yes, that's right! Thanks for spotting this!

anomalib/data/inference.py
```diff
@@ -47,7 +47,7 @@ def __init__(
         """
         super().__init__()
 
-        self.image_filenames = get_image_filenames(path)
+        self.image_filenames = [str(filename) for filename in get_image_filenames(path)]
```
ashwinvaidya17, line 50: Do we need this? Doesn't `read_image` already support `Path` type

tools/inference/torch_inference.py
```diff
@@ -0,0 +1,91 @@
...
+        predictions = inferencer.predict(image=image)
+        output = visualizer.visualize_image(predictions)
+
+        if args.output:
+            file_path = generate_output_image_filename(input_path=filename, output_path=args.output)
+            visualizer.save(file_path=file_path, image=output)
+
+        # Show the image in case the flag is set by the user.
+        if args.show:
+            visualizer.show(title="Output Image", image=output)
```
ashwinvaidya17, line 87: It would be nice to display a warning if both show and output are None

tools/inference/torch_inference.py
```diff
@@ -0,0 +1,92 @@
...
+def get_args() -> Namespace:
+    """Get command line arguments.
+
+    Returns:
+        Namespace: List of arguments.
+    """
+    parser = ArgumentParser()
+    parser.add_argument("--config", type=Path, required=True, help="Path to a config file")
+    parser.add_argument("--weights", type=Path, required=True, help="Path to model weights")
+    parser.add_argument("--meta_data", type=Path, required=False, help="Path to a JSON file containing the metadata.")
```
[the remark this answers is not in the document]
djdameln: What if we want to deploy the model from the .pth file instead of the .ckpt file? In that case we would need the meta_data dictionary. This is related to another concern that I'm having. The torch inferencer now still uses lightning classes under the hood. We could consider making the torch inferencer pytorch-only, and have it load a .pth model instead of .ckpt model. This way we would not need to import pytorch lightning to run torch inference and keep the different inferencers better separated. The advantage for users would be that it's more in line with how they will likely incorporate their trained models into their software.

anomalib/deploy/inferencers/openvino_inference.py
```diff
@@ -141,9 +149,30 @@ def post_process(
         anomaly_map = predictions.squeeze()
         pred_score = anomaly_map.reshape(-1).max()
 
+        # Common practice in anomaly detection is to assign anomalous
+        # label to the prediction if the prediction score is greater
+        # than the image threshold.
```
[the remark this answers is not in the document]
samet-akcay: This would bring up another discussion. Do you think we should check if they are in meta_data?

For example,

```
if "image_threshold" in meta_data:
    pred_label = pred_score >= meta_data["image_threshold"]
```

We cannot calculate the pred-label if we don't have this. The question is whether we should have this if statement here, and for the rest of the computation?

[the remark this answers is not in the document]
ashwinvaidya17: I think the default threshold is 0.5 so if we don't have `image_threshold` in meta_data then we can throw a warning and reduce the if statement to `pred_label = pred_score >= meta_data.get("image_threshold", 0.5)`

anomalib/deploy/inferencers/openvino_inferencer.py
```diff
@@ -132,18 +140,51 @@ def post_process(
...
+            if "image_shape" in meta_data and anomaly_map.shape != meta_data["image_shape"]:
+                image_height = meta_data["image_shape"][0]
+                image_width = meta_data["image_shape"][1]
+                anomaly_map = cv2.resize(anomaly_map, (image_width, image_height))
+
+                if pred_mask is not None:
+                    pred_mask = cv2.resize(pred_mask, (image_width, image_height))
+
+        return {
+            "anomaly_map": anomaly_map,
```
ashwinvaidya17, line 186: Would this work for task type of classification?

anomalib/deploy/inferencers/torch_inferencer.py
```diff
@@ -161,4 +181,12 @@ def post_process(
             image_width = meta_data["image_shape"][1]
             anomaly_map = cv2.resize(anomaly_map, (image_width, image_height))
 
-        return anomaly_map, float(pred_score)
+            if pred_mask is not None:
+                pred_mask = cv2.resize(pred_mask, (image_width, image_height))
+
+        return {
+            "anomaly_map": anomaly_map,
```
ashwinvaidya17, line 188: Same here.

anomalib/deploy/inferencers/openvino_inference.py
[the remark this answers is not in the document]
djdameln: As long as we throw a warning it should be fine. Because the users need to be made aware that there is a problem with the metadata file.

djdameln (approved):
Thanks! All my comments have been addressed

anomalib/deploy/inferencers/openvino_inferencer.py
samet-akcay: Yeah, it returns none, which ImageResults handles ok.

anomalib/deploy/inferencers/torch_inferencer.py
samet-akcay: same as above.

ashwinvaidya17 (approved):
Thanks!

PR #462, opened 2022-07-29
🐞 Fix HPO

### LINKED ISSUE #452: Error while trying to do HPO of patchcore
I am trying to to do hyperparameter optimisation of patchcore model on a custom dataset but structured in the mvtec format. But the optimization is not happening and it is showing me this error.

ERROR Run wsg1bt94 errored: AttributeError("'PatchcoreLightning' object has no attribute 'adaptive_threshold'")

ashwinvaidya17:
# Description

- Address callbacks issue in hpo script. Fixes #452 
- Change pre-commit to use system version of node otherwise it downloads the latest version which might target libraries which might not be supported in the host OS build.


tools/hpo/sweep.py
```diff
@@ -22,14 +22,17 @@
 from omegaconf import DictConfig, ListConfig, OmegaConf
 from pytorch_lightning import seed_everything
 from pytorch_lightning.loggers import WandbLogger
+from utils import flatten_hpo_params
```
samet-akcay, line 25: I previously got a linting error because of this. Does this work?

ashwinvaidya17: I added `__init__.py` to `tools/benchmarking` so it works now. The issue is that `astroid` (pylint's base) cannot distinguish two folders with the same name in the same hierarchy (tools). But apparently it works when` __init__.py` is provided. I am not sure why would the two folders being a package make a difference.

samet-akcay (approved):

PR #470, opened 2022-08-01
fix inference for draem

djdameln:
# Description
This PR fixes inference for DRAEM and possibly other models.

- The transform config files were not loaded correctly in each of the three inferencers, so the default transform config was always used. This is solved now by retrieving the transform config from the config file using the correct key.
- The torch model from the DRAEM implementation now returns an anomaly map when in inference mode, instead of the confidence scores for both classes. This makes the model compatible with the torch and openvino inferencers.

## Changes

## Checklist

### DIFF
--- a/anomalib/deploy/inferencers/openvino_inferencer.py
+++ b/anomalib/deploy/inferencers/openvino_inferencer.py
@@ -104,9 +104,11 @@ class OpenVINOInferencer(Inferencer):
         Returns:
             np.ndarray: pre-processed image.
         """
-        config = self.config.transform if "transform" in self.config.keys() else None
+        transform_config = (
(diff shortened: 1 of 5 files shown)


ashwinvaidya17:
If OpenVINO inference works for DRAEM can you add it to this test https://github.com/openvinotoolkit/anomalib/blob/main/tests/pre_merge/deploy/test_inferencer.py#L88

samet-akcay:
> If OpenVINO inference works for DRAEM can you add it to this test https://github.com/openvinotoolkit/anomalib/blob/main/tests/pre_merge/deploy/test_inferencer.py#L88

I agree, it might be an idea to add DRAEM to the tests.

samet-akcay (approved):
Thanks for the fix!

ashwinvaidya17 (approved):
Thanks for identifying the issue!

PR #476, opened 2022-08-02
Fix onnx export by rewriting GaussianBlur

### LINKED ISSUE #472: Export to onnx fails for some models with Pytorch 1.12
**Describe the bug**

in Pytorch 1.12, export to onnx fails for all models that rely on [`kornia.filters.gaussianblur2d`](https://github.com/kornia/kornia/blob/c8cd4040a375ec7d9b7bd604634d96721c46c9e0/kornia/filters/gaussian.py#L10).

**To Reproduce**

1. install PyTorch 1.12
2. run a model (e.g. `padim`) with export to onnx enabled

**Expected behavior**

Export to onnx works also in Pytorch 1.12

**Additional context**

I believe the error is due to missing shape annotations of [`torch.nn.functional.interpolate`](https://discuss.pytorch.org/t/how-to-specify-the-interpolate-layers-output-shape-when-export-to-onnx/141898), at least that's where shapes are lost when inspecting via netron.

The problem can be fixed by reimplementing GaussianBlur to not rely on the input shape for computing the kernel (as a drawback `num_channels` have to be provided explicitly).

ORippler:
# Description

* In PyTorch 1.12, export to onnx fails due failing shape-inference of torch.nn.functional.interpolate, refer #472.
* Failing exports can be fixed by reimplementing `GaussianBlur2d` to no longer rely on input shapes for calculation of the kernel size. As a caveat, now the number of channels need to be specified during init

- Fixes #472 

## Changes

## Checklist


ORippler:
2 things from my side:

1. The current `Dockerfile` does not contain all deps needed to develop anomalib. Is this intended?
2. I changed the `AnomalyMapGenerator` to inherit from `nn.Module`, should this also be done for all other instances of `AnomalyMapGenerator` for consistency? Was there a reasoning for not making all `AnomalyMapGenerator` classes instances of `nn.Module` initially?

anomalib/models/utils/blur.py
```diff
@@ -0,0 +1,76 @@
...
+from typing import Tuple, Union
+
+from kornia.filters import get_gaussian_kernel2d
+from kornia.filters.filter import _compute_padding
+from kornia.filters.kernels import normalize_kernel2d
+from torch import Tensor, nn
+from torch.nn import functional as F
+
+
+class GaussianBlur2d(nn.Module):
```
samet-akcay, line 11: One might consider `GaussianBlur2d` as a post-processing step. Would it be an idea to move this `anomalib.post_processing`? @ORippler, @ashwinvaidya17, @djdameln ?

samet-akcay:
> I changed the AnomalyMapGenerator to inherit from nn.Module, should this also be done for all other instances of AnomalyMapGenerator for consistency? Was there a reason I am unaware of for not making all other AnomalyMapGenerator classes instances of nn.Module initially?

As far as I'm aware there was no reason. It would be a good idea to inherit nn.Module for every AnomalyMapGenerator for consistency. It could perhaps be another pr since the scope of this pr is gaussian blur

anomalib/models/utils/blur.py
ORippler: Sure!

In my first understanding elements of `post_processing` were to be applied to the outputs of the complete `AnomalyModule`, which is why I didn't put the blur there initially (`AnomlyMapGenerators` were part of the `AnomalyModule` and thus the "main process" in my head, which is why blurring also wasn't a post-process).

samet-akcay: That's true. post_process is aimed to be applied to the output of the anomaly module. 

Gaussian blur, more specifically, is one of the final steps of the `AnomalyMapGenerator`, which is the final step of the anomaly module. That's why I thought that it could be considered as a post-processing step.

What do you guys think @ashwinvaidya17 @djdameln ?

djdameln: I agree that the `post_processing` should be reserved for operations that are applied to the output of the complete anomaly module. Also, the use-cases of GaussianBlur2d are not limited to the post-processing of the anomaly maps, but it could potentially be used in other parts of the model pipeline as well. That's why I would suggest to store it either in `anomalib/models/components` or `anomalib/utils`. I think my preference would go to the first option.

djdameln (commented):
Thanks for addressing this! I'll approve the PR once we settle on the appropriate location to store the new GaussianBlur class.

anomalib/models/utils/blur.py
ORippler: I am also in favor of `anomalib/models/components`, maybe `anomlib/models/components/utils`

samet-akcay: Moving it to `anomalib/models/components` would make more sense, I agree. However, here is the tree structure of the components.

```bash
models/components
├── __init__.py
├── __pycache__
├── base
├── dimensionality_reduction
├── feature_extractors
├── freia
├── sampling
└── stats
```

So we cannot simply place it there. In this case, there would be need for either a new subpackage called such as `anomaly_map` to place anomaly map related components there. Or, as @ORippler suggested it could be `anomalib/models/components/utils`.

ashwinvaidya17: How about `anomalib/models/components/filters`? If I am not wrong kornia stores gaussian blur under filters package. We can maybe follow their convention.

ashwinvaidya17 (approved):
Thanks for identifying and fixing this! I am fine with the changes but I'd wait for the consensus from others regarding the location of blur.py before merging.

anomalib/models/utils/blur.py
samet-akcay: I'm fine with `anomalib/models/components/filters`

ORippler:
> Thanks for identifying and fixing this! I am fine with the changes but I'd wait for the consensus from others regarding the location of blur.py before merging.

Moved it to `anomalib/models/components/filters`, feel free to merge after tests pass

samet-akcay (approved):

PR #499, opened 2022-08-12
Add coverage

ashwinvaidya17:
# Description

- Add coverage reporter to CI

## Note

- I am not sure how this will behave in `main` branch.

### DIFF
--- a/.github/workflows/pre_merge.yml
+++ b/.github/workflows/pre_merge.yml
@@ -31,11 +31,11 @@ jobs:
           echo ${{ github.event.pull_request.head.sha }}
           if [[ ${{ github.event.pull_request.head.sha }} ]]
           then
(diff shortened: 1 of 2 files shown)


.github/workflows/pre_merge.yml
```diff
@@ -25,8 +25,14 @@ jobs:
...
           export CUDA_VISIBLE_DEVICES=3
           tox -e pre_merge
-      - name: Upload coverage result
-        uses: actions/upload-artifact@v2
-        with:
-          name: coverage
-          path: .tox/coverage.xml
+      - name: Upload coverage report
+        run: |
+
```
samet-akcay, line 30: It might be an idea to explain this `if`-block with a comment for future reference.

samet-akcay (approved):
Thanks a ton!

djdameln (approved):

PR #500, opened 2022-08-12
SSPCAB implementation

djdameln:
# Description

- This PR adds SSPCAB to model components and adds it as an optional extension to DRAEM.

## Changes

## Checklist


anomalib/models/components/layers/sspcab.py
```diff
@@ -0,0 +1,78 @@
+"""Self-Supervised Predictive Convolutional Attention (SSPCAB) block for reconstruction-based models."""
```
samet-akcay, line 1: Can you provide a link to the original paper? Perhaps to the official implementation as well?

anomalib/models/components/layers/sspcab.py
```diff
@@ -0,0 +1,78 @@
...
+        super().__init__()
+
+        out_channels = channels // reduction_ratio
+        self.fc1 = nn.Linear(channels, out_channels)
+        self.fc2 = nn.Linear(out_channels, channels)
+
+    def forward(self, inputs: Tensor) -> Tensor:
+        """Forward pass through the attention module."""
+        # average pool
+        avg_pooled = inputs.mean(dim=(2, 3))
```
samet-akcay, line 29: for future reference or inform the reader, maybe we could comment here why `dim=(2,3)` is averaged

anomalib/models/components/layers/sspcab.py
```diff
@@ -0,0 +1,78 @@
...
+    """SSPCAB block.
+
+    Args:
+        channels (int): Number of input channels.
+        kernel_size (int): Size of the receptive fields of the masked convolution kernel.
+        dilation (int): Dilation factor of the masked convolution kernel.
+        reduction_ratio (int): Reduction ratio of the attention module.
+    """
+
+    def __init__(self, channels, kernel_size: int = 1, dilation: int = 1, reduction_ratio: int = 8):
```
samet-akcay, line 53: `channels: int` ?

samet-akcay (commented):
I like it! Only few minor points

anomalib/models/draem/lightning_model.py
```diff
@@ -30,12 +31,40 @@ class Draem(AnomalyModule):
...
+        def get_activation(name: str) -> Callable:
+            """Retrieves the activations.
+
+            Args:
+                name (str): Identifier for the retrieved activations.
+            """
+
+            def hook(_, __, output: Tensor):
+                """Hook for retrieving the activations."""
+                self.sspcab_activations[name] = output.detach()
```
ashwinvaidya17, line 62: I don't get this. If we detach the outputs then will it be able to compute gradient? My understanding might be wrong here but shouldn't we only detach the input?

djdameln: yes, good catch. I've changed it

ashwinvaidya17 (approved):
Thanks!

anomalib/models/draem/config.yaml
```diff
@@ -24,6 +24,8 @@ model:
   name: draem
   anomaly_source_path: null # optional, e.g. ./datasets/dtd
   lr: 0.0001
+  enable_sspcab: false
+  sspcab_lambda: 0.1
```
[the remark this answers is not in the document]
djdameln: This would be possible, but I think it would be more clear for the user to have an explicit parameter to enable/disable SSPCAB. Especially because we would have to add some logic to enable/disable the sspcab block in the architecture of the torch model based on the value of the lambda parameter. As a user I would not expect the value of the lambda parameter to affect the architecture of the model.

samet-akcay (approved):

PR #508, opened 2022-08-18
Add PRO metric

djdameln:
# Description
This PR adds the PRO metric, and speeds up AUPRO computation when on cpu.

- Add PRO metric
- Add alternative connected component computation method to speed up evaluation on CPU.

## Changes

## Checklist


anomalib/utils/metrics/pro.py
```diff
@@ -0,0 +1,132 @@
...
+
+    if not converged:
+        warnings.warn(
+            f"Max iterations ({max_iterations}) reached before converging. Connected component results may be "
+            f"inaccurate. Consider increasing the maximum number of iterations."
+        )
+
+    # remap component values from 0 to N
+    labels = components.unique()
+    for new_label, old_label in enumerate(labels):
```
ashwinvaidya17, line 106: Not sure if I quite understood this correctly. When computing connected components in cpu, we go over each mask in the batch and find all the connected components. Then each CC is given a unique label across the image across each batch. I may be wrong here but won't maxpool give similar values across images in the batch. In that case the number of unique component labels might be lower than actual components. And what's the need to relabel the components as it is done here?
Also, why can't we use kornia's CC?

ashwinvaidya17 (approved):
I am not very familiar with this technique so I didn't understand some parts of it but other than that the PR looks good. Thanks!

anomalib/utils/metrics/pro.py
djdameln: The reason to re-label the components is that the `pro_score` function assumes that the components passed as the `comps` argument are labeled from 0 to N, where 0 is background and N is the total number of connected components in the batch. Because the connected component labeling algorithms that we use do not produce label maps that satisfy this requirement, we need to re-label them before passing them to the pro score computation. Since the cc labeling algorithms on cpu and gpu produce label maps with different characteristics, this re-labeling method also differs.

On cpu, we apply the connected component labeling on the image-level and then stack the labeled components back together in the batch dimension. So every connected component within the image will have a unique label, but labels will be duplicated across the batch. We need to re-label so that the labels are not only unique across the image, but also across the full batch.

On GPU, the connected component labeling works by assigning a unique label id to each pixel and then eliminating ids by merging connected regions together through iterative max pooling. This algorithm is applied on the full batch, so after the labeling, every connected component in the complete batch will have a unique label. So the number of unique values in the label map is equal to the number of connected components. However, due to the initial assignment of component id's [here](https://github.com/openvinotoolkit/anomalib/blob/2a29e535332fdd2e79a0546373714382f87a7b24/anomalib/utils/metrics/pro.py#L87), the components will not be labeled from 0 to N, but rather b. So we still need to re-label, but this time we can just create a mapping from the original ids to the [0, N] range.

djdameln: I just had a look at the Kornia implementation and it's essentially the same algorithm, so we could use the Kornia one instead (we'll still need to perform the re-labeling though).

ashwinvaidya17: Thanks for the detailed explanation. It is clear now 🙂

anomalib/utils/metrics/pro.py
```diff
@@ -0,0 +1,111 @@
...
+        """Compute the macro average of the PRO score across all regions in all batches."""
+        target = dim_zero_cat(self.target)
+        preds = dim_zero_cat(self.preds)
+
+        if target.is_cuda:
+            comps = connected_components_gpu(target.unsqueeze(1))
+        else:
+            comps = connected_components_cpu(target.unsqueeze(1))
+        if not len(comps.unique()) > comps.max():
+            print("here")
```
samet-akcay, line 43: I guess this is a left over?

anomalib/utils/metrics/pro.py
```diff
@@ -0,0 +1,111 @@
...
+
+    preds = comps.clone()
+    preds[~predictions] = 0
+    if n_comps == 1:  # only background
+        return torch.Tensor([1.0])
+    pro = recall(preds.flatten(), comps.flatten(), num_classes=n_comps, average="macro", ignore_index=0)
+    return pro
+
+
+def connected_components_gpu(binary_input: torch.Tensor, num_iterations: int = 1000) -> Tensor:
```
samet-akcay, line 73: Majority of the module uses `Tensor` to annotate the tensor variables, but this one, `binary_input` uses `torch.Tensor`. Would be good to stick to `Tensor` here as well.

samet-akcay (commented):
Thanks, only 2 minor comments.

anomalib/utils/metrics/pro.py
```diff
@@ -0,0 +1,111 @@
...
+        target = dim_zero_cat(self.target)
+        preds = dim_zero_cat(self.preds)
+
+        if target.is_cuda:
+            comps = connected_components_gpu(target.unsqueeze(1))
+        else:
+            comps = connected_components_cpu(target.unsqueeze(1))
+        if not len(comps.unique()) > comps.max():
+            print("here")
+            connected_components_cpu(target.unsqueeze(1))
```
samet-akcay, line 44: Why `connected_components_cpu(target.unsqueeze(1))` doesn't return a value. I thought this would be something like;

```
comps = connected_components_cpu(target.unsqueeze(1))
```

anomalib/utils/metrics/pro.py
```diff
@@ -0,0 +1,111 @@
+"""Implementation of PRO metric based on TorchMetrics."""
+from typing import List
```
samet-akcay, line 2: Would be good to leave a blank line between the docstring and the import

samet-akcay: Also, can we license the module?

anomalib/utils/metrics/pro.py
djdameln: This was a leftover from debugging. Removed it now

djdameln:
@samet-akcay I didn't realize this PR was still open. I just addressed your comments. Could you do another round of review?

samet-akcay (approved):
looking good 🔥, thanks

PR #509, opened 2022-08-18
Enable generic exporting of a trained model to ONNX or OpenVINO IR

ashishbdatta:
# Description

Allowing the user to export a model to either just onnx model or onnx model + openvino IR

## API Breakage 

This breaks backwards compatibility in the config file where a user used to apply OpenVINO optimization.

It goes from 

```yaml
optimization:
  openvino:
    apply: true
```

to

```yaml
optimization:
    export_mode: "onnx/openvino"
```

## Changes

- [X] Refactor (non-breaking change which refactors the code base)

## Checklist

- [X] I have added tests that prove my fix is effective or that my feature works


anomalib/utils/callbacks/__init__.py
```diff
@@ -114,18 +114,36 @@ def get_callbacks(config: Union[ListConfig, DictConfig]) -> List[Callback]:
...
+            callbacks.append(
+                ExportCallback(
+                    input_size=config.model.input_size,
+                    dirpath=config.project.path,
+                    filename="model",
+                    export_mode="onnx",
+                )
+            )
+        else:
+            logger.info("Export option not found")
```
ashwinvaidya17, line 146: Why is this info? Either we should throw an error or warn users that the option was incorrect and the training is proceeding without export callback.  Also, I prefer info to show what was passed `f"... option {config.optimization.export_mode} ..."` so that the users can see if they made a typo. Additionally, it would be nice to reiterate the options in the warning message.

ashwinvaidya17 (approved):
Thanks. I have one minor comment

anomalib/deploy/optimize.py
```diff
@@ -46,6 +46,7 @@ def export_convert(
     input_size: Union[List[int], Tuple[int, int]],
     onnx_path: Union[str, Path],
     export_path: Union[str, Path],
```
samet-akcay, line 48: Would it be possible to have a single path? Like remove `onnx_path` and keep `export_path` only?

anomalib/models/cflow/config.yaml
```diff
@@ -62,6 +62,9 @@ logging:
   logger: [] # options: [tensorboard, wandb, csv] or combinations.
   log_graph: false # Logs the model graph to respective logger.
 
+optimization:
+  export_mode: ""
```
samet-akcay, line 66: can we have `null` value here? and a comment showing the options? Something like:

```yaml
optimization:
   export_mode: null    # options: onnx, openvino
```

anomalib/models/dfkde/config.yaml
```diff
@@ -50,6 +50,8 @@ logging:
   logger: [] # options: [tensorboard, wandb, csv] or combinations.
   log_graph: false # Logs the model graph to respective logger.
 
+optimization:
+  export_mode: ""
```
samet-akcay, line 54: samet as above, and for the rest of the configuration files.

anomalib/utils/callbacks/__init__.py
```diff
@@ -114,18 +114,36 @@ def get_callbacks(config: Union[ListConfig, DictConfig]) -> List[Callback]:
                     export_dir=os.path.join(config.project.path, "compressed"),
                 )
             )
-        if "openvino" in config.optimization and config.optimization.openvino.apply:
-            from .openvino import (  # pylint: disable=import-outside-toplevel
-                OpenVINOCallback,
```
samet-akcay, line 119: Following this approach would cause a code duplication. We could perhaps assign `export_mode` in the if-statement, and then append `ExportCallback` only once? Does it make sense?

anomalib/utils/cli/cli.py
```diff
@@ -213,23 +215,37 @@ def __set_callbacks(self) -> None:
...
-                OpenVINOCallback(
+                ExportCallback(
                     input_size=config.data.init_args.image_size,
                     dirpath=os.path.join(config.trainer.default_root_dir, "compressed"),
                     filename="model",
+                    export_mode="onnx",
                 )
             )
+        else:
+            logger.info("No model export")
```
samet-akcay, line 248: Same duplication here

tests/pre_merge/deploy/test_inferencer.py
```diff
@@ -107,11 +107,12 @@ def test_openvino_inference(self, model_name: str, category: str = "shapes", pat
                 input_size=model_config.dataset.image_size,
                 onnx_path=export_path / "model.onnx",
                 export_path=export_path,
```
samet-akcay, line 109: I might be missing the point of having two paths here. Is there a specific reason?

tests/pre_merge/utils/callbacks/export_callback/test_export.py
```diff
@@ -0,0 +1,55 @@
...
+            val_check_interval=3,
+        )
+        trainer.fit(model, datamodule=datamodule)
+
+        if "openvino" in export_mode:
+            assert os.path.exists(os.path.join(tmp_dir, "openvino/model.bin")), "Failed to generate OpenVINO model"
+        elif "onnx" in export_mode:
+            assert os.path.exists(os.path.join(tmp_dir, "model.onnx")), "Failed to generate ONNX model"
+        else:
+            assert "Unknown export_mode"
```
samet-akcay, line 55: maybe we could do `raise ValueError(f"Unknown export_mode {export_mode}. Supported modes: onnx or openvino.")`

samet-akcay (commented):
Thanks for fixing this! I have few points, other than those, looks great!

anomalib/deploy/optimize.py
ashishbdatta: I think the idea in having two paths is for all of the OpenVINO files to be stored in a single directory while the ONNX file is on the same directory level as `weights`. @ashwinvaidya17 I see you wrote this function is that a correct understanding?

I like having two directories one for ONNX and one for OpenVINO, but I am open to refactoring if needed.

anomalib/utils/callbacks/__init__.py
ashishbdatta: Addressed in: [27282f2](https://github.com/openvinotoolkit/anomalib/pull/509/commits/27282f21e1cb9454e2fbf30764b58bc459df84f1)

anomalib/models/dfkde/config.yaml
ashishbdatta: addressed in: [83d0286](https://github.com/openvinotoolkit/anomalib/pull/509/commits/83d02868fbe31e9b4246a66fac33de1687d4ed5a)

tests/pre_merge/utils/callbacks/export_callback/test_export.py
ashishbdatta: Addressed in: [10497eb](https://github.com/openvinotoolkit/anomalib/pull/509/commits/10497ebe3c0ef626a23ef0e968dfcd44cddaa121)

anomalib/utils/callbacks/__init__.py
ashishbdatta: Addressed in: [27282f2](https://github.com/openvinotoolkit/anomalib/pull/509/commits/27282f21e1cb9454e2fbf30764b58bc459df84f1)

anomalib/utils/cli/cli.py
ashishbdatta: addressed in: [237f9ff](https://github.com/openvinotoolkit/anomalib/pull/509/commits/237f9ffedd31499f1b1779e2206cbed600161171)

anomalib/deploy/optimize.py
samet-akcay: if we want to have two path variables for onnx and openvino, then the name of `export_path` would be `openvino_path`. This is because `onnx` is also part of `export` 

> Maybe we can put all of them in a single folder and then see what feedback we get from the community.

Maybe we could have `export_path` as `Optional[Union[str, Path]] = None`. If None, it could be the same directory as the original weight file. Otherwise, we could save it to the `export_path`. thoughts?

ashishbdatta: Okay I tried a stab at it in: [a4b0604](https://github.com/openvinotoolkit/anomalib/pull/509/commits/a4b060460905a5e4ce7340f128a4ee8717c4cd66)

Let me know if this works. We still internal to `export_convert` function has `onnx_path` but nothing the user has to set from the config

djdameln (approved):
Thanks, this is a nice addition

anomalib/deploy/optimize.py
samet-akcay: Thanks, only a single comment

ashishbdatta: Fixed in [019460d](https://github.com/openvinotoolkit/anomalib/pull/509/commits/019460da3881f777ff5d3f2f9bc7495c641508af)

samet-akcay (approved):
Thanks, great addition to the repo!

PR #514, opened 2022-08-24
FIX: PaDiM didn't use config.model.pre_trained.

jingt2ch:
# Description

PaDim did not use 'pre_trained' value in config.yaml. This PR fixes it.

## Changes

## Checklist

### DIFF
--- a/anomalib/models/padim/lightning_model.py
+++ b/anomalib/models/padim/lightning_model.py
@@ -118,6 +118,7 @@ class PadimLightning(Padim):
             input_size=hparams.model.input_size,
             layers=hparams.model.layers,
             backbone=hparams.model.backbone,
+            pre_trained=hparams.model.pre_trained,
         )
         self.hparams: Union[DictConfig, ListConfig]  # type: ignore
         self.save_hyperparameters(hparams)


samet-akcay:
@jingt2ch, thanks for the fix. Looks like the rest of the models do not use pre_trained configuration neither. Would you like to modify those as well, or are you good with padim only?

jingt2ch:
@samet-akcay, thank for your review. I'm developing a PaDiM based model, so I'm good enough with PaDiM only.

I'm not sure about the other models. If I understand other models and need fix, I will send PR again.

samet-akcay (approved):

djdameln:
@jingt2ch thanks for fixing this. We're trying to understand the use case to determine if we should provide this fix for the other models as well. Would you mind explaining the added value of allowing a non pre-trained backbone in feature extraction based models such as PaDiM? Since the backbone in these types of models is not fine-tuned during training, the model weights would always remain at their randomly initialized values and the backbone would not be able to extract meaningful features.

jingt2ch:
@djdameln Sorry for my late reply.
In my use case, I wanted to evaluate the impact of backbone weights. When I evaluated my PaDiM based model, I found that the weights had too small effect. I thought there might be an implementation mistake in my model. So I used anomalib's pre_trained flag to evaluate original PaDiM.
I don't know about other models enough. However, someone may want to see the effect of pre_trained flag on other models as well.

Incidentally, even in the original PaDiM, the effect of untrained weights was smaller than expected.
image

djdameln:
@jingt2ch Thanks for clarifying. I guess it makes sense from a research perspective to compare the models with a non-pretrained backbone. It could be used as a sort of baseline against which to compare backbones trained on different datasets. In this respect it would make sense to allow this for the other Anomalib models as well.

> Incidentally, even in the original PaDiM, the effect of untrained weights was smaller than expected.

I noticed the same thing when I was investigating this, and was surprised by how well the PaDiM model performs with random weights. I guess without the pretrained weights the backbone provides a random non-linear dimensionality reduction of the input images that is still sufficiently descriptive to allow discriminating the classes.

PR #529, opened 2022-08-31
🔨 Pass `pre-trained` from config to `ModelLightning`

samet-akcay:
# Description

- Following #514, this PR adds `pre_trained` parameter to `ModelLightning` implementation.

## Changes

## Checklist

### DIFF
--- a/anomalib/models/cflow/lightning_model.py
+++ b/anomalib/models/cflow/lightning_model.py
@@ -185,6 +185,7 @@ class CflowLightning(Cflow):
             input_size=hparams.model.input_size,
             backbone=hparams.model.backbone,
             layers=hparams.model.layers,
+            pre_trained=hparams.model.pre_trained,
             fiber_batch_size=hparams.dataset.fiber_batch_size,
             decoder=hparams.model.decoder,
             condition_vector=hparams.model.condition_vector,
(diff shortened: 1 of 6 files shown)


djdameln (commented):
I'm having some doubts about adding this parameter. For feature extraction based models such as padim and dfm, what would be the use case of having a non pre-trained backbone? Since the training does not involve finetuning the backbone weights, the normality model would be built upon random layer activations.

samet-akcay:
> I'm having some doubts about adding this parameter. For feature extraction based models such as padim and dfm, what would be the use case of having a non pre-trained backbone? Since the training does not involve finetuning the backbone weights, the normality model would be built upon random layer activations.

I don't know if there is a specific use-case for this. config.yaml file contains pre_trained flag. If we dont want this, maybe we could remove this from the config file itself.

Perhaps you could ask this question in PR 514 to confirm whether this actually has any use-case?

djdameln (approved):
Based on the discussion in the other PR (https://github.com/openvinotoolkit/anomalib/pull/514) I think this is a viable use case, so I'm happy to merge this one.

ashwinvaidya17 (approved):
Thanks!

Issue #523, opened 2022-08-29
Error(s) in loading state_dict for PatchcoreLightning


Luksalos:
**Describe the bug**

Running  `tools/inference/torch_inference.py` throws:

```
Traceback (most recent call last):
  File "/workspace/tools/inference/torch_inference.py", line 97, in <module>
    infer()
  File "/workspace/tools/inference/torch_inference.py", line 73, in infer
    inferencer = TorchInferencer(config=args.config, model_source=args.weights)
  File "/workspace/anomalib/deploy/inferencers/torch_inferencer.py", line 54, in __init__
    self.model = self.load_model(model_source)
  File "/workspace/anomalib/deploy/inferencers/torch_inferencer.py", line 85, in load_model
    model.load_state_dict(torch.load(path)["state_dict"])
  File "/opt/conda/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1497, in load_state_dict
    raise RuntimeError('Error(s) in loading state_dict for {}:\n\t{}'.format(
RuntimeError: Error(s) in loading state_dict for PatchcoreLightning:
	Unexpected key(s) in state_dict: "normalization_metrics.min", "normalization_metrics.max".
```

**To Reproduce**
1) Using default `patchcore/config.yaml`
2) Train the model: `python tools/train.py --config anomalib/models/patchcore/config.yaml`
3) Use trained model for `torch_inference`: 
```
python tools/inference/torch_inference.py \
--config anomalib/models/patchcore/config.yaml \
--weights results/patchcore/mvtec/bottle/weights/model-v14.ckpt \
--input datasets/MVTec/bottle/test/broken_large/000.png \
--output results/patchcore/mvtec/bottle/images
```

**Additional context**

Error is present since commit d6951ebaaf36477bf245b17b5b6a563d35892e81

djdameln:
@ashwinvaidya17 can you have a look at this? It seems to be related to the normalization redesign

alevangel:
I'm having a simlar issue, here the runtime error:

```
RuntimeError: Error(s) in loading state_dict for PatchcoreLightning:
[...]

PR #530, opened 2022-08-31
🐞 Set normalization method from anomaly module

ashwinvaidya17:
# Description

- Loads the correct normalization method from anomaly module whenever the model weights are loaded before the normalization callback is called.

- Proposal to address #523


samet-akcay (approved):
Looking good for now, thanks. I was wondering how we are going to handle when you implement more normalization methods. Adding those methods in `_load_normalization_class`?

djdameln (approved):

Issue #277, opened 2022-04-25
Support training with only normal images (no evaluation)


djdameln:
Some users would like to train models with only normal images available, see https://github.com/openvinotoolkit/anomalib/discussions/192 for example. Our current data classes do not support this and always expect a folder of anomalous images to be provided for evaluation purposes. I feel it would be good to add support for training without evaluation.

innat:
@djdameln 
In such cases, where all data are good (when task=segmentation, pixel mask all 0), what will be the metrices to evaluate model performance? (Partially [related](https://github.com/openvinotoolkit/anomalib/discussions/689)).

djdameln:
This has been addressed on the datamodules feature branch.

PR #558, opened 2022-09-14
Refactor data modules

djdameln:
# Description

- This PR refactors the data modules to reduce duplication.
- The `AnomalibDataset` class was introduced, which replaces the individual dataset classes.
- The base class `AnomalibDataModule` was introduced, and the `setup` method was moved from the concrete subclasses to the new base class.
- The `make_mvtec_dataset`, `make_btech_dataset` and `make_dataset` functions are replaced by the `_create_samples` method. The implementation of this method differs between the individual data module classes.

- The refactor also makes it possible to train when no anomalous images are present in the dataset. In this case, Anomalib will not run the test stage after training, so no performance metrics are reported. The user will be notified about this through a logging message.

- Fixes https://github.com/openvinotoolkit/anomalib/issues/277

## Changes

## Checklist

### DIFF
--- a/anomalib/data/base.py
+++ b/anomalib/data/base.py
@@ -33,7 +33,7 @@ class AnomalibDataset(Dataset):
 
     def contains_anomalous_images(self):
         """Check if the dataset contains any anomalous images."""
-        return "anomalous" in list(self.samples.label)
+        return 1 in list(self.samples.label_index)
 
     def __len__(self) -> int:
         """Get length of the dataset."""
--- a/anomalib/data/base.py
+++ b/anomalib/data/base.py
@@ -124,6 +124,7 @@ class AnomalibDataModule(LightningDataModule, ABC):
         | 0 | path/to/image.png | anomalous | 0           | path/to/mask.png | train |
         |---|-------------------|-----------|-------------|------------------|-------|
         """
+        raise NotImplementedError
 
     def setup(self, stage: Optional[str] = None) -> None:
         """Setup train, validation and test data.
--- a/anomalib/data/base.py
+++ b/anomalib/data/base.py
@@ -32,10 +32,6 @@ class AnomalibDataset(Dataset):
         self.split = split
         self.pre_process = pre_process
 
-    def contains_anomalous_images(self):
-        """Check if the dataset contains any anomalous images."""
-        return 1 in list(self.samples.label_index)
-
     def __len__(self) -> int:
         """Get length of the dataset."""
         return len(self.samples)
@@ -109,6 +105,8 @@ class AnomalibDataModule(LightningDataModule, ABC):
         self.val_data: Optional[AnomalibDataset] = None
         self.test_data: Optional[AnomalibDataset] = None
 
+        self._samples: Optional[DataFrame] = None
+
     @abstractmethod
     def _create_samples(self) -> DataFrame:
         """This method should be implemented in the subclass.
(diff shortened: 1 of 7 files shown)


anomalib/data/base.py
```diff
@@ -0,0 +1,165 @@
...
+        if self.split in ["val", "test"]:
+            label_index = self.samples.label_index[index]
+
+            item["image_path"] = image_path
+            item["label"] = label_index
+
+            if self.task == "segmentation":
+                mask_path = self.samples.mask_path[index]
+
+                # Only Anomalous (1) images has masks in MVTec AD dataset.
```
ashwinvaidya17, line 67: ~MVTec AD~

anomalib/data/base.py
```diff
@@ -0,0 +1,165 @@
...
+        Additionally, when the task type is segmentation, the dataframe must have the mask_path column, which contains
+        the path the ground truth masks (for the anomalous images only).
+
+        Example of a dataframe returned by calling this method from a concrete class:
+        |---|-------------------|-----------|-------------|------------------|-------|
+        |   | image_path        | label     | label_index | mask_path        | split |
+        |---|-------------------|-----------|-------------|------------------|-------|
+        | 0 | path/to/image.png | anomalous | 0           | path/to/mask.png | train |
+        |---|-------------------|-----------|-------------|------------------|-------|
+        """
```
ashwinvaidya17, line 126: Should we add `raise NotImplementedError()`

ashwinvaidya17 (commented):
Thanks the code is much cleaner now. I have two very minor comments.

anomalib/data/base.py
```diff
@@ -0,0 +1,166 @@
...
+                mask_path = self.samples.mask_path[index]
+
+                # Only Anomalous (1) images have masks in anomaly datasets
+                # Therefore, create empty mask for Normal (0) images.
+                if label_index == 0:
+                    mask = np.zeros(shape=image.shape[:2])
+                else:
+                    mask = cv2.imread(mask_path, flags=0) / 255.0
+
+                pre_processed = self.pre_process(image=image, mask=mask)
```
samet-akcay, line 74: pre-processing is also applied in L55 for image. Does this mean, we apply the pre-processing twice? Would this be efficient?

jpcbertoldo (commented):
nice refactor :)
i just wanted to give my 10 cents with two minor improvement suggestions (althogh the one about setup may not be _that_ minor)

anomalib/data/base.py
```diff
@@ -0,0 +1,166 @@
...
+            val_samples = samples[samples.split == "val"]
+            val_samples = val_samples.reset_index(drop=True)
+            self.val_data = AnomalibDataset(
+                samples=val_samples,
+                split="val",
+                task=self.task,
+                pre_process=self.pre_process_val,
+            )
+
+        test_samples = samples[samples.split == "test"]
```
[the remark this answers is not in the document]
djdameln: Thanks, that's a good point. The reason we didn't account for the "validate" stage is that we never specifically call `trainer.validate()` in our entrypoint scripts. But looking at the implementation now, I agree that it would be better to add a case for this stage specifically.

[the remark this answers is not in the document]
djdameln: I've added the case for the validate stage. While implementing this I also noticed that the test dataset was always created on every `setup()` call, so even in the "fit" stage. After fixing this I had to do some additional refactoring to make sure that we can still check the test set for the presence of anomalous images (so not that minor indeed :smile:).

@samet-akcay @ashwinvaidya17 I requested another round of review from both of you, because addressing this required some changes to the design of the base class.

samet-akcay (approved):
Thanks, a lot more cleaner

samet-akcay (approved):

anomalib/data/base.py
```diff
@@ -0,0 +1,207 @@
...
+        """
+        assert self._samples is not None, "Samples have not been created yet."
+        if split is None:
+            return self._samples
+        samples = self._samples[self._samples.split == split]
+        return samples.reset_index(drop=True)
+
+    def setup(self, stage: Optional[str] = None) -> None:
+        """Setup train, validation and test data.
+
```
jpcbertoldo, line 150:
```suggestion

If `stage` is `None`, all splits are setup.
```

anomalib/data/base.py
```diff
@@ -0,0 +1,207 @@
...
+            samples = self.get_samples("train")
+            self.train_data = AnomalibDataset(
+                samples=samples,
+                split="train",
+                task=self.task,
+                pre_process=self.pre_process_train,
+            )
+
+        if stage in (None, "fit", "validate"):
+            samples = self.get_samples("val") if self.create_validation_set else self.get_samples("test")
```
jpcbertoldo, line 167:
```suggestion

            if `self.create_validation_set`:
                samples = self.get_samples("val")
            else:
                warnings.warn("Validation split is not availabe. Test split will be used for validation.") 
                samples = self.get_samples("test")
```

jpcbertoldo: if you decide to move `_create_samples` to `AnomalibDataset` then this won't be necessary here

anomalib/data/base.py
```diff
@@ -0,0 +1,207 @@
...
+
+        Args:
+            split (str): the subset of interest ("train", "val" or "test"). When left empty, the full dataset will be
+                checked.
+
+        Returns:
+            bool: Boolean indicating if any anomalous images have been assigned to the dataset or subset.
+        """
+        samples = self.get_samples(split)
+        return 1 in list(samples.label_index)
```
jpcbertoldo, line 195:
```suggestion
        return LABEL_INDEX_ANOMALOUS in list(samples.label_index)
```
a bit picky but a constant would clearer

jpcbertoldo (commented):
`AnomalibDataset.__getitem__` is better without those conditions on `self.split` IMO, tnx : )

---

i noticed a few minor things (i marked "[minor]" in my comments) but i have a bigger concern about the `AnomalibDataModule._create_samples` -- feel free to disconsider if you want to move forward with this sooner 

a conceptual question: is an `AnomalibDataset` assumed to be split-specific?

if yes (IMO it should be), then the  `self._samples.split == split` should be encapsulated somewhere in the `AnomalibDataset` (I'd put `._create_samples` as an abstract method inside  `AnomalibDataset`) and the dataframes should be split-specific (i.e. no need for a `split` column)

`AnomalibDataModule.get_samples(split)` would be just a multiplexer:

```
if split == "train":
    return self.train_data.samples  # which btw could be a `@property` hidding `_samples` 
``` 

another reason to do this: the `AnomalibDataset` would be more standalone (in the current form, one needs to instantiate the `AnomalibDataModule` in order to construct the former properly, or at least without hacking the latter)

jpcbertoldo:
i created a sketch of the things i suggested in #564 
i only modified `base.py` to be clear about the things i said above
feel free to consider it or not but if you do then the necessary modifications should be done in the downstream classes (mvtec, folder, and btech)

anomalib/data/base.py
ashwinvaidya17: In this case would using enums be better? We can then have something like `Label.Normal` and `Label.Anomalous`

ashwinvaidya17 (approved):
I am fine with the changes

Issue #440, opened 2022-07-18
export patchcore to onnx error


mrljwlm:
RuntimeError: Exporting the operator cdist to ONNX opset version 11 is not supported

alexriedel1:
You could replace the `torch.cdist` by a simple implementation of the euclidian distance like this one: https://discuss.pytorch.org/t/understanding-cdist-function/76296/13

```
def my_cdist(x1, x2):
    x1_norm = x1.pow(2).sum(dim=-1, keepdim=True)
    x2_norm = x2.pow(2).sum(dim=-1, keepdim=True)
    res = torch.addmm(x2_norm.transpose(-2, -1), x1, x2.transpose(-2, -1), alpha=-2).add_(x1_norm)
    res = res.clamp_min_(1e-30).sqrt_()
    return res
```
I guess the operators are all supported in onnx. 


or maybe even just: https://discuss.pytorch.org/t/efficient-distance-matrix-computation/9065/5

ORippler:
Just a quick update from my side:

support for `torch.cdist` was added in PyTorch 1.12 by partitioning/refactorign `cdist` via supported ops in onnx (see [this PR](https://github.com/pytorch/pytorch/pull/75278), so export to onnx should work and does work.

However, the exports of all models that use gaussian blur from kornia fails in 1.12 (due to dynamic determination of kernel_size).

Side note 2:

You won't be able to convert patchcore for application on VPUs though, see [here](https://github.com/openvinotoolkit/openvino/issues/11598)

fujikosu:
@ORippler Thanks for the info here! I tried to install pytorch 1.12 with anomalib 0.35. But pytorch version was automatically degraded to 1.11. I suspect the reason why is because that newest pytorch version is only compatible with torchvision 0.13 for now according to [this page](https://pypi.org/project/torchvision/).

torch | torchvision | python
-- | -- | --
main / nightly | main / nightly | >=3.7, <=3.10
1.11.0 | 0.12.0 | >=3.7, <=3.10
1.10.2 | 0.11.3 | >=3.6, <=3.9

anomalib has currently torchvision>=0.9.1,<=0.12.0 as dependency set in [requirements/base.txt](https://github.com/openvinotoolkit/anomalib/blob/d6951ebaaf36477bf245b17b5b6a563d35892e81/requirements/base.txt)
[...]

PR #565, opened 2022-09-19
Add patchcore to openvino export test + upgrade lightning

ashwinvaidya17:
# Description

We had limited torchvision to 0.12.0 because it ended up installing torch 1.12.0. (PR https://github.com/openvinotoolkit/anomalib/pull/396). From my experiments right now everything seems to be working. The tests pass locally. So hopefully this should address patchcore onnx export (https://github.com/openvinotoolkit/anomalib/issues/440) for those using torch==1.12.0


samet-akcay (approved):
thanks

djdameln (approved):

PR #570, opened 2022-09-21
Refactor `PreProcessor` and fix `Visualizer` denormalization issue.

### LINKED ISSUE #512: Allow custom Data Normalization and consider it in Post-processing

By default (without warning the user!), if no `transform_config_{train,val}` is supplied to the datamodule, the code uses ImageNet statistics to normalize the images, which may be a good default for the included datasets, but should print some warning to the developer on custom data.

At the Post-Processing step, e.g. when visualizing the images using the `ImageVisualizerCallback`, `Denormalize()(batch["image"][i].cpu())` is applied in `Visualizer.visualize_batch` ([anomalib/anomalib/post_processing/visualizer.py:89](https://github.com/openvinotoolkit/anomalib/blob/ac5546261784daa81bcec4a2bc86a499555c94cf/anomalib/post_processing/visualizer.py#L78)), which again defaults to the ImageNet statistics. But unlike the input datamodule, the visualizer cannot be configured to use the normalization-configuration as configured in the input-dataset, which leads to the images being "denormalized" to different ranges so that $denorm(norm(image)) \ne image$

- Add parameter to Visualizer to configure Normalization
- Or: automatically set parameters on callback when `transform_config_{train,test} != None` is discovered in datamodule
- Warn user (at least for custom datasets), that by default, ImageNet Normalization is applied, when `transform_config_{train,val} is None` 

**Describe alternatives you've considered**

- ImageNet normalization may be great to reproduce paper results, but when used in other projects, custom input data normalization is common.

**Additional context**

samet-akcay:
# Description

- Fix `Visualizer` denormalization issue.
- Refactor `PreProcessor` to improve the flexibility of getting the transforms without relying on a `PreProcessor` object.
- Refactor `read_image` to add flexibility of resizing the image while reading. 

- Fixes #512 

## Changes

## Checklist


anomalib/pre_processing/pre_process.py
```diff
@@ -7,11 +7,111 @@
...
 
 import albumentations as A
 from albumentations.pytorch import ToTensorV2
 
+from anomalib.data.utils import get_image_height_and_width
+
+logger = logging.getLogger(__name__)
+
+
+def get_transforms(
```
ashwinvaidya17, line 21: Should we also move this function to `data/utils`?

ashwinvaidya17 (approved):
Thanks I don't have anything to add here

djdameln (approved):
As discussed in the meeting, we could probably deprecate the PreProcessor class now that it's reduced to a wrapper around the `get_transforms` function. It would require some changes to the datamodule, so let's address this once the datamodule refactor has been merged.

anomalib/pre_processing/pre_process.py
samet-akcay: i think this is still one of the most important components of `pre-processing` sub-package. If we move it to `data/utils`, then `pre_processing` would become quite a weak package that we could consider removing in the future

djdameln: In that case we could ask ourselves if we really need the `pre-processing` module

samet-akcay: yes, that's the question, I think

samet-akcay: I'm merging this one, we could discuss this later on

PR #572, opened 2022-09-22
New datamodules design

### LINKED ISSUE #277: Support training with only normal images (no evaluation)

djdameln:
# Description

- This is a proof of concept for a new datamodule design. The base classes and all three datasets (MVTec, Folder, BTech) have been updated according to the new design.
- Fixes #277 

# Summary of the design
- The `AnomalibDataset` and its subclasses now have a `setup` method, which is called from `DataModule._setup()`. When called, the dataframe will be created by using the `create_dataset` function. `setup` must be called before the dataset class can be used. This is so that we can instantiate the dataset in the constructor of the datamodule, before `prepare_data` is called.
- The `DataModule` is responsible for performing the random subset splitting. It first creates the fixed subsets in the constructor, and then performs any additional subset splitting in `DataModule._setup()`.
- The `_setup` method from the `DataModule` is called only once, and is independent of the `stage` argument.
- To facilitate the dynamic subset splitting on the `DataModule` side, helper functions can be added to the data utils (see `concatenate_datasets` and `random_split` for an example).

## Responsibilites of the different classes:
- `create_dataset` functions: Create a dataframe with information about the samples, including any information about the fixed train/val/test split that may follow from the folder organization or annotation files of the dataset.
- `AnomalibDataset` and subclasses: Prepare dataset items and ground truth labels + masks to be used as model input.
- `AnomalibDataModule`: Create dataloaders and perform dynamic subset splitting.

## Advantages
- Clear responsibility of the different classes.
- The Dataset classes can be used standalone.
- No seeds needed, because dynamic subset splitting is done only once.

## Known issues
- ~~Some terminology is still a bit confusing, e.g. inconsistent use of `val`, `infer` and `test` (`tranform_config_val` vs `pre_process_infer` vs `test_batch_size`)~~
- ~~Tests need to be updated.~~


anomalib/data/base.py
```diff
@@ -107,92 +176,34 @@ def __init__(
...
         if stage in (None, "fit", "validate"):
-            samples = self.get_samples("val") if self.create_validation_set else self.get_samples("test")
-            self.val_data = AnomalibDataset(
-                samples=samples,
-                split="val",
-                task=self.task,
-                pre_process=self.pre_process_val,
-            )
-
+            if self.contains_anomalous_images("val"):
```
ashwinvaidya17, line 201: Maybe we should use `subset.val` here as this calls `get_samples` which has subset as it's parameter

anomalib/data/base.py
```diff
@@ -107,92 +176,34 @@ def __init__(
...
-            samples = self.get_samples("train")
-            self.train_data = AnomalibDataset(
-                samples=samples,
-                split="train",
-                task=self.task,
-                pre_process=self.pre_process_train,
-            )
-
+            self.train_data = self.data.train_subset(pre_process=self.pre_process_train)
         if stage in (None, "fit", "validate"):
```
ashwinvaidya17, line 200: Maybe to make it easier for new users we should add a comment stating that if the validation set does not have anomalous images, it uses the test set

ashwinvaidya17 (commented):
I went over it and I don't have any major feedback. I feel this is a much cleaner design and more flexible.

jpcbertoldo (commented):
(i'm going to be a bit annoying because i think this refactor is important, and it will my work a lot)

this version is better than the other one IMO, but overall both are creating divergences with how `LightningDatamodule` is supposed to be used (and how their downstream `Dataset` is supposed to be designed)

There are some advantages in this design but I think following lightnings's patterns is a better way to go because then you make better use of it

Another guideline i'm considering: the behavior of parent classes should be kept minimal (or use the template method pattern, like lightning) to allow more composability.

i am trying to keep this short but below i give you 2 reasons to not go this way

i wrote another draft in my fork (didn't open another PR yet): [my branch](https://github.com/jpcbertoldo/anomalib/tree/jpcbertoldo/datamodule-refactor-draft)

- [base.py](https://github.com/jpcbertoldo/anomalib/blob/4b3c95e3a48b8ddc1f2a84ace8a6aa8ae94ce972/anomalib/data/base.py)
- [mvtec.py](https://github.com/jpcbertoldo/anomalib/blob/4b3c95e3a48b8ddc1f2a84ace8a6aa8ae94ce972/anomalib/data/mvtec.py) (got much cleaner)

---

**reason 1**

the order of things is not quite right

example:

1. `LightningDatamodule.prepare_data()` is supposed to not change state and only manage the files in the system to make sure they`re in place
2. `LightningDatamodule.setup()` should be loading whatever necessary thing to the memory

In the way things are being done here (correct me if i'm wrong), the setup is being made even before the `prepare_data()`.

well, because of (2), the `Dataset` somehow has to load things to the memory either (a) lazyly (following `LightningDatamodule`'s pattern) or (b) it should be instantiated at `LightningDatamodule.setup()`, which means that the data preparation should not be a method (but a staticmethod could do).

in my draft i solved that by making the `Dataset` follow exactly the same pattern:  `.prepare_data()` (which is calls a `staticmethod`) and `.setup()` (which will make use of the args from `__init__`).

---

**reason 2**

i think the current design is biased by the behavior in `folder.py` and forcing to fit the (already functional) code of the `make_datate`-like functions

from what i understan, a `Dataset` should keep the knowlege of how to find and load samples from a specific pre-defined split; it should *not* deal with dynamic splitting (i.e. creating random subsplits to create a validation set)

i think that kind of behavior should be at the `Datamodule` level because it can ensure the compatibiilty between the splits when they've all already been setup

in my draft i used `torch.utils.data.Subset` to do that and transfered all this behavior to `AnomalibDatamodule`; two advantages:
- downstream class (see `mvtec.py`) become cleaner
- the behavior is better factorized out 

you will see that, for instance, `seed` and `create_validation_set`  (smt. like that) dont need to be passed downstream to the child class (then to the function)

jpcbertoldo:
@djdameln i'm very willing to help on this refactor in particular but unfortunately i'm struggling to find time to create a neat and well explained PR like yours

maybe can we make a call? i think it would be more productive

djdameln:
@jpcbertoldo Thanks for your comments. As it turns out, it's quite tricky to get this design right, so it's nice to have an extra pair of eyes on this. I agree with some, but not all of your suggestions.

> In the way things are being done here (correct me if i'm wrong), the setup is being made even before the prepare_data().

The call to `create_dataset` can be made at the end of the `prepare_data` method, so in this case we would create the samples after downloading the data. But yeah, I do agree that this does not really fit the intended use of `prepare_data` and `setup`.

> in my draft i solved that by making the Dataset follow exactly the same pattern: .prepare_data() (which is calls a staticmethod) and .setup() (which will make use of the args from __init__).

I had a look at your draft and correct me if I'm wrong, but I don't think it's necessary to have the implementation of `prepare_data` in the dataset class. We could keep this functionality in the datamodule. That way it would be called only once for the entire datamodule, instead of three times (once for every subset). 

I do see the added value of moving the setup method to the dataset class, because it allows us to instantiate the dataset object in the constructor of the datamodule. That way we can get rid of the awkward `create_dataset` method of my design.

> you will see that, for instance, seed and create_validation_set (smt. like that) dont need to be passed downstream to the child class (then to the function)

This is true for MVTec, where the train/test set is fixed, but not for the Folder dataset where we create a random train/test split at runtime. When following your design, we would still need to ensure somehow that the same seed is used between `train_data.setup` and `test_data.setup`.

This was actually the main motivation behind my latest design. By starting  with a common dataset object with the 'Full' split, we ensure that we only have to call `create_samples` once, so there's no need to pass the seed around.

Anyway let's continue the discussion in a call. I'll schedule one for early next week.

ashwinvaidya17:
I went over the new design. It looks good to me from an overall perspective. I have a few minor preferences such as warning messages and all but that can be part of the code review. But the design is good.

anomalib/models/cflow/config.yaml
```diff
@@ -13,7 +13,7 @@ dataset:
   transform_config:
     train: null
     val: null
-  create_validation_set: false
+  validation_split_mode: same_as_test
```
samet-akcay, line 16: It might be an idea to list the other options here as well.

samet-akcay: and for the other configs.

anomalib/models/padim/config.yaml
```diff
@@ -58,7 +58,7 @@ logging:
   log_graph: false # Logs the model graph to respective logger.
 
 optimization:
-  export_mode: null #options: onnx, openvino
+  export_mode: openvino #options: onnx, openvino
```
samet-akcay, line 61: Is this intentional?

samet-akcay (commented):
Thanks, it's a lot better now !

anomalib/data/__init__.py
```diff
@@ -7,17 +7,18 @@
...
 
-from .btech import BTech
-from .folder import Folder
+from anomalib.data.base import AnomalibDataModule
+
+from .btech import BTechDataModule
+from .folder import FolderDataModule
 from .inference import InferenceDataset
-from .mvtec import MVTec
+from .mvtec import MVTecDataModule
```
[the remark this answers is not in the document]
djdameln: I didn't realize this naming was chosen for the CLI. I changed it because I felt the name itself was more suitable for the dataset class, not the datamodule class. It's also more in line with torchvision naming convention, where the dataset class gets the name of the dataset. But I'll revert this change then.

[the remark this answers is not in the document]
samet-akcay: I'm not against it. I'm just wondering whether adding `DataModule` suffix would be user-friendly. If yes, we could keep it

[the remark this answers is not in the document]
djdameln: No I agree. For CLI usage it would be better to just use `--data=MVTec`

anomalib/models/cflow/config.yaml
```diff
@@ -6,14 +6,14 @@ dataset:
...
+  eval_batch_size: 16
   inference_batch_size: 16
   fiber_batch_size: 64
   num_workers: 8
   transform_config:
     train: null
-    val: null
-  create_validation_set: false
+    eval: null
+  validation_split_mode: same_as_test # options: [same_as_test, from_test]
```
samet-akcay, line 16: Do you think if this will be informative enough for the user? Would be need some explanation somewhere?

samet-akcay (commented):
Shaping up well, thanks. Single comment left only

anomalib/models/cflow/config.yaml
djdameln: I'll add some description to the docs

samet-akcay (approved):
I think I'm happy with the changes.

ashwinvaidya17 (approved):
Thanks for the efforts! Let's merge this ⚡

djdameln:
@jpcbertoldo we're about to merge this one. Do you have any comments on the design? If not, we will go ahead and merge this. We could always do another round of refactor later if needed.

jpcbertoldo:
> @jpcbertoldo we're about to merge this one. Do you have any comments on the design? If not, we will go ahead and merge this. We could always do another round of refactor later if needed.

hey, sorry i couldn't give enough attention to this lately, i'm reading it now but feel free to merge if you want to speed it up

anomalib/data/utils/split.py
```diff
@@ -11,76 +11,97 @@
...
+                "Zero subset length encountered during splitting. This means one of your subsets might be"
+                " empty or devoid of either normal or anomalous images."
+            )
+        # perform random subsampling
+        indices = torch.randperm(len(label_dataset))
+        subsets.append(
+            [label_dataset.subsample(subset_indices) for subset_indices in torch.split(indices, subset_lengths)]
+        )
+
+    # concatenate and return
```
jpcbertoldo, line 105:
```suggestion
    # invert outter/inner lists
    # outter list: subsets with the given ratio, inner list: per-label unique
```

anomalib/data/utils/split.py
```diff
@@ -11,76 +11,97 @@
...
+
+    # create list of source data
+    if label_aware:
+        indices_per_label = [group.index for _, group in dataset.samples.groupby("label_index")]
+        per_label_datasets = [dataset.subsample(indices) for indices in indices_per_label]
+    else:
+        per_label_datasets = [dataset]
+
+    # split each (label-aware) subset of source data
+    subsets = []
```
jpcbertoldo, line 85: just to make it easier to follow
```suggestion
    # outter list: per-label unique, inner list: random subsets with the given ratio 
    subsets: List[List[AnomalibDataset]] = []
```

anomalib/data/utils/split.py
```diff
@@ -11,76 +11,97 @@
...
 
-    indices_to_sample = random.sample(population=normal_test_image_indices, k=num_normal_valid_images)
-    samples.loc[indices_to_sample, "split"] = "val"
 
-    # Split abnormal images.
-    abnormal_test_image_indices = samples.index[(samples.split == "test") & (samples.label != normal_label)].to_list()
-    num_abnormal_valid_images = len(abnormal_test_image_indices) // 2
+def random_split(
+    dataset: AnomalibDataset, split_ratio: Union[float, Sequence[float]], label_aware: bool = False
+) -> List[AnomalibDataset]:
```
jpcbertoldo, line 59: would be nice to have a `seed` argument here (`torch.randperm` can be replaced by `random.Random.shuffle` if that is easier)

tests/pre_merge/datasets/test_dataset.py
```diff
@@ -1,243 +1,74 @@
...
+        """Test the random subset splitting."""
+
+        # subset splitting
+        subsets = random_split(folder_dataset, [0.4, 0.35, 0.25])
+        # check if subset splitting has been performed correctly
+        assert len(subsets) == 3
+        # reconstruct the original dataset by concatenating the subsets
+        reconstructed_dataset = concatenate_datasets(subsets)
+        # check if reconstructed dataset is equal to original dataset
+        assert folder_dataset.samples.equals(reconstructed_dataset.samples)
```
jpcbertoldo, line 55: picky comment: not a great pratice to test your lib by using another util from your own lib 

for instance

```suggestion
        folder_dataset_indices = set(folder_dataset.samples.index)
        subsets_indices = set.union(*[set(ss.samples.index) for ss in subsets])
        assert folder_dataset_indices == subsets_indices, f"diff {folder_dataset_indices.symmetric_difference(subsets_indices)}"
```

i also added this message `f"diff {folder_dataset_indices.symmetric_difference(subsets_indices)}"` to make it readily debuggable as a suggesttion but maybe it's discussible wether this should happen in a test suite (i believe so because it makes easier to dig into errors but the code will grow a bit if this is done everywhere)

jpcbertoldo (commented):
i started with the easier parts
now i'm going to read `anomalib/data/base/datamodule.py`, `.../base/dataset.py`, and their children classes

anomalib/models/cflow/config.yaml
jpcbertoldo:
```suggestion
  val_split_mode: same_as_test # see options at anomalib.data.utils.ValSplitMode
```

[minor]
like this new users would check all the updated options 
e.g. with your other pr there will be a 3rd option

anomalib/data/base/dataset.py
```diff
@@ -0,0 +1,155 @@
...
+        assert isinstance(other_dataset, self.__class__), "Cannot concatenate datasets that are not of the same type."
+        assert self.is_setup and other_dataset.is_setup, "Cannot concatenate uninitialized datasets. Call setup first."
+        dataset = copy.deepcopy(self)
+        dataset.samples = pd.concat([self.samples, other_dataset.samples], ignore_index=True)
+        return dataset
+
+    def setup(self) -> None:
+        """Load data/metadata into memory."""
+        if not self.is_setup:
+            self._setup()
```
jpcbertoldo, line 133: shouldn't it be

```suggestion
            self.samples = self._setup()
```
?

anomalib/data/base/dataset.py
```diff
@@ -0,0 +1,155 @@
...
+            item["mask"] = pre_processed["mask"]
+        else:
+            raise ValueError(f"Unknown task type: {self.task}")
+        item["image"] = pre_processed["image"]
+
+        return item
+
+    def __add__(self, other_dataset: AnomalibDataset) -> AnomalibDataset:
+        """Concatenate this dataset with another dataset."""
+        assert isinstance(other_dataset, self.__class__), "Cannot concatenate datasets that are not of the same type."
```
jpcbertoldo, line 124: `isinstance` will yield a tricky behavior because it considers the parent class (`AnomalibDataset`) of `other_dataset` (see example)
i'm not sure it is intentional

---
a simplified example: 

```
class Parent:
    def __add__(self, other: "Parent") -> "Parent":
	    assert isinstance(other, self.__class__)

class Son(Parent):
    pass

class Daughter(Parent):
   pass

p = Parent()
s = Son()
d = Daughter()

s + d  # (1) breaks
s + p  # (2) breaks
p + s  # (3) does NOT break
```

I believe you wanted to prevent (1). 
Although cases (2) and (3) may also happen and they are assymetric.

anomalib/data/base/dataset.py
```diff
@@ -0,0 +1,155 @@
...
+        assert self.is_setup and other_dataset.is_setup, "Cannot concatenate uninitialized datasets. Call setup first."
```
jpcbertoldo, line 125: this way the user knows which one is not setup
```suggestion
        assert self.is_setup, "Cannot concatenate uninitialized datasets. Call setup first."
        assert other_dataset.is_setup, "Cannot concatenate uninitialized datasets. Call setup first."
```

anomalib/data/base/dataset.py
```diff
@@ -0,0 +1,155 @@
...
+        """Get dataset item for the index ``index``.
+
+        Args:
+            index (int): Index to get the item.
+
+        Returns:
+            Union[Dict[str, Tensor], Dict[str, Union[str, Tensor]]]: Dict of image tensor during training.
+                Otherwise, Dict containing image path, target path, image tensor, label and transformed bounding box.
+        """
+        assert isinstance(self._samples, DataFrame)
```
jpcbertoldo, line 92:
```suggestion
```
should be tested in the setter, here it's too late (harder to debug)

anomalib/data/base/dataset.py
```diff
@@ -0,0 +1,155 @@
...
+        return self._samples
+
+    @samples.setter
+    def samples(self, samples: DataFrame):
+        """Overwrite the samples with a new dataframe.
+
+        Args:
+            samples (DataFrame): DataFrame with new samples.
+        """
+        self._samples = samples.sort_values(by="image_path", ignore_index=True)
```
jpcbertoldo, line 70: the class depends a lot on `samples` so it would be easier to debug some basic expectations about it

```suggestion
        assert isinstance(samples, DataFrame), f"samples must be a pandas.DataFrame, found {type(samples)}"
        EXPECTED_COLUMNS =  _EXPECTED_COLS_PERTASK[self.task] 
        assert all(col in samples.columns for col in EXPECTED_COLUMNS), f"samples must have (at least) columns {EXPECTED_COLUMNS}, found {samples.columns}"
        assert samples["image_path"].apply(lambda p: Path(p).exists()).all(), "missing file pointed by samples"
        
        self._samples = samples.sort_values(by="image_path", ignore_index=True)
```

these problems are easier to correct when the show up at setup rather then training phase

---

this should be up there in the module level

```python
_EXPECTED_COlS_CLASSIFICATION = ["image_path", "label", "label_index", "mask_path", "split"]
_EXPECTED_COLS_SEGMENTATION = _EXPECTED_COlS_CLASSIFICATION + ["mask_path"]
_EXPECTED_COLS_PERTASK = {
    "classification": _EXPECTED_COlS_CLASSIFICATION,
    "segmentaion": _EXPECTED_COLS_SEGMENTATION,
}
```

jpcbertoldo (commented):
i think the design is quite nice like this; `AnomalibDataset` and `AnomalibDataModule` became quite simple : )))

the logic of `AnomalibDataModule.is_setup()`, `AnomalibDataModule._setup()`, and `AnomalibDataModule.setup()` are still not taking much advantage of lightning's design -- the `LightningDataModule` is loaded in several process and not all of them need to have all the datasets loaded (e.g. the test one probably only needs the test dataset)

so, just to make sure: is that an intentional behavior? (have the datamodule always load all dataset)

anomalib/data/btech.py
```diff
@@ -186,94 +157,31 @@ def __init__(
...
-            warnings.warn(
-                "seed is None."
-                " When seed is not set, images from the normal directory are split between training and test dir."
-                " This will lead to inconsistency between runs."
-            )
+        super().__init__(task, pre_process)
 
-        self.root = Path(root) if isinstance(root, str) else root
-        self.category: str = category
+        self.root_category = Path(root) / Path(category)
```
jpcbertoldo, line 162:
```suggestion
        self.root_category = Path(root) / category
```

anomalib/data/base/dataset.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class AnomalibDataset(Dataset, ABC):
+    """Anomalib dataset."""
+
+    def __init__(self, task: str, pre_process: PreProcessor):
+        super().__init__()
+        self.task = task
+        self.pre_process = pre_process
+        self._samples = None
```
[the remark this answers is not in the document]
jpcbertoldo: if i understood well, `self._samples` is private (not protected) (i.e. only `AnomalibDataset` should access it, not its children classes), so the "more correct" naming would be `self.__samples` (dunderscore)

another minor but following this convention makes it easier to guess the expected behavior

djdameln:
@jpcbertoldo Thanks for your feedback. I've read your minor comments and will be pushing some updates.

> the logic of AnomalibDataModule.is_setup(), AnomalibDataModule._setup(), and AnomalibDataModule.setup() are still not taking much advantage of lightning's design -- the LightningDataModule is loaded in several process and not all of them need to have all the datasets loaded (e.g. the test one probably only needs the test dataset)

Yes, this is an intentional behaviour and was motivated by the subset splitting responsibility of the `setup` method.

When we perform the dynamic subset splitting in anomaly tasks, we often extract the validation set from the pre-supplied test set (for example, when `val_split_mode` is set to `from_test`). In this case, to create the validation set, we first need to create the test set and then randomly split the test set into validation and testing subsets. If we would use stage-related logic in the `setup` method, we would end up creating and splitting the test set in both the `fit` and `test` stages. This would also mean that we would have pass a seed to the subset splitting, to ensure that we'll end up with the same val/test split during validation and testing.

In supervised tasks, this problem is not present, because the validation set is usually extracted from the training set instead of the test set. This means that the splitting only happens in the `fit` stage, and we can safely use stage-dependent logic in the `setup` method without having to pass a seed.

I feel that having to pass a seed around between different parts of the code to ensure non-overlapping subsets is error-prone and should be avoided. This is why I decided to deviate from the intended use of the `stage` argument in `setup`. I guess the major drawback of this approach is that it leads to higher memory consumption, but since our dataset classes do not keep all of their images in memory but rather read the images from the file system when needed in `__getitem__`, I expect this to be negligible in most cases.

anomalib/data/base/dataset.py
djdameln: I don't think this would be a problem in practice. In our case `Parent` is an abstract class, so we would never encounter an instance of that type.

anomalib/data/base/dataset.py
djdameln: Good catch. This is a leftover from the old design, I didn't make any changes to __getitem__ in the new design.

anomalib/data/base/dataset.py
djdameln: Thanks, great suggestion!

jpcbertoldo:
> When we perform the dynamic subset splitting in anomaly tasks, we often extract the validation set from the pre-supplied test set (for example, when `val_split_mode` is set to `from_test`). In this case, to create the validation set, we first need to create the test set and then randomly split the test set into validation and testing subsets. If we would use stage-related logic in the `setup` method, we would end up creating and splitting the test set in both the `fit` and `test` stages. 

Hmm, ok I see. Makes sense.

> This would also mean that we would have pass a seed to the subset splitting, to ensure that we'll end up with the same val/test split during validation and testing.

Got your point but still would be nice to have a `seed` attrribute (enforced to be `not None` when something random is involved, maybe?) in the datamodule, no?

Right now there is that option using the `random_split` but no seed being kept in the datamodule to keep a trace.
And I think it is ok to stop at the datamodule level, the datasets from the split shouldn't need to keep track of the seed IMO.

Btw, it's be nice to have an option `ValSplitMode.NONE` for the case where the user doesn't care about having a validation split (my case sometimes :upside_down_face: ). (don't forget to adapt the logic in `is_setup()` in that case)

> I feel that having to pass a seed around between different parts of the code to ensure non-overlapping subsets is error-prone and should be avoided. 

Yup.

> I guess the major drawback of this approach is that it leads to higher memory consumption, but since our dataset classes do not keep all of their images in memory but rather read the images from the file system when needed in `__getitem__`, I expect this to be negligible in most cases.

Yup, good point!

---

Are you going to make any other minor changes?

djdameln:
> Got your point but still would be nice to have a seed attrribute (enforced to be not None when something random is involved, maybe?) in the datamodule, no?

Good idea, I agree that this could be useful to have. I've added the seed argument to the datamodule. Please note that we're not using the seed parameter when running the training from the entrypoint scripts, because we already set the global seed using `seed_everything`. So we don't need to pass the seed to ensure consistency between runs.

> Btw, it's be nice to have an option ValSplitMode.NONE for the case where the user doesn't care about having a validation split (my case sometimes 🙃 ). (don't forget to adapt the logic in is_setup() in that case)

Done.

> Are you going to make any other minor changes?

We're planning to merge this to a feature branch for now, as we're working on adding more functionality to the data side of the library (synthetic anomaly generation, support for video datasets), and we don't want to expose this to the main branch until it's stable. So we'll keep making incremental changes to the datamodules on the feature branch. In terms of overall design I don't expect many more changes though, so if you want to start building stuff on top of these base classes that should be fine. You could just target your PR to the [feature branch](https://github.com/openvinotoolkit/anomalib/tree/feature/datamodules).

> I think it'd be nice to explicitly state this in the code and/or doc for the sake of the record and to have future devs not question themselves hahaha.

I've added an explanation to the docstring for now. We'll probably update the documentation on the datamodules at a later point.

anomalib/data/base/datamodule.py
```diff
@@ -55,26 +64,36 @@ def setup(self, stage: Optional[str] = None):
     def _setup(self, _stage: Optional[str] = None) -> None:
         """Set up the datasets and perform dynamic subset splitting.
 
-        May be overridden in subclass for custom splitting behaviour.
+        This method yay be overridden in subclass for custom splitting behaviour.
```
jpcbertoldo, line 67: typo
```suggestion
        This method may be overridden in subclass for custom splitting behaviour.
```

jpcbertoldo (commented):
nice :) 

---

> So **_we don't need to pass the seed to ensure consistency between runs_** (but I still see the added value of the seed parameter for custom use-cases)

Yes, you are right. 
BUT, there ways of ensuring consistency better than others haha. 
Let's leave that for another chat : )

---

Thanks for the changes, great work!

PR #607, opened 2022-10-03
Anomalib CLI Improvements - Update metrics and create post_processing section in the config file

### LINKED ISSUE #384: Anomalib CLI Improvements - Update `metrics` and create `post_processing` section in the `config` file.
The new Anomalib CLI has `threshold` and `normalization_method` parameters in the `metrics` section. However, this may not be intuitive. Instead, it would be better to create a `post_processing` to store these parameters.

In addition, `image_metric_names` and `pixel_metric_names` are a bit verbose, especially for the API version. Instead, those could be renamed as `image_metrics` and `pixel_metrics`, respectively.

Overall, the updated config file would be something like:

```yaml
.
.
.
metrics:
  image_metrics:
    - F1Score
    - AUROC
  pixel_metrics:
    - F1Score
    - AUROC

post_processing:
  adaptive_threshold: true
  default_image_threshold: null
  default_pixel_threshold: null
  normalization_method: min_max

visualization:
  task: segmentation
  log_images_to:
    - local
  inputs_are_normalized: true
.
.
.
```

- [ ] Update the metrics_names parameters
- [ ] Assign default values to metrics_names
- [ ] Move threshold and normalization_method parameters to post_processing section.

samet-akcay:
# Description

- Update metrics and create post_processing section in the config file
- Fixes #384 

## Changes

## Checklist

### DIFF
--- a/anomalib/utils/callbacks/post_processing_configuration.py
+++ b/anomalib/utils/callbacks/post_processing_configuration.py
@@ -1,0 +1,64 @@
+"""Post-Processing Configuration Callback."""
+
+# Copyright (C) 2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+
+import logging
+from typing import Optional
+
+import torch
+from pytorch_lightning import Callback, LightningModule, Trainer
+from pytorch_lightning.utilities.cli import CALLBACK_REGISTRY
+
+from anomalib.models.components.base.anomaly_module import AnomalyModule
+
+logger = logging.getLogger(__name__)
+
+__all__ = ["PostProcessingConfigurationCallback"]
+
+
+@CALLBACK_REGISTRY
+class PostProcessingConfigurationCallback(Callback):
+    """Post-Processing Configuration Callback.
+
+    Args:
+        normalization_method(Optional[str]): Normalization method. <None, min_max, cdf
+        adaptive_threshold (bool): Flag indicating whether threshold should be adaptive.
+        default_image_threshold (Optional[float]): Default image threshold value.
+        default_pixel_threshold (Optional[float]): Default pixel threshold value.
+    """
+
+    def __init__(
+        self,
+        normalization_method: str = "min_max",
+        adaptive_threshold: bool = True,
+        default_image_threshold: Optional[float] = None,
+        default_pixel_threshold: Optional[float] = None,
+    ) -> None:
+        super().__init__()
+        self.normalization_method = normalization_method
+
+        assert (
+            adaptive_threshold or default_image_threshold is not None and default_pixel_threshold is not None
+        ), "Default thresholds must be specified when adaptive threshold is disabled."
+
+        self.adaptive_threshold = adaptive_threshold
+        self.default_image_threshold = default_image_threshold
+        self.default_pixel_threshold = default_pixel_threshold
+
+    # pylint: disable=unused-argument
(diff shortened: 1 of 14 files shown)


anomalib/utils/callbacks/post_processing_configuration.py
```diff
@@ -0,0 +1,64 @@
...
+
+__all__ = ["PostProcessingConfigurationCallback"]
+
+
+@CALLBACK_REGISTRY
+class PostProcessingConfigurationCallback(Callback):
+    """Post-Processing Configuration Callback.
+
+    Args:
+        normalization_method(Optional[str]): Normalization method. <None, min_max, cdf
```
ashwinvaidya17, line 26: Minor issue but there is a `>` missing after cdf

ashwinvaidya17 (approved):
Looks good to me. Maybe later we can look into setting up these values without using a callback.

anomalib/utils/callbacks/post_processing_configuration.py
```diff
@@ -0,0 +1,63 @@
...
+        default_pixel_threshold: Optional[float] = None,
+    ) -> None:
+        super().__init__()
+        self.normalization_method = normalization_method
+
+        assert (
+            adaptive_threshold or default_image_threshold is not None and default_pixel_threshold is not None
+        ), "Default thresholds must be specified when adaptive threshold is disabled."
+
+        self.adaptive_threshold = adaptive_threshold
```
djdameln, line 46: We could consider replacing this parameter with a selectable called `thresholding_method` with the options `fixed` and `adaptive`. This might be a bit less confusing (for the unknowing reader, `adaptive_threshold` could sound like it holds a threshold value). Later we could extend it with `synthetic` option.

djdameln (commented):
> Maybe later we can look into setting up these values without using a callback.

Yeah. I'm a bit concerned about adding too many `ConfigurationCallback`s, because this is not the intended use of PL callbacks. But at this point I also don't have a better design in mind that would work with the CLI. So for now we could keep it this way.

anomalib/utils/callbacks/post_processing_configuration.py
samet-akcay: I was also having a similar idea in my mind. I was not quite sure how exactly to address fixed image and pixel thresholds though

samet-akcay:
> Yeah. I'm a bit concerned about adding too many `ConfigurationCallback`s, because I feel this is not in line with the intended use of PL's `Callback` class. But at this point I also don't have a better design in mind that would work with the CLI. So for now we could keep it this way.

This is the phase-2 of this cli refactoring. It requires some additional extensions to the lightning cli

anomalib/utils/callbacks/post_processing_configuration.py
djdameln: It would be the equivalent of `adaptive_threshold: false` in the current implementation

samet-akcay: yeah, but let's say we set `thresholding_method` to `fixed`. We would still need to set `fixed_image_threshold` and `fixed_pixel_threshold`, right?

djdameln: yeah, but I don't see any problems with that. If users want to use a fixed threshold then they will have to provide a value for that. If they don't know which value to provide then they should just use the adaptive threshold.

PR #629, opened 2022-10-13
add deprecation warning to denormalize class

djdameln:
# Description

Denormalize is no longer used and will be removed in future version. This PR adds the deprecation warning.


samet-akcay (approved):
Thanks!

PR #634, opened 2022-10-17
Synthetic anomaly for testing and validation

djdameln:
# Description

This PR adds synthetic anomaly generation for both testing and validation.

- Adds `SyntheticAnomalyDataset` class, which implements a dataset consisting of 50% normal and 50% synthetic anomalies.
- It uses the augmenter from the DRAEM implementation, which was moved to `data/utils` for this purpose.
- The synthetic validation set can be used by setting `val_split_mode` in the config to `synthetic`.
- The synthetic test set can be used by setting `test_split_mode` in the config to `synthetic`.
- Compatible with all image datasets in the library.

## Known Issues
- Only supports image (not video) datasets
- Documentation needs to be updated (WIP)

## Changes

## Checklist


anomalib/data/utils/augmenter.py
```diff
@@ -38,7 +38,15 @@ class Augmenter:
         noise. If not specified, random noise will be used instead.
     """
 
-    def __init__(self, anomaly_source_path: Optional[str] = None):
+    def __init__(
+        self,
+        anomaly_source_path: Optional[str] = None,
+        p_anomalous: float = 0.5,
+        beta: Union[float, Tuple[float, float]] = (0.2, 1.0),
+    ):
```
jpcbertoldo, line 46: since you're refactoring, could we enjoy the opportunity to add a seed parameter here?
```suggestion
    def __init__(
        self,
        anomaly_source_path: Optional[str] = None,
        p_anomalous: float = 0.5,
        beta: Union[float, Tuple[float, float]] = (0.2, 1.0),
        seed: Optional[int] = None,
    ):
        self.seed = seed or random.randrange(sys.maxsize)
        self.random_generator = random.Random(self.seed)
```

`sys` was not imported!

i found this `random.randrange(sys.maxsize)` [here](https://stackoverflow.com/a/5012617/9582881)

if accepted then all the `random.func_name()` must be replaced by `self.random_generator.func_name()`

---

There is a call to `np.random.choice`, which i think could be replaced by `Random.sample`, giving 

```
# Random.sample() is *without* replacement
aug_ind = np.array(self.random_generator.sample(range(len(self.augmenters)), k=3))
# instead of
# aug_ind = np.random.choice(np.arange(len(self.augmenters)), 3, replace=False)

```

anomalib/data/utils/augmenter.py
```diff
@@ -147,9 +155,14 @@ def augment_batch(self, batch: Tensor) -> Tuple[Tensor, Tensor]:
         masks = torch.stack(masks_list).to(batch.device)
 
         # Apply perturbations batch wise
-        beta = torch.rand(batch_size) * 0.8
-        beta = beta.view(batch_size, 1, 1, 1).expand_as(batch).to(batch.device)
-
-        augmented_batch = batch * (1 - masks) + (1 - beta) * perturbations + beta * batch * (masks)
+        if isinstance(self.beta, float):
```
jpcbertoldo, line 158: a remark: i tend to do all sorts of validations and parameter normalization (in the sense dealing with multiple cases like this one) at the very beginning of `__init__`

it's kind of a matter of tast, but i think it's a sligthtly better practice to make the code structure more predictable

anomalib/data/synthetic.py
```diff
@@ -0,0 +1,133 @@
...
+        self.im_dir = self.root / "images"
+        self.mask_dir = self.root / "ground_truth"
+
+        # clean up any existing data that may be left over from previous run
+        if os.path.exists(self.root):
+            shutil.rmtree(self.root)
+
+        # create directories
+        os.makedirs(self.im_dir)
+        os.makedirs(self.mask_dir)
```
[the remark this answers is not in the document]
jpcbertoldo: `pathlib.Path` has most of these utility functions implemented
i think the sugar syntax it gives makes it easier to read 
there are other stuff like `os.path.isdir()` --> `mypath.is_dir()` that could be changed as well

anomalib/data/utils/augmenter.py
```diff
@@ -38,7 +38,15 @@ class Augmenter:
         noise. If not specified, random noise will be used instead.
     """
 
-    def __init__(self, anomaly_source_path: Optional[str] = None):
+    def __init__(
+        self,
+        anomaly_source_path: Optional[str] = None,
+        p_anomalous: float = 0.5,
+        beta: Union[float, Tuple[float, float]] = (0.2, 1.0),
```
ashwinvaidya17, line 45: We need to add these to the docstrings as well

anomalib/data/utils/augmenter.py
ashwinvaidya17: Not sure if this will work in `__init__`. Say we select a batch size of 8 and it happens that the last batch is of size 7 or 1, then beta will have a wrong shape and the `augment_batch` computation might throw an error. While in most cases these checks and computations are redundant, it makes sense to keep them here for the edge cases.

ashwinvaidya17 (commented):
I have one minor comment.

ashwinvaidya17 (approved):
Thanks for the efforts! Looks good to me.

anomalib/data/base/video.py
```diff
@@ -93,3 +94,25 @@ def __getitem__(self, index: int) -> Dict[str, Union[str, Tensor]]:
             item.pop("mask")
 
         return item
+
+
+class VideoAnomalibDataModule(AnomalibDataModule):
```
samet-akcay, line 99: Which one do you guys think would sound more natural `VideoAnomalibDataModule ` or `AnomalibVideoDataModule`?

anomalib/data/synthetic.py
```diff
@@ -0,0 +1,160 @@
+"""Dataset that generates synthetic anomalies.
+
+This dataset can be used when there is a lack of real anomalous data.
+"""
+
```
samet-akcay, line 5: License is missing

anomalib/data/utils/split.py
```diff
@@ -116,3 +125,14 @@ def random_split(
     # outer list: subsets with the given ratio, inner list: per-label unique
     subsets = list(map(list, zip(*subsets)))
     return [concatenate_datasets(subset) for subset in subsets]
+
+
+def split_normal_and_anomalous(dataset: AnomalibDataset) -> Tuple[AnomalibDataset, AnomalibDataset]:
```
samet-akcay, line 130: a bit verbose, but how about renaming this to `split_to_normal_and_anomalous`. By reading the name, I was not clear what exactly it does.

samet-akcay (changes requested):
Some minor comments.

anomalib/data/base/video.py
djdameln: Yeah I also noticed this earlier. I guess AnomalibVideoDataset would be better. But since it's not related to this PR, let's address it separately.

anomalib/data/utils/split.py
djdameln: How about `split_by_label`?

anomalib/data/base/video.py
djdameln: nvm, it was a small effort, so I added it to this one to speed things up

samet-akcay (approved):
thanks!

PR #636, opened 2022-10-17
Create meta_data.json with ONNX export as well as OpenVINO export 

### LINKED ISSUE #638: onnx Infer problem?
when infer the onnx model,don‘t find meta_data.json file?
if i want infer onnx,must be using openvino libs?
maybe provide some scripts for onnx model infer ,such as using **onnxruntime** infer? or other?

calebmm:
# Description
Create meta_data.json with ONNX export as well as OpenVINO export . This is used to solve OpenVINO version mismatches by exporting the ONNX model in preferred version is required by the users' application.

- Fixes #638 

## Changes

## Checklist

- [NA] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/anomalib/deploy/optimize.py
+++ b/anomalib/deploy/optimize.py
@@ -47,7 +47,7 @@ def export_convert(
     export_mode: str,
     export_path: Optional[Union[str, Path]] = None,
 ):
-    """Export the model to onnx format and convert to OpenVINO IR.
+    """Export the model to onnx format and convert to OpenVINO IR. Metadata.json is generated regardless of export mode.
 
     Args:
         model (AnomalyModule): Model to convert.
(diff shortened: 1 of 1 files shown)


anomalib/deploy/optimize.py
```diff
@@ -65,14 +65,14 @@ def export_convert(
         input_names=["input"],
         output_names=["output"],
     )
+    export_path = os.path.join(str(export_path), "openvino")
```
ashwinvaidya17, line 68: I think we should move this line under `export_model == "openvino"` and move the `with open ...` lines to the top. Otherwise it will create meta_data.json under openvino directory. Since the idea is to have it common for both, we can just create the metadata file in `export_path`.

ashwinvaidya17 (changes requested):
I agree with the motivation. Just one minor comment.

anomalib/deploy/optimize.py
samet-akcay: can't we set `export_path` as 

```
export_path = os.path.join(str(export_path), export_mode)
```

anomalib/deploy/optimize.py
```diff
@@ -65,14 +65,14 @@ def export_convert(
         input_names=["input"],
         output_names=["output"],
     )
+    export_path = os.path.join(str(export_path), export_mode)
```
ashwinvaidya17, line 68: I would prefer if this line is above line 58 otherwise onnx model will live in the root directory (`export_path/model.onnx`) of exported models and metadata json will be under `export_path/onnx/meta_data.json`. But with this if the export_mode is openvino, we will get .bin and .onnx models under `openvino` folder. I am fine with this but I am not sure what you guys prefer.

ashwinvaidya17 (approved):

PR #660, opened 2022-10-31
correct folder mask path

### LINKED ISSUE #659: `anomalib.data.folder.Folder` is not properly adapting the `mask_dir` arg
**Describe the bug**

In `anomalib/data/folder.py`, method `Folder.__init__()`.

There are 4 directory args expected to be relative to the arg `root`. One of them (`mask_dir`) does not get the root properly pre-appended.

image

Because the function `make_dataset()` will deduce the mask file names from the images, the code does not realize that the dir is wrong until one tries to eval the model and the image loading breaks with file not found.

**To Reproduce**

I used `tools/train` with this dataset config

```yaml
dataset:
  name: loreal-rhe
  format: folder
  path: data/cuda/datasets/mydataset
  normal_dir: train/normal/images
  abnormal_dir: test/anomalous/images
  normal_test_dir: test/normal/images
  task: segmentation  # classification or segmentation
  mask: test/anomalous/masks
  extensions: .tif
  split_ratio: null  # not necessary bc the test set already has normals
  image_size: 256
  train_batch_size: 32
  test_batch_size: 32
  num_workers: 8
  transform_config:
    train: null
    val: null
  create_validation_set: false
  tiling:
    apply: false
    tile_size: null
    stride: null
    remove_border_count: 0
    use_random_tiling: False
    random_tile_count: 16
```

The organization of my files are as expected and can be deduced from the config above.
I'm cutting the report short because I already solved it.

jpcbertoldo:
# Description

Fixes #659 

I did not create new tests but I added an assertion in `make_dataset` which should not make the method slower and would force any similar issue to appear much earlier (first time I saw it was after the whole training, which is quite incovenient for a basic bug).

## Changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist


jpcbertoldo:
@samet-akcay this one is quite simple, can i get a review, plz?

samet-akcay (approved):
Thanks for spotting this @jpcbertoldo! Yeah, I agree, it's good to check if the files exist. Annoying to figure it out during training or evaluation.

anomalib/data/folder.py
```diff
@@ -139,6 +139,13 @@ def make_dataset(
             if row.label_index == 1:
                 samples.loc[index, "mask_path"] = str(mask_dir / row.image_path.name)
 
+    # make sure all the files exist, dirs_str is used for error message
+    dirs_str = f"normal_dir={normal_dir}, abnormal_dir {abnormal_dir}, normal_test_dir={normal_test_dir}"
+    assert samples.image_path.apply(lambda x: x.exists()).all(), f"missing image files, {dirs_str}"
```
ashwinvaidya17, line 144: Do we need this check? Isn't `image_path` populated by all the images in the directory. We populate mask_path based on the image in the image directory so the second check would make sense.

ashwinvaidya17 (commented):
Thanks! I have one comment

anomalib/data/folder.py
jpcbertoldo: good point

jpcbertoldo:
done @ashwinvaidya17

ashwinvaidya17 (approved):
Thanks!

djdameln (commented):
I'm fine with the changes, though it might make sense to target these changes to the [datamodules feature branch](https://github.com/openvinotoolkit/anomalib/tree/feature/datamodules) instead of the main branch. That way it would be part of the datamodules redesign and we would avoid conflicts.

jpcbertoldo:
> I'm fine with the changes, though it might make sense to target these changes to the [datamodules feature branch](https://github.com/openvinotoolkit/anomalib/tree/feature/datamodules) instead of the main branch. That way it would be part of the datamodules redesign and we would avoid conflicts.

@djdameln I thought the same and forgot to write it. haha.

Although, I was trying to use the branch and could not really use the CLI (just got tired of debugging and went back to main), so I'm more tending to work on main for the sake of practicity. 

Can we find something smart to do about it? Like, at lest keep a least of PRs to re-do as long as it remains simple (like this one).

djdameln (approved):
Alright, let's merge this to main for now. I've created [an issue](https://github.com/openvinotoolkit/anomalib/issues/680) so that we don't forget to add it to the feature branch later on. Any other changes that we need to repeat on the feature branch we could append them to this checklist [here](https://github.com/openvinotoolkit/anomalib/issues/681).

jpcbertoldo:
> Alright, let's merge this to main for now. I've created [an issue](https://github.com/openvinotoolkit/anomalib/issues/680) so that we don't forget to add it to the feature branch later on. Any other changes that we need to repeat on the feature branch we could append them to this checklist [here](https://github.com/openvinotoolkit/anomalib/issues/681).

NIce! Thanks @djdameln !

jpcbertoldo:
image

Was or has this been planned to be fixed?

samet-akcay:
I think it's alright. We could merge this.

PR #661, opened 2022-10-31
add `unique_dir` option to `config.project` (for script `tools/train.py`)

jpcbertoldo:
# Description

It's a tiny functionality that makes sure that the script `tools/train.py` will always create a new directory to dump the results in.

If `config.project.unique_dir == True`, then `project_path = project_path / datetime.now().strftime("%Y-%m-%d_%H-%M-%S")`.

It should not break the current script because it defaults to `False` even if the option is not present in the `.yaml`, falling back to the previous behavior.

## Changes

## Checklist


samet-akcay (commented):
Would users like this minor feature? If yes, it might be an idea to document this somewhere so it would not be a hidden gem

samet-akcay:
I guess this change would also be consistent with the behaviour in the new CLI.

jpcbertoldo:
> Would users like this minor feature? If yes, it might be an idea to document this somewhere so it would not be a hidden gem

Of course :)

I'm just writting sketchy (i.e. only the necessary code) PRs first to get some approval before writting documentation and tests.

@samet-akcay does it look good to proceed? (in particular, do you think this needs unittest?)

anomalib/config/config.py
```diff
@@ -140,6 +141,19 @@ def get_configurable_parameters(
...
+        config.project.unique_dir = False  # backward compatibility
+
+    if config.project.unique_dir:
+        project_path = project_path / datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    else:
+        warn(
+            "config.project.unique_dir is set to False. "
+            "This does not ensure that your results will be written in an empty directory and you may overwrite files."
+        )
+
```
djdameln, line 156: It would be better to place this after line 158, so that the category folder comes before the unique folder.

djdameln (changes requested):

anomalib/config/config.py
jpcbertoldo: Ah yess. Thanks.

Btw, another idea : would it make sense to name it (in addition to the datetime) a hash of the config file? 

Like this you know that two folders were run with exactly the same options.

djdameln: As long as it doesn't lead to excessively long and unreadable folder names I'm fine with adding it. We could also store a copy of the config in the folder to make it easier to find back a specific run.

jpcbertoldo:
So, a few updates: 

1) I am computing a hash of the config object 

The does not consider all the keys in the config in order to ignore things that only concern the run itself (like directories) and should not influence the outcomes of the optimization ("math-related stuff").

Its digest 8-byte/16-charachter long.

2) the hash is saved in a text file in the root of the project dir

3) category will be taken in consideration as you asked @djdameln 

4) i used the opportunity to save the config file in the project dir as well (it makes debugging easier), keeping the original and normalized version.

ashwinvaidya17:
Not sure if I am in favour of hashing as it adds more complexity to the code. I would suggest users to use comet or wandb as they automatically upload the configuration. This makes comparison and filtering easier.

jpcbertoldo:
It was supposed to be a tiny easy PR to pass quickly 😬. I do use wandb and i see what you mean but i see some added value in adding the hash but i don't have a strong opinion on it so Im good with whatever is decided.

@samet-akcay @djdameln do you have hard opinions on this?

samet-akcay:
@jpcbertoldo, my concern is the verbosity that will be added to the name of the directory, which already increases with the addition of date-time. In addition, this will be obsolete when moved to the new cli and deprecate this one. Perhaps, we could only stick to the addition of date and time.

jpcbertoldo:
Done : )

Where should the docs be updated?

djdameln (approved):
Thanks!

jpcbertoldo:
image

I think the failling check here is not related to my changes.
I only changes `anomalib/config/config.py`, cf screenshot.

samet-akcay:
@jpcbertoldo, this looks like an issue from previous commits. Have you rebased your branch?

jpcbertoldo:
image

Same unrelated problem with coverage report.

I think everything else is fine.

Can we merge?

PR #666, opened 2022-11-02
Fix `black` formatting issues.

samet-akcay:
# Description

- This is to address the black formatting issues.

## Changes

## Checklist

### DIFF
--- a/anomalib/models/patchcore/torch_model.py
+++ b/anomalib/models/patchcore/torch_model.py
@@ -157,8 +157,6 @@ class PatchcoreModel(DynamicBufferModule, nn.Module):
         if n_neighbors == 1:
             # when n_neighbors is 1, speed up computation by using min instead of topk
             patch_scores, locations = distances.min(1)
-            # compute anomaly scoree expects patch_scores and locations to be a 2D tensor
-            patch_scores, locations = patch_scores.unsqueeze(1), locations.unsqueeze(1)
         else:
             patch_scores, locations = distances.topk(k=n_neighbors, largest=False, dim=1)
         return patch_scores, locations
--- a/anomalib/models/patchcore/torch_model.py
+++ b/anomalib/models/patchcore/torch_model.py
@@ -157,6 +157,8 @@ class PatchcoreModel(DynamicBufferModule, nn.Module):
         if n_neighbors == 1:
             # when n_neighbors is 1, speed up computation by using min instead of topk
             patch_scores, locations = distances.min(1)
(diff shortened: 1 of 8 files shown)


djdameln (approved):

anomalib/models/patchcore/torch_model.py
```diff
@@ -157,6 +157,8 @@ def nearest_neighbors(self, embedding: Tensor, n_neighbors: int) -> Tuple[Tensor
         if n_neighbors == 1:
             # when n_neighbors is 1, speed up computation by using min instead of topk
             patch_scores, locations = distances.min(1)
+            # compute anomaly scoree expects patch_scores and locations to be a 2D tensor
+            patch_scores, locations = patch_scores.unsqueeze(1), locations.unsqueeze(1)
```
djdameln, line 161: I don't understand why this is necessary. The tensor gets reshaped in the forward pass [here](https://github.com/openvinotoolkit/anomalib/blob/hotfix/main/anomalib/models/patchcore/torch_model.py#L83), right after they're returned from this method, so what's the point of adding the dimension?

anomalib/models/patchcore/torch_model.py
```diff
@@ -178,15 +180,15 @@ def compute_anomaly_score(self, patch_scores: Tensor, locations: Tensor, embeddi
         # 1. Find the patch with the largest distance to it's nearest neighbor in each image
         max_patches = torch.argmax(patch_scores, dim=1)  # (m^test,* in the paper)
         # 2. Find the distance of the patch to it's nearest neighbor, and the location of the nn in the membank
-        score = patch_scores[torch.arange(len(patch_scores)), max_patches]  # s in the paper
-        nn_index = locations[torch.arange(len(patch_scores)), max_patches]  # m^* in the paper
+        score = patch_scores[torch.arange(patch_scores.shape[0]), max_patches]  # s in the paper
```
djdameln, line 183: for a tensor `x`, `len(x)` should return the size of the first dimension of `x`. So I don't think this is needed either.

djdameln: it may be a bit more robust though, so we could keep it :)

anomalib/models/patchcore/torch_model.py
samet-akcay: After fixing L189, this line is not needed anymore. You're right

anomalib/models/patchcore/torch_model.py
samet-akcay: Yeah, torch suggested this usage instead of `len`, which is the reason for this change.

PR #667, opened 2022-11-02
Fix >100% confidence issue for OpenVINO inference

djdameln:
# Description

- Quick fix for the OpenVINO inference normalization issues.
- The problem was due to not normalizing the openvino outputs when in classification mode.
- Solved by making `anomaly_maps` argument optional in `Inferencer._normalize` and calling `_normalize` with only the pred scores in classification case.

This works as a quick fix, but I feel after this we should revisit the logic in `post_process` for the different inferencers.

- Fixes # (issue)

## Changes

## Checklist


anomalib/deploy/inferencers/openvino_inferencer.py
```diff
@@ -159,11 +159,16 @@ def post_process(
...
+            _, pred_score = self._normalize(pred_scores=pred_score, meta_data=meta_data)
+        elif task == "segmentation":
             if "pixel_threshold" in meta_data:
                 pred_mask = (anomaly_map >= meta_data["pixel_threshold"]).astype(np.uint8)
 
-            anomaly_map, pred_score = self._normalize(anomaly_map, pred_score, meta_data)
+            anomaly_map, pred_score = self._normalize(
+                pred_scores=pred_score, anomaly_maps=anomaly_map, meta_data=meta_data
+            )
+            assert anomaly_map is not None
```
samet-akcay, line 171: Would it be an idea to add `else` here to raise a `ValueError` for the rest of the `task` values?

samet-akcay (changes requested):
Thanks! only a single comment

ashwinvaidya17 (approved):
Thanks for investigating!

samet-akcay (approved):

PR #676, opened 2022-11-03
Video Datamodules

djdameln:
# Description

- Adds video datamodules.
- Introduces `VideoAnomalibDataset` class, which serves as the base class for all video datasets.
- Adds UCSDped and Avenue datasets as first video anomaly datasets.
- The `VideoAnomalibDataset` contains a `samples` dataframe, similar to the `AnomalibDataset`, but it lists the video files instead of the image files. This helps ensure that frames from the same video file always stay together in the same subset after splitting or subsampling.
- The central component of the `VideoAnomalibDataset` is a new entity called the `ClipsIndexer`. This class is based on torchvision's `VideoClips`, and creates a list of consecutive sub-videos (referred to as `clips`). When the indexer object is instantiated, it indexes all clips that can be extracted from the videos in the dataset, so that the `i`th clip in the dataset can be retrieved by calling `clips_indexer.get_item(i)`. When the dataset is a validation or test set, the output of the `ClipsIndexer` also contains a ground truth mask for each frame in the clip.
- The number of frames in each clip and the number of frames between consecutive clips can be configured using the constructor arguments `clip_length_in_frames` and `frames_between_clips`, which are also exposed to the `config.yaml`. In the special case where `clip_length_in_frames` is set to 1, the video dataset behaves like an image dataset, which allows us to use the dataset with any of the current Anomalib models.
- Since different datasets use different file structures for storing the masks and videos, a subclass which inherits from `ClipsIndexer` will need to be implemented for every new dataset that gets added to the library. The concrete subclass should implement at least the `get_masks` method, which defines how a mask is retrieved from the file system given the index of the video and the indices of the frames within the video. If the videos of the dataset are not stored in standard video files (e.g. the videos are folders of images where each image is a single video frame), the subclass should also override `get_clips` and `_compute_frame_pts` (see `UCSDpedClipsIndexer` for an example).

## Known issues
- Visualization only supported for single-frame, and only as separate image files (as opposed to re-composing the video).
- No inference support.

## Example configs
```
dataset:
  name: ucsd
  format: ucsdped
  path: ./datasets/ucsdped
  category: UCSDped2
  task: segmentation
  clip_length_in_frames: 1
  frames_between_clips: 1
  image_size: 256
  train_batch_size: 32
  eval_batch_size: 32
  num_workers: 8
  transform_config:
    train: null
    eval: null
  val_split_mode: same_as_test # options: [same_as_test, from_test]
  tiling:
    apply: false
    tile_size: null
    stride: null
    remove_border_count: 0
    use_random_tiling: False
    random_tile_count: 16
```

```
dataset:
  name: avenue
  format: avenue
  path: ./datasets/avenue
  gt_dir: ./datasets/avenue/ground_truth_demo
  task: segmentation
  clip_length_in_frames: 1
  frames_between_clips: 10
  image_size: 256
  train_batch_size: 32
  eval_batch_size: 32
  num_workers: 8
  transform_config:
    train: null
    eval: null
  val_split_mode: same_as_test # options: [same_as_test, from_test]
  tiling:
    apply: false
    tile_size: null
    stride: null
    remove_border_count: 0
    use_random_tiling: False
    random_tile_count: 16
```
## Changes

## Checklist


anomalib/data/base/video.py
```diff
@@ -0,0 +1,88 @@
...
+
+        item = self.indexer.get_item(index)
+        # include the untransformed image for visualization
+        item["original_image"] = item["image"].to(torch.uint8)
+
+        # apply transforms
+        if "mask" in item.keys() and item["mask"] is not None:
+            processed_frames = [
+                self.pre_process(image=frame.numpy(), mask=m.numpy()) for frame, m in zip(item["image"], item["mask"])
+            ]
```
samet-akcay, line 75: `m` looks a bit ugly compared to the other variables given that they are full names.
```suggestion
            processed_frames = [
                self.pre_process(image=frame.numpy(), mask=mask.numpy()) for frame, mask in zip(item["image"], item["mask"])
            ]
```

anomalib/data/ucsd_ped.py
```diff
@@ -0,0 +1,269 @@
...
+        >>> samples.head()
+           root             folder image_path                    mask_path                         split
+        0  UCSDped/UCSDped2 Test   UCSDped/UCSDped2/Test/Test001 UCSDped/UCSDped2/Test/Test001_gt  test
+        1  UCSDped/UCSDped2 Test   UCSDped/UCSDped2/Test/Test002 UCSDped/UCSDped2/Test/Test002_gt  test
+        ...
+
+    Returns:
+        DataFrame: an output dataframe containing samples for the requested split (ie., train or test)
+    """
+    folders = [filename for filename in sorted(Path(path).glob("*/*")) if os.path.isdir(filename)]
```
samet-akcay, line 65:
```suggestion
    folders = [f for f in sorted(path.glob("**/*")) if f.is_dir()]
```

anomalib/data/ucsd_ped.py
```diff
@@ -0,0 +1,269 @@
...
+            return None
+        frames = self.clips[video_idx][frames_idx]
+
+        mask_frames = sorted(glob.glob(mask_folder + "/*"))
+        mask_paths = [mask_frames[idx] for idx in frames.int()]
+
+        masks = torch.stack([Tensor(cv2.imread(mask_path, flags=0)) / 255.0 for mask_path in mask_paths])
+        return masks
+
+    def _compute_frame_pts(self) -> None:
```
samet-akcay, line 105: What does `pts` mean?

anomalib/data/ucsd_ped.py
```diff
@@ -0,0 +1,269 @@
...
+        """Retrieve the number of frames in each video."""
+        self.video_pts = []
+        for video_path in self.video_paths:
+            n_frames = len(glob.glob(video_path + "/*"))
+            self.video_pts.append(Tensor(range(n_frames)))
+
+        self.video_fps = [None] * len(self.video_paths)  # fps information cannot be inferred from folder structure
+
+    def get_clip(self, idx: int) -> Tuple[torch.Tensor, torch.Tensor, Dict[str, Any], int]:
```
samet-akcay, line 114:
```suggestion
    def get_clip(self, idx: int) -> Tuple[Tensor, Tensor, Dict[str, Any], int]:
```

anomalib/data/ucsd_ped.py
```diff
@@ -0,0 +1,269 @@
...
+            info (Dict)
+            video_idx (int): index of the video in `video_paths`
+        """
+        if idx >= self.num_clips():
+            raise IndexError(f"Index {idx} out of range ({self.num_clips()} number of clips)")
+        video_idx, clip_idx = self.get_clip_location(idx)
+        video_path = self.video_paths[video_idx]
+        clip_pts = self.clips[video_idx][clip_idx]
+
+        frames = sorted(glob.glob(video_path + "/*"))
```
samet-akcay, line 132: pathlib

anomalib/data/ucsd_ped.py
```diff
@@ -0,0 +1,269 @@
...
+    def get_mask(self, idx) -> Optional[Tensor]:
+        """Retrieve the masks from the file system."""
+
+        video_idx, frames_idx = self.get_clip_location(idx)
+        mask_folder = self.mask_paths[video_idx]
+        if mask_folder == "":  # no gt masks available for this clip
+            return None
+        frames = self.clips[video_idx][frames_idx]
+
+        mask_frames = sorted(glob.glob(mask_folder + "/*"))
```
samet-akcay, line 99: pathlib

anomalib/data/ucsd_ped.py
```diff
@@ -0,0 +1,269 @@
...
+        mask_paths = [mask_frames[idx] for idx in frames.int()]
+
+        masks = torch.stack([Tensor(cv2.imread(mask_path, flags=0)) / 255.0 for mask_path in mask_paths])
+        return masks
+
+    def _compute_frame_pts(self) -> None:
+        """Retrieve the number of frames in each video."""
+        self.video_pts = []
+        for video_path in self.video_paths:
+            n_frames = len(glob.glob(video_path + "/*"))
```
samet-akcay, line 109: pathlib

anomalib/data/utils/video.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+        self.mask_paths = mask_paths
+
+    def last_frame_idx(self, video_idx: int) -> int:
+        """Returns the index of the last frame for a given video."""
+        return self.clips[video_idx][-1][-1].item()
+
+    @abstractmethod
+    def get_mask(self, idx: int) -> Optional[Tensor]:
+        """Return the masks for the given index."""
```
samet-akcay, line 43:
```suggestion
        """Return the masks for the given index."""
        raise NotImplementedError
```

anomalib/data/base/video.py
```diff
@@ -0,0 +1,88 @@
...
+    def __getitem__(self, index: int) -> Dict[str, Union[str, Tensor]]:
+        """Return mask, clip and file system information."""
+        assert isinstance(self.indexer, ClipsIndexer)
+
+        item = self.indexer.get_item(index)
+        # include the untransformed image for visualization
+        item["original_image"] = item["image"].to(torch.uint8)
+
+        # apply transforms
+        if "mask" in item.keys() and item["mask"] is not None:
```
samet-akcay, line 72: This is what `refurb` proposes
```suggestion
        if "mask" in item and item["mask"] is not None:
```

anomalib/data/base/video.py
```diff
@@ -0,0 +1,88 @@
...
+            mask = item["mask"]
+            item["mask"] = torch.stack([item["mask"] for item in processed_frames]).squeeze(0)
+            item["label"] = Tensor([1 in frame for frame in mask]).int().squeeze(0)
+        else:
+            item["image"] = torch.stack(
+                [self.pre_process(image=frame.numpy())["image"] for frame in item["image"]]
+            ).squeeze(0)
+
+        if item["mask"] is None:
+            del item["mask"]
```
samet-akcay, line 86: Another proposal by `refurb`
```suggestion
            item.pop("mask")
```

anomalib/data/ucsd_ped.py
```diff
@@ -0,0 +1,269 @@
...
+           root             folder image_path                    mask_path                         split
+        0  UCSDped/UCSDped2 Test   UCSDped/UCSDped2/Test/Test001 UCSDped/UCSDped2/Test/Test001_gt  test
+        1  UCSDped/UCSDped2 Test   UCSDped/UCSDped2/Test/Test002 UCSDped/UCSDped2/Test/Test002_gt  test
+        ...
+
+    Returns:
+        DataFrame: an output dataframe containing samples for the requested split (ie., train or test)
+    """
+    folders = [filename for filename in sorted(Path(path).glob("*/*")) if os.path.isdir(filename)]
+    folders = [folder for folder in folders if len(list(folder.glob("*.tif"))) > 0]
```
samet-akcay, line 66: Another suggestion by `refurb`
```suggestion
    folders = [folder for folder in folders if folder.glob("*.tif")]
```

samet-akcay (changes requested):
Thanks for this tremendous effort!

I've got couple of feedback, most of which is regarding the use of `Path`, `os` and `glob`. The code uses these libraries interchangeably. I think we should use only `Path` since it does what both `os` and `glob`.

tests/pre_merge/datasets/test_datamodule.py
```diff
@@ -165,6 +198,74 @@ def test_non_overlapping_splits(self, folder_data_module):
         ), "Found train and test split contamination"
 
 
+class TestAvenueDataModule:
```
ashwinvaidya17, line 201: Since both these tests have the same code should we pass the fixtures as a list similar to what we do for model tests.?

ashwinvaidya17 (commented):
Thanks. One minor comment

anomalib/data/avenue.py
```diff
@@ -0,0 +1,289 @@
+"""CUHK Avenue Dataset."""
```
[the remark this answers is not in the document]
djdameln: I've added the description and reference, but I could not find any license information for this dataset.

anomalib/data/ucsd_ped.py
djdameln: I prefer to use glob here, because `self.video_paths` is a list of strings, which is inherited behaviour from torchvision `VideoClips`.

anomalib/data/base/dataset.py
```diff
@@ -21,7 +21,7 @@
 from anomalib.data.utils import read_image
 from anomalib.pre_processing import PreProcessor
 
-_EXPECTED_COLS_CLASSIFICATION = ["image_path", "label", "label_index", "split"]
+_EXPECTED_COLS_CLASSIFICATION = ["image_path", "split"]
```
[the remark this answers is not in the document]
djdameln: For video datasets, the samples dataframe contains a list of video files, which can't be assigned a single label. This is because each video file may consist of a combination of normal and anomalous frames.

tests/pre_merge/datasets/test_datamodule.py
ashwinvaidya17: Aren't TestAvenueDataModule and TestUCSDpedDataModule basically the same?

ashwinvaidya17 (commented):
I have the same comment but not sure if github is updating on my browser. I cannot see the new comments you have added on the web ui but I can see them in the notification email.

tests/pre_merge/datasets/test_datamodule.py
djdameln: wow. I thought I removed those but apparently the changes were not pushed to Github.

djdameln: Fixed it now

ashwinvaidya17 (approved):
Thanks!

anomalib/data/avenue.py
```diff
@@ -0,0 +1,289 @@
...
+
+class Avenue(AnomalibDataModule):
+    """Avenue DataModule class.
+
+    Args:
+        root (str): Path to the root of the dataset
+        gt_dir (str): Path to the ground truth files
+        clip_length_in_frames (int, optional): Number of video frames in each clip.
+        frames_between_clips (int, optional): Number of frames between each consecutive video clip.
+        task (str): Task type, either 'classification' or 'segmentation'
```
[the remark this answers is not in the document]
djdameln: Let's address this separately. I've added it to the todo list in https://github.com/openvinotoolkit/anomalib/issues/681

anomalib/data/avenue.py
```diff
@@ -0,0 +1,289 @@
...
+        if matfile == "":  # no gt masks available for this clip
+            return None
+        frames = self.clips[video_idx][frames_idx]
+
+        # read masks from .png files if available, othwerise from mat files.
+        mask_folder = Path(matfile).with_suffix("")
+        if mask_folder.exists():
+            mask_frames = sorted(glob.glob(str(mask_folder) + "/*"))
+            mask_paths = [mask_frames[idx] for idx in frames.int()]
+            masks = torch.stack([Tensor(cv2.imread(mask_path, flags=0)) for mask_path in mask_paths])
```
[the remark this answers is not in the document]
djdameln: This was to keep the masks consistent with the video frames, which are also retrieved as Tensor, as dictated by the torchvision's VideoClips class. But since it leads to 2 unnecessary conversions, it might be better to retrieve the masks as numpy array actually.

anomalib/data/ucsd_ped.py
djdameln: This has to be glob as well, since `video_path` is a string. Otherwise it would become `len(list(Path(video_path).glob("*")))` which I find a bit verbose.

anomalib/data/avenue.py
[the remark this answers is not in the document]
djdameln: done

anomalib/data/base/video.py
```diff
@@ -0,0 +1,88 @@
...
+        super(VideoAnomalibDataset, self.__class__).samples.fset(self, samples)
+        self._setup_clips()
+
+    def _setup_clips(self) -> None:
+        """Compute the video and frame indices of the subvideos.
+
+        Should be called after each change to self._samples
+        """
+        assert callable(self.indexer_cls)
+        self.indexer = self.indexer_cls(  # pylint: disable=not-callable
```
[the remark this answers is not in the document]
djdameln: Not sure, but it seems to be a false positive. I tried several things to get it to stop complaining, but none worked. So to speed thing up I silenced it.

anomalib/data/ucsd_ped.py
djdameln: This method signature is inherited from torchvision's `VideoClips` class. Documentation is a bit poor, but it seems to stand for "presentation times" or "presentation time stamps", which just means the collection of time stamps in number of frames for each video/video_clip.

samet-akcay (approved):

PR #695, opened 2022-11-08
make it possible to use tox pre_merge to test single file

jpcbertoldo:
Can we do this, please?

It allows one to pass command line arguments to tox while keeping the previous behaviour as default.

Source: [https://stackoverflow.com/a/35287375/9582881](https://stackoverflow.com/a/35287375/9582881)

So `tox -e pre_merge` will run all the tests in `tests/pre_merge`, like currently.

And `tox -e pre_merge tests/pre_merge/path/to/single/test_single.py` will only run the tests in `test_single.py`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -26,7 +26,7 @@ deps =
     -r{toxinidir}/requirements/openvino.txt
 commands =
     coverage erase
-    coverage run --include=anomalib/* -m pytest tests/pre_merge/ -ra --showlocals
+    coverage run --include=anomalib/* -m pytest {posargs:tests/pre_merge/} -ra --showlocals
     ; https://github.com/openvinotoolkit/anomalib/issues/94
     coverage report -m --fail-under=85
     coverage xml -o {toxworkdir}/coverage.xml


samet-akcay:
Wasn't aware that this would be possible

jpcbertoldo:
> Wasn't aware that this would be possible

Yes, it's quite handy : )

---

image

I think it mergeable?

ashwinvaidya17:
Nice! Do you guys think we should mention it here as well? https://openvinotoolkit.github.io/anomalib/developer_guide/using_tox.html. It's a handy feature that people might not be aware of.

samet-akcay:
> Nice! Do you guys think we should mention it here as well? https://openvinotoolkit.github.io/anomalib/developer_guide/using_tox.html. It's a handy feature that people might not be aware of.

yeah, that'd be great

jpcbertoldo:
> Nice! Do you guys think we should mention it here as well? https://openvinotoolkit.github.io/anomalib/developer_guide/using_tox.html. It's a handy feature that people might not be aware of.

Sure :) 

Would you mind to take care of that part? I'm focused on something else atm :grimacing:

PR #698, opened 2022-11-08
[DRAFT] CLI  proposal

ashwinvaidya17:
# Description

- CLI proposal with image visualizer as an example
image


ashwinvaidya17:
Since Registries are going to be deprecated (https://github.com/Lightning-AI/lightning/pull/13221) we will have to look for an alternative design.

PR #688, opened 2022-11-05
add option to load metrics with kwargs

jpcbertoldo:
# Description

- Fixes #687 

## Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor (non-breaking change which refactors the code base)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update

## Checklist

- [X] My code follows the [pre-commit style and check guidelines](https://github.com/openvinotoolkit/anomalib/blob/main/CONTRIBUTING.md) of this project.
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with my changes

PR #713, opened 2022-11-15
Migrate old configs to new CLI format

ashwinvaidya17:
# Description

- Migrate old configs to match the new CLI design

# TODO

- [x] Compare performance after migration (ensure that all the optimizers and callbacks are loaded correctly)
- [x] Remove `LightningModels` (they exist for backward compatibility) with the new design we can use the same one as LightningCLI
- [x] Extend to cflow, dfkde, dfm, draem, fastflow, ganomaly, strpm, patchcore, reverse_distillation
- [x] Modify tests


anomalib/utils/cli/helpers/configure_optimizer.py
```diff
@@ -0,0 +1,28 @@
...
+    """Adds optmizer to the Lightning module.
+
+    Args:
+        model (AnomalyModule): Lightning module
+        config (DictConfig): Config.yaml loaded using OmegaConf
+    """
+    if "optimizer" in config:
+        optimizer_module = import_module(".".join(config.optimizer.class_path.split(".")[:-1]))
+        optimizer_class = getattr(optimizer_module, config.optimizer.class_path.split(".")[-1])
+        optimizer = optimizer_class(params=model.parameters(), **config.optimizer.init_args)
```
ashwinvaidya17, line 27: For stfpm this should be `model.student_model.parameters()`

ashwinvaidya17:
Ignore the feature extractor part as those changes will be merged into the feature branch from PR https://github.com/openvinotoolkit/anomalib/pull/748

djdameln (commented):
Sorry for the long list of comments, but I have some concerns about the CLI design. In general, I'm afraid we will lose a lot of flexibility.

- As a result of forcing the `config.yaml` to follow the PL CLI format, we can no longer re-use parameters from one section of the config in another section. This leads to duplicate parameter definitions in the config, where in our current design we only have to define these parameters once. For example, `data` and `visualization` now both have the `task` parameter, where previously we only had `dataset.task`. Another example is the duplication of `input_size` between `data` and `model`. I can think of many ways in which this could lead to bugs and unexpected behaviour.

- Every parameter that is in some way needed by a module must be be defined in its init args, so we are much less free in how we structure our `config.yaml`. For example, let's  say we want to separate the pre-processing related parameters from the dataset-related parameters, that's no longer possible, because both are needed by the DataModule.

- The code that parses the config, especially when it comes to adding the callbacks, is much harder to read than before. (I guess this is mainly because it now has to accommodate both the CLI and the entrypoint scripts).

- Some settings like the early stopping and project path that are now moved under `trainer` may be harder to find for the user.

- I'm still not convinced that the use of `class_path` and `init_args` is an improvement, especially as these keywords now used by some config sections, but not by others, which makes the config file look inconsistent. (This is a minor issue and could be a matter of taste, so feel free to ignore this).

anomalib/models/cflow/config.yaml
```diff
@@ -1,73 +1,77 @@
...
-  show_images: False # show images on the screen
-  save_images: True # save images to the file system
-  log_images: True # log images to the available loggers (if any)
-  image_save_path: null # path to which images will be saved
-  mode: full # options: ["full", "simple"]
-
-project:
-  seed: 0
-  path: ./results
+  task: segmentation
```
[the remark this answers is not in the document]
jpcbertoldo: the trainer (?) should raise an exception : )

jpcbertoldo:
> As a result of forcing the config.yaml to follow the PL CLI format, we can no longer re-use parameters from one section of the config in another section. This leads to duplicate parameter definitions in the config, where in our current design we only have to define these parameters once. For example, data and visualization now both have the task parameter, where previously we only had dataset.task. Another example is the duplication of input_size between data and model. I can think of many ways in which this could lead to bugs and unexpected behaviour.

In this particular case, for example, it can indeed happen that they have different sizes but there is a preprocesing step in the model that does the resizing.

I see your point but it honestly doesn't look so catastrophic IMO.

jpcbertoldo:
> Every parameter that is in some way needed by a module must be be defined in its init args, so we are much less free in how we structure our config.yaml. For example, let's say we want to separate the pre-processing related parameters from the dataset-related parameters, that's no longer possible, because both are needed by the DataModule.

Not trying to disagree on every point (haha), but this actually sounds like an advantage to me. 

By enforcing this structure, it becomes easier to find the correspondance of things between the config file and the code. 
In other words, it's easier to figure out (for exampel) with a Ctrl+F  the parts of the code affected by a parameter (more intuitive to debug).

djdameln:
> In this particular case, for example, it can indeed happen that they have different sizes but there is a preprocesing step in the model that does the resizing.
> 
> I see your point but it honestly doesn't look so catastrophic IMO.

It's not just the handling of the duplicate parameters that I'm concerned about, but also the usability of the config file from the user perspective. Having to change multiple parameters in the config to change a single setting does not feel very user-friendly and could easily lead to confusion. If two different parameters are always forced to have the same value, then from a user perspective it makes no sense to have them as different parameters.


> By enforcing this structure, it becomes easier to find the correspondance of things between the config file and the code.
> In other words, it's easier to figure out (for exampel) with a Ctrl+F the parts of the code affected by a parameter (more intuitive to debug).

I guess that's true, and I agree that in general it would be desirable to have a config file that matches the class structure, but my point is that we lose the flexibility. With the current design we have the freedom to structure the config in any way we see fit, which means that we can structure the parameters based on the design of our modules, but we can deviate from this if needed. We lose this freedom when we switch to the CLI design.

Another side effect is that we cannot pass complex data types to the constructors of our classes. Any complex object that we need in our lightning modules must be instantiated from within the module, which can lead to huge lists of constructor arguments, as we're already seeing in the datamodule.

jpcbertoldo:
> It's not just the handling of the duplicate parameters that I'm concerned about, but also the usability of the config file from the user perspective. Having to change multiple parameters in the config to change a single setting does not feel very user-friendly and could easily lead to confusion. If two different parameters are always forced to have the same value, then from a user perspective it makes no sense to have them as different parameters.

A possible but not necessarily simple solution to have the best of both worlds: a new abstraction layer (yay! haha).

There could be a (1) user-oriented config (solving the issues you're saying) that gets "normalized"/"standardized" (better term?) to a (2) machine-oriented config file (the one passed to the cli). How I imagine it: the first field of the config file states if its (1) or (2). If it's (1), the cli will first generate (2) before running.

To be noted that (1) can have many simplifications (e.g. custom dimension reduction instead of the ones used in the papers) if relevant, while the user can still use (2) for detailed configuration.

> I guess that's true, and I agree that in general it would be desirable to have a config file that matches the class structure, but my point is that we lose the flexibility. With the current design we have the freedom to structure the config in any way we see fit, which means that we can structure the parameters based on the design of our modules, but we can deviate from this if needed. We lose this freedom when we switch to the CLI design.
> 
> Another side effect is that we cannot pass complex data types to the constructors of our classes. Any complex object that we need in our lightning modules must be instantiated from within the module, which can lead to huge lists of constructor arguments, as we're already seeing in the datamodule. (EDIT: I guess it helps that the tiler settings are now moved to their own section, so this might not be that big an issue).

That's true. From my personal experience, I actually think it's nice to put some limitations to developper creativity : ) (more seriously, it makes you think twice before deviating from the general framework).

djdameln:
> A possible but not necessarily simple solution to have the best of both worlds: a new abstraction layer (yay! haha).
>
> There could be a (1) user-oriented config (solving the issues you're saying) that gets "normalized"/"standardized" (better term?) to a (2) machine-oriented config file (the one passed to the cli). How I imagine it: the first field of the config file states if its (1) or (2). If it's (1), the cli will first generate (2) before running.
>
> To be noted that (1) can have simplifications, while the user can still use (2) for detailed configuration.

Hmm yes, that could work, although I would be concerned about the maintenance. I believe @ashwinvaidya17 found a way to link parameters between different sections of a PL CLI, which could potentially address my concern about parameter duplication.

jpcbertoldo:
> Hmm yes, that could work, although I would be concerned about the maintenance

yeah... hehe

anomalib/config/config.py
```diff
@@ -109,21 +109,40 @@ def update_multi_gpu_training_config(config: Union[DictConfig, ListConfig]) -> U
...
+def get_default_root_directory(config: Union[DictConfig, ListConfig]) -> Path:
+    """Sets the default root directory."""
+    # If `resume_from_checkpoint` is not specified, it means that the project has not been created before.
+    # Therefore, we need to create the project directory first.
+    if config.trainer.resume_from_checkpoint is None:
+        root_dir = config.trainer.default_root_dir if config.trainer.default_root_dir else "./results"
+        model_name = config.model.class_path.split(".")[-1].lower()
+        data_name = config.data.class_path.split(".")[-1].lower()
+        category = config.data.init_args.category if "category" in config.data.init_args else ""
+        time_stamp = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
```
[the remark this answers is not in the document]
ashwinvaidya17: Currently the `unique_dir` key is under the `project` key. Where should the `unique_dir` key be placed now?

anomalib/data/__init__.py
```diff
@@ -30,61 +31,24 @@ def get_datamodule(config: Union[DictConfig, ListConfig]) -> LightningDataModule
...
-        )
-    else:
-        raise ValueError(
-            "Unknown dataset! \n"
-            "If you use a custom dataset make sure you initialize it in"
-            "`get_datamodule` in `anomalib.data.__init__.py"
-        )
+    # Since in the new config based on LightningCLI, seed_everything is a separate parameter, we set the one in dataset
+    # to point to it.
+    if config.data.init_args.seed == 0:
```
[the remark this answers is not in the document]
ashwinvaidya17: This check is same as the main. It is just to warn the users that if they set seed to 0, they will see a different behaviour than the previous versions.

anomalib/models/cflow/config.yaml
```diff
@@ -1,73 +1,77 @@
...
+    permute_soft: false
+
+optimizer:
+  class_path: torch.optim.Adam
+  init_args:
+    lr: 0.0001
+
+post_processing:
+  normalization_method: MIN_MAX # <null, min_max, cdf>
+  threshold_method: ADAPTIVE # options: [adaptive, manual]
```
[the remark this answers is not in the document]
ashwinvaidya17: Pytorch lightning automatically parses these as the names of the enums.

anomalib/data/__init__.py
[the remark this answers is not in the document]
djdameln: I don't follow. I thought this was to ensure that the dataset seed is the same as the global seed in the case where the user does not provide a specific seed for the dataset (which is actually not necessary, because just setting a global seed will suffice to ensure consistent splits between runs).

Let's say the user wants random training runs, but consistent subset splits between runs. In this case the user will set `seed_everything` to `null`, and `data.seed` to a fixed integer. If the user picks 0 as dataset seed, the seed will be overwritten by `None` as a result of this check, which silently disables deterministic subset splitting.

Besides, it's not necessary to warn the users here, because the `dataset.seed` parameter was previously not exposed to the users. So I think we can safely remove this check altogether.

anomalib/config/config.py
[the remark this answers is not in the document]
ashwinvaidya17: Alright I have now modified it like this
```yaml
results_dir:
  path: ./results
  unique: false
```

anomalib/models/cflow/config.yaml
```diff
@@ -1,73 +1,77 @@
...
-  permute_soft: false
-  lr: 0.0001
-  early_stopping:
-    patience: 2
-    metric: pixel_AUROC
-    mode: max
-  normalization_method: min_max # options: [null, min_max, cdf]
+  class_path: anomalib.models.Cflow
+  init_args:
+    input_size: [256, 256]
```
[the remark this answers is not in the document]
ashwinvaidya17: Our `update_input_size_config` method overwrites the model input size with image size from data class. And I'll link the arguments form the cli parser.

anomalib/config/config.py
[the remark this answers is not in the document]
jpcbertoldo: > Currently it is optional to create a unique directory with a time stamp. Do you think we could/should keep it that way?

I think he meant that it should become mandatory to be unique. Right?

It was kept as optional (and `False` as default) for back compatibility.

djdameln (commented):
Thanks for the changes.

I still have some comments about the handling of `input_size`. To make sure we're on the same page, let me outline why we added this parameter in the first place.

The reason for defining a `input_size` parameter is that some models (e.g. Padim) need to know the dimensions of the input images when being initialized. This is a bit problematic, because the model does not have access to the dataset parameters, and the dataloaders are not set up when the model is instantiated (which means we can't find the input size by taking a sample from the dataset).

As a workaround, we add a `model.input_size` parameter at runtime when parsing the config. To this parameter we assign the effective input size as seen by the model. Its value is dictated by the `image_size`, but previously also by the `center_crop` and `tile_size` parameters. By adding the input_size automatically, we could prevent the situation where the user has to manually enter the effective image size as a parameter in the config.

Note that the `input_size` parameter only existed in the parsed OmegaConf object, never in the `config.yaml` file.

---

With these changes, we now explicitly list the `input_size` parameter in the config, which could be confusing from a user perspective, and since we overwrite its value anyway, there is no real reason to have it there except make the CLI parser happy. Ideally we would find a workaround that allows us to omit the `model.input_size` parameter from the `config.yaml`, but still pass the effective input size, as dictated by the pre-processing settings, to the model constructor.

jpcbertoldo:
@djdameln  could it be managed in the trainer? 
The trainer has access to both the datamodule/dataset (therefore it can compute the effective image size, seen by the model) and the model.

Idk exatcly how, but it sounds like something that could be manabed with a hook or a callback.

Btw, it would perhaps be nice to have a more precise vocabulary in the args/variables so we know that we're talking about the `nominal_input_size` (from the preprocessin resize) and the `effective_input_size` (seen by the model).

> _these names are just suggestions_

anomalib/config/config.py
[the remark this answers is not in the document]
jpcbertoldo: > Maybe we could set the default value to true because I think that would be desirable for most users.

djdameln:
> @djdameln could it be managed in the trainer?
The trainer has access to both the datamodule/dataset (therefore it can compute the effective image size, seen by the model) and the model.

This could be an option, though I'm not entirely clear on how this would play out with the CLI parser.

We have recently started looking into [custom lightning loops](https://devblog.pytorchlightning.ai/train-anything-with-lightning-custom-loops-4be32314c961), which we are thinking could help to simplify the overall design of the library. I guess it would be possible to manage the effective input size from a custom training loop.

> Btw, it would perhaps be nice to have a more precise vocabulary in the args/variables so we know that we're talking about the nominal_input_size (from the preprocessin resize) and the effective_input_size (seen by the model).

I agree, using this for the internal variable names would make the code more easy to follow. For the user-exposed parameters I'm not so sure. An editable parameter which is named as `effective_input_size` could be even more confusing from a user perspective.

jpcbertoldo:
> For the user-exposed parameters I'm not so sure. An editable parameter which is named as `effective_input_size` could be even more confusing from a user perspective.

In general terms, yes, I agree that the user interface should be kept simple. 
But this particular case sounds reasonable to require some knowledge from the user IMO.  One must be aware of the existence of a preprocessing pipeline that modifies the data.

tests/helpers/model.py
```diff
@@ -79,21 +80,23 @@ def setup_model_train(
         config.init_weights = None
 
     # reassign project path as config is updated in `update_config_for_nncf`
-    config.project.path = project_path
+    config.trainer.default_root_dir = project_path
```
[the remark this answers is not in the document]
ashwinvaidya17: we update default_root_dir path with results_dir.path and use default_root_dir elsewhere. So in this case it is alright.

djdameln:
> Argument linking is only supported from the lightning cli. I haven't added it to this PR as I cli.py is going to be a separate PR

Ah, sorry, I misunderstood your previous comment then. nvm then, let's address it in the next one.

anomalib/deploy/export.py
```diff
@@ -52,21 +52,21 @@ def get_model_metadata(model: AnomalyModule) -> Dict[str, Tensor]:
 def export(
     model: AnomalyModule,
     input_size: Union[List[int], Tuple[int, int]],
-    export_mode: ExportMode,
+    output_format: ExportMode,
```
djdameln, line 55: If we change the variable name we should probably also change the enum to be consistent

djdameln (commented):
Thanks. Just one minor comment left

djdameln (approved):

anomalib/deploy/export.py
```diff
@@ -17,7 +17,7 @@
 from anomalib.models.components import AnomalyModule
 
 
-class ExportMode(str, Enum):
+class OutputFormat(str, Enum):
```
samet-akcay, line 20: Is there any motivation or design discussions for this?

samet-akcay: I have the feeling that we should have `Export` in the naming. How about `ExportFormat`?

anomalib/models/cflow/config.yaml
```diff
@@ -1,74 +1,79 @@
...
+
+post_processing:
+  normalization_method: MIN_MAX # <null, min_max, cdf>
+  threshold_method: ADAPTIVE # options: [adaptive, manual]
+  manual_image_threshold: null
+  manual_pixel_threshold: null
 
 metrics:
-  image:
+  image_metrics:
```
samet-akcay, line 51: Similar to other class_paths, should we define the metrics with full class_paths such as `torchmetrics.F1Score etc?`

anomalib/deploy/export.py
djdameln: I feel `format` is more descriptive than `mode`, because the only thing this parameter changes is the final format of the model files that the exporter produces at the output. I'm fine with `ExportFormat`.

anomalib/utils/cli/cli.py
```diff
@@ -91,9 +84,7 @@ def add_arguments_to_parser(self, parser: LightningArgumentParser) -> None:
             parser (LightningArgumentParser): Lightning Argument Parser.
         """
         # TODO: https://github.com/openvinotoolkit/anomalib/issues/20
-        parser.add_argument(
-            "--export_mode", type=str, default="", help="Select export mode to ONNX or OpenVINO IR format."
-        )
+        parser.add_argument("--format", type=str, default="", help="Select export mode to ONNX or OpenVINO IR format.")
```
samet-akcay, line 87: Same as above, `format` is not self-descriptive to me when reading the code.

tests/pre_merge/deploy/test_inferencer.py
```diff
@@ -87,7 +90,7 @@ def test_torch_inference(self, model_name: str, category: str = "shapes", path:
     @TestDataset(num_train=20, num_test=1, path=get_dataset_path(), use_mvtec=False)
     def test_openvino_inference(self, model_name: str, category: str = "shapes", path: str = "./datasets/MVTec"):
         """Tests OpenVINO inference.
-        Model is not trained as this checks that the inferencers are working.
+        Model is not trained as this checks that the inferencers are working.image
```
samet-akcay, line 93: Is this change intentional? It does not look like a complete sentence?

anomalib/models/cflow/config.yaml
ashwinvaidya17: The problem is that it might become too verbose 
image
But we do support adding metrics like this since PR https://github.com/openvinotoolkit/anomalib/pull/688

samet-akcay: That's fair. I just wanted to know if users were aware of this support.

notebooks/300_benchmarking/302_hpo_wandb.ipynb
```diff
@@ -165,7 +165,7 @@
       "source": [
         "folder_stfpm = \"\"\"dataset:\n",
         "  name: hazelnut\n",
-        "  format: folder\n",
+        "  output_format folder\n",
```
samet-akcay, line 168:
```suggestion
        "  export_format folder\n",
```

notebooks/300_benchmarking/302_hpo_wandb.ipynb
```diff
@@ -233,8 +233,8 @@
         "  logger: [] # options: [comet, tensorboard, wandb, csv] or combinations.\n",
         "  log_graph: false # Logs the model graph to respective logger.\n",
         "\n",
-        "optimization:\n",
-        "  export_mode: null #options: onnx, openvino\n",
+        "export:\n",
+        "  output_format null #options: onnx, openvino\n",
```
samet-akcay, line 237:
```suggestion
        "  export_format null #options: onnx, openvino\n",
```

tests/pre_merge/utils/callbacks/export_callback/dummy_config.yml
```diff
@@ -1,24 +1,27 @@
...
+    threshold:
+      image_default: 0.0
+      pixel_default: 0.0
 
-project:
-  path: ./results
-
-optimization:
-  export_mode: "openvino"
+export: output_format "openvino"
```
samet-akcay, line 22: ?
```suggestion
export: export_format "openvino"
```

samet-akcay (commented):
thanks, correct me if I'm missing something, but I guess `output_format`s left in the code base should be `export_format`?

samet-akcay (approved):
Thanks!

PR #748, opened 2022-11-30
Refactor feature extraction key

ashwinvaidya17:
# Description

- All backbone/feature extraction keys are now in
```yaml
model:
    feture_extractor:
        backbone:
        layers:
````
- `FeatureExtractor` selects timm or torchfx based on the parameters passed to it.
- This PR ended up way bigger than I thought 😃 

## Changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### CODE REVIEW

File: anomalib/models/cflow/lightning_model.py [hunk 1]
```diff
@@ -30,9 +34,7 @@ class Cflow(AnomalyModule):
     def __init__(
         self,
         input_size: Tuple[int, int],
-        backbone: str,
-        layers: List[str],
-        pre_trained: bool = True,
+        feature_extractor: Union[TimmFeatureExtractorParams, TorchFXFeatureExtractorParams],
```
jpcbertoldo, line 37: a minor suggestion (but it helps to keep things consistent as the code scales): there could be something like

```python
FeatureExtractorParamsTypes = Union[TimmFeatureExtractorParams, TorchFXFeatureExtractorParams]
```

to avoid this double-import-and-union

anomalib/models/components/feature_extractors/__init__.py
```diff
@@ -3,14 +3,142 @@
...
+
+import torch
+from omegaconf import DictConfig
+from torch import Tensor, nn
+
+from .timm import TimmFeatureExtractor, TimmFeatureExtractorParams
+from .torchfx import TorchFXFeatureExtractor, TorchFXFeatureExtractorParams
+
+
+class FeatureExtractor(nn.Module):
```
jpcbertoldo, line 16: i find it weird to have behavior/interfaces defined in `__init__.py` 
idk if there is a practical advantage/disavantage but i don't recall to have seen this in other projects

jpcbertoldo (commented):
great effort, thanks!

anomalib/models/cflow/lightning_model.py
ashwinvaidya17: That's a good idea

anomalib/models/components/feature_extractors/feature_extractor.py
```diff
@@ -0,0 +1,159 @@
...
+            Union[TimmFeatureExtractorParams, TorchFXFeatureExtractorParams]: Converted feature extractor parameters.
+        """
+        if isinstance(feature_extractor_params, (DictConfig, dict)):
+            if "layers" in feature_extractor_params:
+                feature_extractor_params = TimmFeatureExtractorParams(**feature_extractor_params)
+            else:
+                feature_extractor_params = TorchFXFeatureExtractorParams(**feature_extractor_params)
+        if not isinstance(feature_extractor_params, (TimmFeatureExtractorParams, TorchFXFeatureExtractorParams)):
+            raise ValueError(f"Unknown feature extractor params type: {type(feature_extractor_params)}")
+        return feature_extractor_params
```
djdameln, line 116: If I understand correctly, the mechanism to distinguish between TorchFX and Timm depends on the occurrence of the `layers` key in the dictionary. I'm not sure if this is desirable. It might lead to confusing situations where the user thinks they're using torchvision backbone but actually it's Timm. 

How about setting the framework explicitly, by passing a `framework` parameter that defaults to `torchvision`?

anomalib/models/components/feature_extractors/feature_extractor.py
```diff
@@ -0,0 +1,159 @@
...
+
+import torch
+from omegaconf import DictConfig
+from torch import Tensor, nn
+
+from .timm import TimmFeatureExtractor, TimmFeatureExtractorParams
+from .torchfx import TorchFXFeatureExtractor, TorchFXFeatureExtractorParams
+
+
+class FeatureExtractor(nn.Module):
```
djdameln, line 16: I have some concerns about the class structure of the feature extractors, mainly caused by the design which involves several layers of wrappers around the feature extractor network.

I feel there are too many objects living in the torch model which are all referred to as 'feature extractor'. First of all, the torch model has a constructor argument named `feature extractor`, which is used to construct the `FeatureExtractor` class, which is then assigned to `self.feature_extractor`. 

```
self.feature_extractor = FeatureExtractor(feature_extractor)
```
This already looks a bit confusing.

Then, due to the wrapper design, this `self.feature_extractor` object now contains several layers of attributes which are all called `feature_extractor`. As a result, the actual network that performs the feature extraction resides at
```
self.feature_extractor.feature_extractor.feature_extractor
```
This makes the code around the feature extraction functionality a bit confusing and hard to navigate.

----
While I understand the motivation behind the design, I wonder if there might be a way to achieve the same thing without relying on this many wrappers. How about using a base `FeatureExtractor` class which has `TorchFXFeatureExtractor` and `TimmFeatureExtractor` as subclasses.

djdameln (commented):
Thanks. I had a close look at the design and I have some concerns, which are outlined below.

anomalib/models/cflow/lightning_model.py
jpcbertoldo: I think it's nice good practice to follow and cheap to implement.
Maybe an exception in mypy checks?

anomalib/models/dfkde/torch_model.py
```diff
@@ -48,7 +49,7 @@ def __init__(
         self.threshold_steepness = threshold_steepness
         self.threshold_offset = threshold_offset
 
-        self.feature_extractor = FeatureExtractor(backbone=backbone, pre_trained=pre_trained, layers=layers).eval()
+        self.feature_extractor = FeatureExtractor(feature_extractor).eval()
```
djdameln, line 52: Would it be an idea to set the feature extractor modules to eval mode by default? I guess eval mode will be desired in most cases.

jpcbertoldo: My personal preferences for this is to not assume much and leave it to the user adding option -- I know, it would be yet a new param...

There are many papers on feature adaptation showing up recently and I think there is no consensus on rather to update the feature extractor or not, so IMO it's hard to assume a better default.

samet-akcay (changes requested):
Similar to what @djdameln raised, my primary concern is the following line in `torch_model.py` and `lightning_model.py`:

```python
feature_extractor: FeatureExtractorParams
```

`feaature_extractor` variable actually contains feature extractor parameters. Storing these as feature_extractor causes a confusion since the codebase has so many feature_extractor here and there.

djdameln (commented):
Thanks, huge improvement now that one layer of wrapper has been removed. I am not 100% convinced that we need the wrapper design at all, since the functionality added by the wrapper classes is only used for the construction of the model. But to speed things up, let's stick with this design for now. We can revisit the design once we deprecate timm.

anomalib/models/components/feature_extractors/wrapper.py
```diff
@@ -0,0 +1,112 @@
...
+        Using Backbone params
+
+        >>> from anomalib.models.components.feature_extractors import TorchFXFeatureExtractorParams
+        >>> from torchvision.models.efficientnet import EfficientNet_B5_Weights
+        >>> params = TorchFXFeatureExtractorParams(backbone="efficientnet_b5",
+        ...                     return_nodes=["features.6.8"],
+        ...                     weights=EfficientNet_B5_Weights.DEFAULT
+        ... )
+        >>> FeatureExtractor(params)
+        FeatureExtractor(
```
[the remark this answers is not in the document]
ashwinvaidya17: I need to update the example from `FeatureExtractor` to `get_feature_extractor` but you can directly pass the params without wrapping them in the params class. See this test

https://github.com/openvinotoolkit/anomalib/blob/7f1a42ccbacba154ba77cb1fb1b589202080c91e/tests/pre_merge/models/test_feature_extractor.py#L44

jpcbertoldo:
Is there a github discussion/issue from which one can read why

```
"TimmFeatureExtractor will be removed in the future version."
```

?

samet-akcay:
> Is there a github discussion/issue from which one can read why
> 
> ```
> "TimmFeatureExtractor will be removed in the future version."
> ```
> 
> ?

@jpcbertoldo, we have recently spotted that timm feature extractor provides inconsistent results, which overall adversely impact the performance of the models. We therefore decided to use torchfx feature extractor

ashwinvaidya17:
> Is there a github discussion/issue from which one can read why
> 
> ```
> "TimmFeatureExtractor will be removed in the future version."
> ```
> 
> ?

I've gathered our motivation here https://github.com/openvinotoolkit/anomalib/discussions/779

anomalib/models/components/feature_extractors/__init__.py
```diff
@@ -3,14 +3,16 @@
 # Copyright (C) 2022 Intel Corporation
```
djdameln, line 3: Should this module be called `feature_extraction`? This would be more in line other modules under `models.components` (e.g. `dimensionality_reduction` and `sampling`)

anomalib/models/components/feature_extractors/utils.py
```diff
@@ -1,27 +1,126 @@
-"""Utility functions to manipulate feature extractors."""
+"""Convenience wrapper for feature extraction methods."""
```
djdameln, line 1:
```suggestion
"""Utility functions for parsing feature extraction parameters."""
```

djdameln (commented):
Thanks. Just two minor comments left:

anomalib/models/stfpm/lightning_model.py
```diff
@@ -27,22 +28,19 @@ class Stfpm(AnomalyModule):
 
     Args:
         input_size (Tuple[int, int]): Size of the model input.
-        backbone (str): Backbone CNN network
-        layers (List[str]): Layers to extract features from the backbone CNN
+        student_teacher_model (FeatureExtractorParams): Teacher model parameters.
```
samet-akcay, line 31: Can you remind me the main motivation of naming this as `student_teacher_model`? Can't this be `teacher_model`?

samet-akcay (commented):
Thanks, only a single comment. Thanks a lot for your effort!

anomalib/models/stfpm/lightning_model.py
ashwinvaidya17: Looks like I forgot to update the description of this argument, but the idea is that since we use the same backbone for both student and teacher model, the argument should indicate the same.

samet-akcay (approved):

djdameln (approved):
Thanks!

PR #774, opened 2022-12-07
Allow setting manual threshold in PL inference

djdameln:
# Description

- This PR makes a few small changes which allows users to set a manual threshold when running lightning inference, even after the model has been trained with adaptive thresholding enabled. To use the manual threshold, the user can simply set the thresholding method to `manual` in the config, and specify the image and pixel-level thresholds. These values will then overwrite the adaptive threshold values which are loaded from the model weights.

- The model loading functionality in LoadModelCallback was moved from `on_test_start` and `on_predict_start`, which removes duplication and is more in line with the intended use of `setup`.
- By adding the model loader callback before the post-processing configuration callbacks, the threshold value is first loaded from the model file. Then, when setup is called in post-processing configuration callback, the thresholds are overwritten by the manual values specified by the user in the `config.yaml`.

Based on my experimentation I could not find any side effects of these changes, but it would be good if the reviewers could double check this on their side.

Related discussions: https://github.com/openvinotoolkit/anomalib/discussions/578, https://github.com/openvinotoolkit/anomalib/discussions/767

## Changes

## Checklist

### DIFF
--- a/docs/source/tutorials/inference.rst
+++ b/docs/source/tutorials/inference.rst
@@ -9,6 +9,10 @@ PyTorch (Lightning) Inference
 =============================
 The entrypoint script in ``tools/inference/lightning.py`` can be used to run inference with a trained PyTorch model. The script runs inference by loading a previously trained model into a PyTorch Lightning trainer and running the ``predict sequence``. The entrypoint script has several command line arguments that can be used to configure inference:
 
+Like the other PyTorch Lightning entrypoints, the inference script reads the model configuration from a config file. It is generally advisible to use the same config file that was used when training the model, although some inference-related paremeter values may be changed to accommodate your needs. Some settings that you might want to change before inference are the batch size, transform config and number of workers for the dataloader.
+
+It is also possible to run inference with a custom anomaly score threshold that differs from the threshold used/computed during training. To change the threshold for inference, simply set the `metrics.threshold.method` to `manual` and specify the custom threshold values in `metrics.threshold.manual_image` and `metrics.threshold.manual_pixel`.
+
 +---------------------+----------+---------------------------------------------------------------------------------+
 |      Parameter      | Required |                                   Description                                   |
 +=====================+==========+=================================================================================+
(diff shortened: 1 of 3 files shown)


ashwinvaidya17 (approved):
Thanks. I feel we should document this behaviour somewhere though.

samet-akcay (approved):
I agree with @ashwinvaidya17. Would be great if we mention this somewhere in the documentation.

djdameln:
> Thanks. I feel we should document this behaviour somewhere though.

> I agree with @ashwinvaidya17. Would be great if we mention this somewhere in the documentation.

Yeah, I think we need an extensive guide in our documentation which explains the thresholding and normalization mechanisms of anomalib, and gives pointers on how to configure this. We're getting many questions in the discussion section on these topics.

ashwinvaidya17 (approved):
Thanks for updating the docs

docs/source/tutorials/inference.rst
```diff
@@ -9,6 +9,10 @@ PyTorch (Lightning) Inference
 =============================
 The entrypoint script in ``tools/inference/lightning.py`` can be used to run inference with a trained PyTorch model. The script runs inference by loading a previously trained model into a PyTorch Lightning trainer and running the ``predict sequence``. The entrypoint script has several command line arguments that can be used to configure inference:
 
+Like the other PyTorch Lightning entrypoints, the inference script reads the model configuration from a config file. It is generally advisible to use the same config file that was used when training the model, although some inference-related paremeter values may be changed to accommodate your needs. Some settings that you might want to change before inference are the batch size, transform config and number of workers for the dataloader.
+
+It is also possible to run inference with a custom anomaly score threshold that differs from the threshold used/computed during training. To change the threshold for inference, simply set the `metrics.threshold.method` to `manual` and specify the custom threshold values in `metrics.threshold.manual_image` and `metrics.threshold.manual_pixel`.
```
samet-akcay, line 14:
```suggestion
It is also possible to run inference with a custom anomaly score threshold that differs from the threshold used/computed during training. To change the threshold for inference, simply set the ``metrics.threshold.method`` to ``manual`` and specify the custom threshold values in ``metrics.threshold.manual_image`` and ``metrics.threshold.manual_pixel``.
```

samet-akcay (changes requested):
thanks, just a single comment to properly render the rst file.

samet-akcay (approved):

PR #811, opened 2022-12-22
🧹 Replace resume_from_checkpoint with ckpt_path

### LINKED ISSUE #349: Model Loader Callback
  - Currently, the Pytorch Lightning(PL) trainer uses a callback `LoadModelCallback` from `anomalib/utils/callbacks/model_loader.py` where it uses Pytorch's `torch.load` function to load the best weights which may cause device-related issues (trained on GPU, testing/predicting on CPU).

  - I suggest removing Pytorch's `torch.load` function from `LoadModelCallback` and letting the PL trainer handle this issue.
 ```python
trainer.test(model=model, datamodule=datamodule, ckpt_path='best')    #or
trainer.test(model=model, datamodule=datamodule, ckpt_path='$path_to_the_checkpoint_user_wish_to_test')
```
 ```python
trainer.predict(model=model, datamodule=datamodule, ckpt_path='best')    #or
trainer.predict(model=model, datamodule=datamodule, ckpt_path='$path_to_the_checkpoint_user_wish_to_predict')
```

Reference: [[PL docs](https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.trainer.trainer.Trainer.html#pytorch_lightning.trainer.trainer.Trainer.test)]

ashwinvaidya17:
# Description

- `resume_from_checkpoint` is going to be deprecated in torch v2.0 and will be replaced by `ckpt_path`.
- This also makes our model configs inline with the new LightningCLI.
- **Note:** This is a breaking change to main 

- Fixes #349 

## Changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


tools/train.py
```diff
@@ -57,14 +57,12 @@ def train():
 
     trainer = Trainer(**config.trainer, logger=experiment_logger, callbacks=callbacks)
     logger.info("Training the model.")
-    trainer.fit(model=model, datamodule=datamodule)
 
-    logger.info("Loading the best model weights.")
-    load_model_callback = LoadModelCallback(weights_path=trainer.checkpoint_callback.best_model_path)
-    trainer.callbacks.insert(0, load_model_callback)
+    ckpt_path = config.trainer.get("ckpt_path", None)
+    trainer.fit(model=model, datamodule=datamodule, ckpt_path=ckpt_path)
```
djdameln, line 62: What's the reason for passing the checkpoint path during training?

anomalib/models/__init__.py
```diff
@@ -95,7 +94,8 @@ def get_model(config: Union[DictConfig, ListConfig]) -> AnomalyModule:
     else:
         raise ValueError(f"Unknown model {config.model.name}!")
 
-    if "init_weights" in config.keys() and config.init_weights:
-        model.load_state_dict(load(os.path.join(config.project.path, config.init_weights))["state_dict"], strict=False)
+    # This is needed for nncf. Might possibly break nncf but are tests don't cover nncf.
+    if "ckpt_path" in config.keys() and config.ckpt_path:
+        model.load_state_dict(load(config.ckpt_path)["state_dict"], strict=False)
```
djdameln, line 99: Is this really needed? This kind of beats the purpose of using PL's checkpoint loading mechanism

djdameln (commented):
Did you confirm that our achieve the same test performance with these changes? Previously we had some accuracy drop when we tried using PL's checkpoint loading mechanism, which is why we created the `LoadModelCallback`. Would be good to confirm that this is no longer an issue.

---

If this is a breaking change to main, should we maybe include it in CLI branch instead?

tools/train.py
ashwinvaidya17: The idea is to allow the users to continue training from a checkpoint.

anomalib/models/__init__.py
ashwinvaidya17: I think this has something to do with nncf but since we don't test that we can remove it. Plus I can see in the tests that the nncf tests set this to None.

ashwinvaidya17:
> Did you confirm that our achieve the same test performance with these changes? Previously we had some accuracy drop when we tried using PL's checkpoint loading mechanism, which is why we created the `LoadModelCallback`. Would be good to confirm that this is no longer an issue.
> 
> If this is a breaking change to main, should we maybe include it in CLI branch instead?

Yeah. I think it will be a good idea to target this to the feature branch. I tested on stfpm, padim, cflow, dfkde, dfm and patchcore. The results are the same.

tools/train.py
djdameln: Did you check if this works?

anomalib/models/dfm/lightning_model.py
```diff
@@ -37,6 +38,7 @@ class Dfm(AnomalyModule):
     def __init__(
         self,
         feature_extractor: FeatureExtractorParams,
+        input_size: Tuple[int, int],
```
djdameln, line 41: Instead of repeating these changes here, would it make sense to rebase the feature branch with main? That would save us some merge conflicts in the future

tools/train.py
ashwinvaidya17: I tried this with stfpm and it resumes training. But I noticed that this should be `config.get` rather than `config.trainer`. I'll push the fix. Thanks

anomalib/models/dfm/lightning_model.py
ashwinvaidya17: I'll do that

djdameln (approved):
Thanks!

PR #822, opened 2022-12-29
[Datamodules] Merge feature branch

djdameln:
# Description

Monster PR to finalize the datamodules refactor. 

Most important changes:

- Switch to new design with base classes for Dataset and DataModule.
- Add base classes for video datasets, and implementations for the UCSDped and Avenue datasets.
- Add Detection task type for data loaders and visualizations.
- Add Synthetic anomaly generation for validation and evaluation purposes.
- Deprecate PreProcessor class and use Albumentations transforms directly.

- Fixes #681 

## Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Refactor (non-breaking change which refactors the code base)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist

- [x] My code follows the [pre-commit style and check guidelines](https://github.com/openvinotoolkit/anomalib/blob/main/CONTRIBUTING.md) of this project.
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with my changes
- [x] I have added a summary of my changes to the [CHANGELOG](https://github.com/openvinotoolkit/anomalib/blob/main/CHANGELOG.md) (not for minor changes, docs and tests).

PR #824, opened 2022-12-30
[Dataset] Add VisA dataset

djdameln:
# Description

- This PR adds the [Visual Anomaly](https://github.com/amazon-science/spot-diff) (VisA) dataset.

- The dataset follows the same format as MVTec, so we could re-use the `make_mvtec_dataset` function.
- The `make_mvtec_dataset` function was slightly modified to make the mask file naming convention a bit more flexible (mvtec uses `"000_mask.png"`, while visa uses `"000.png"`).
- There was a lot of duplication in the download and extract functionality of the different datasets, so this was moved to a shared location.

Currently targeted to feature branch, but will re-target to main once #822 has been merged.

Some examples:

## Known Issues

- ~~CI will probably fail, because the dataset is not yet installed on the CI machine.~~

## Changes

## Checklist


anomalib/data/visa.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+class VisaDataset(AnomalibDataset):
+    """VisA dataset class.
+
+    Args:
+        task (TaskType): Task type, ``classification``, ``detection`` or ``segmentation``
+        transform (A.Compose): Albumentations Compose object describing the transforms that are applied to the inputs.
+        split (Optional[Union[Split, str]]): Split of the dataset, usually Split.TRAIN or Split.TEST
+        root (str): Path to the root of the dataset
```
ashwinvaidya17, line 65: `Union[str, Path]`?

anomalib/data/visa.py
```diff
@@ -0,0 +1,241 @@
...
+        category (str): Category of the MVTec dataset (e.g. "bottle" or "cable").
+        image_size (Optional[Union[int, Tuple[int, int]]], optional): Size of the input image.
+            Defaults to None.
+        center_crop (Optional[Union[int, Tuple[int, int]]], optional): When provided, the images will be center-cropped
+            to the provided dimensions.
+        normalize (bool): When True, the images will be normalized to the ImageNet statistics.
+        train_batch_size (int, optional): Training batch size. Defaults to 32.
+        eval_batch_size (int, optional): Test batch size. Defaults to 32.
+        num_workers (int, optional): Number of workers. Defaults to 8.
+        task TaskType): Task type, 'classification', 'detection' or 'segmentation'
```
ashwinvaidya17, line 100: ~task TaskType)~ task (TaskType)

ashwinvaidya17 (commented):
Thanks. Some very minor comments related to docstrings but the rest looks good. Great efforts!

ashwinvaidya17 (approved):
Thanks

anomalib/data/mvtec.py
```diff
@@ -24,30 +24,39 @@
...
+
+DOWNLOAD_INFO = DownloadInfo(
+    name="mvtec",
+    url="https://www.mydrive.ch/shares/38536/3830184030e49fe74747669442f0f282/download/420938113-1629952094",
+    hash="eefca59f2cede9c3fc5b6befbfec275e",
+)
+
+
+def make_mvtec_dataset(
+    root: Union[str, Path], split: Optional[Union[Split, str]] = None, extensions: Sequence[str] = (".png")
```
samet-akcay, line 58: Rather than providing a new default value for `extensions`, can we use `IMG_EXTENSIONS` directly? Something like

```python
    root: Union[str, Path], split: Optional[Union[Split, str]] = None, extensions: Sequence[str] = IMG_EXTENSIONS
```

anomalib/data/visa.py
```diff
@@ -0,0 +1,245 @@
...
+    ValSplitMode,
+    download_and_extract,
+    get_transforms,
+)
+
+from .mvtec import make_mvtec_dataset
+
+logger = logging.getLogger(__name__)
+
+EXTENSIONS = [".png", ".jpg", ".JPG"]
```
samet-akcay, line 49:
```python
EXTENSIONS = (".png", ".jpg", ".JPG")
```

anomalib/data/visa.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+class VisaDataset(AnomalibDataset):
+    """VisA dataset class.
+
+    Args:
+        task (TaskType): Task type, ``classification``, ``detection`` or ``segmentation``
+        transform (A.Compose): Albumentations Compose object describing the transforms that are applied to the inputs.
+        split (Optional[Union[Split, str]]): Split of the dataset, usually Split.TRAIN or Split.TEST
+        root (Union[str, Path]): Path to the root of the dataset
```
samet-akcay, line 65: Same as my previous comment. Should we consider renaming `root` to `path`? I don't remember the motivation of using `root`

anomalib/data/visa.py
```diff
@@ -0,0 +1,245 @@
...
+        category (str): Sub-category of the dataset, e.g. 'bottle'
```
samet-akcay, line 66: Does this dataset have `bottle` category?

samet-akcay (changes requested):
Thanks great dataset addition, and nice refactor! I only have minor comments.

anomalib/data/__init__.py
```diff
@@ -132,6 +133,24 @@ def get_datamodule(config: Union[DictConfig, ListConfig]) -> AnomalibDataModule:
             val_split_mode=config.dataset.val_split_mode,
             val_split_ratio=config.dataset.val_split_ratio,
         )
+    elif config.dataset.format.lower() == "visa":
+        datamodule = Visa(
+            root=config.dataset.path,
```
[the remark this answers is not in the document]
djdameln: I prefer `root` because this should point to the root directory of the dataset, without category information. When using `path` it would be less clear if the user should provide the path to the root (e.g. `datasets/mvtec/`) or the category (`datasets/mvtec/cable`). We could update the description to `root (Path): Path to the root directory of the dataset`.

If you disagree I am fine with using `path` as well.

anomalib/data/mvtec.py
djdameln: I tried this earlier but some of the code quality checks were complaining

anomalib/data/__init__.py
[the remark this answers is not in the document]
samet-akcay: i'm neutral, but just curriout to know which one would be the best. @ashwinvaidya17, what is your thought here?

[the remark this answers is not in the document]
ashwinvaidya17: I agree that `root` is more clear. And in the new configs we also have `data.init_args.root` that points to the dataset root. Personally I don't mind if we use both `root` and `path` in the repo to make the distinction clear. We can use `path` in the cases where we point to a particular folder that contains images.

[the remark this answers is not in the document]
samet-akcay: Alright, thanks for your inputs!

samet-akcay (approved):
Thanks!

ORippler:
Thanks for integrating the dataset! Do you think you will add a comparison of methods similar as done for MVTec in the [Readme](https://github.com/openvinotoolkit/anomalib#mvtec-ad-dataset)?

PR #845, opened 2023-01-17
Refactor anomalib to new annotation format, add `refurb` and `pyupgrade`

samet-akcay:
# Description

- Refactor anomalib to new annotation format
- Modify the code using `refurb`. The tool has not yet been added to the pre-commit configuration since it requires Python3.10 to execute.
- Add pyupgrade to pre-commit settings and modify the code accordingly.

## Changes

## Checklist


anomalib/data/avenue.py
```diff
@@ -172,23 +173,23 @@ class Avenue(AnomalibVideoDataModule):
         clip_length_in_frames (int, optional): Number of video frames in each clip.
         frames_between_clips (int, optional): Number of frames between each consecutive video clip.
         task TaskType): Task type, 'classification', 'detection' or 'segmentation'
-        image_size (Optional[Union[int, Tuple[int, int]]], optional): Size of the input image.
+        image_size (int | tuple[int, int] | None, optional): Size of the input image.
```
djdameln, line 176: Could you explain why we switch from `Tuple[int, int]` to `tuple[int, int]`?

samet-akcay: With the new annotations, there is no need for `from typing import Tuple`. We could just do `tuple[int, int]`

anomalib/models/cfa/lightning_model.py
```diff
@@ -73,43 +74,40 @@ def on_train_start(self) -> None:
         """Initialize the centroid for the memory bank computation."""
         self.model.initialize_centroid(data_loader=self.trainer.datamodule.train_dataloader())  # type: ignore
 
-    def training_step(self, batch) -> STEP_OUTPUT:
+    def training_step(self, batch: dict[str, str | Tensor], *args, **kwargs) -> STEP_OUTPUT:
```
djdameln, line 77: Is it really needed to add `*args` and `**kwargs` here and in the other lightning models?

anomalib/models/components/base/anomaly_module.py
```diff
@@ -32,14 +35,14 @@ class AnomalyModule(pl.LightningModule, ABC):
...
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__()
         logger.info("Initializing %s model.", self.__class__.__name__)
 
         self.save_hyperparameters()
         self.model: nn.Module
-        self.loss: Tensor
-        self.callbacks: List[Callback]
+        self.loss: nn.Module
```
djdameln, line 44: I don't understand this change. As far as I know, the `loss` attribute holds the value of the loss between epochs, so it should be a Tensor.

anomalib/models/components/base/anomaly_module.py
```diff
@@ -50,78 +53,81 @@ def __init__(self):
...
+            batch (Any): Current batch
             batch_idx (int): Index of current batch
-            _dataloader_idx (int): Index of the current dataloader
+            dataloader_idx (int): Index of the current dataloader
 
         Return:
             Predicted output
         """
-        outputs = self.validation_step(batch, batch_idx)
+        del batch_idx, dataloader_idx  # These variables are not used.
```
djdameln, line 85: can't we solve this with `_batch_idx` and `_dataloader_idx`

anomalib/models/components/feature_extractors/timm.py
```diff
@@ -22,7 +23,7 @@ class TimmFeatureExtractor(nn.Module):
 
     Args:
         backbone (nn.Module): The backbone to which the feature extraction hooks are attached.
-        layers (Iterable[str]): List of layer names of the backbone to which the hooks are attached.
+        layers (Iterable[str]): list of layer names of the backbone to which the hooks are attached.
```
djdameln, line 26: Minor detail, but I think we can keep this capitalized, because it's the first word of the sentence.

setup.py
```diff
@@ -48,27 +50,27 @@ def get_version() -> str:
...
-        List[str]: List of required packages
+        list[str]: List of required packages
     """
 
-    required_packages: List[str] = []
+    required_packages: list[str] = []
 
     for requirement_file in requirement_files:
-        with open(f"requirements/{requirement_file}.txt", "r", encoding="utf8") as file:
+        with open(f"requirements/{requirement_file}.txt", encoding="utf8") as file:
```
djdameln, line 73: Why this change?

djdameln (changes requested):
Thanks. Much more readable now. I have a few comments. My main concern is with the deletion of unused attributes with the `del` keyword, which feels a bit hacky. Maybe we could think of an alternative approach.

anomalib/models/cfa/lightning_model.py
samet-akcay: > Is it really needed to add `*args` and `**kwargs` here and in the other lightning models?

The original `LightningModule` has this signature. Looks like this is the proper way of inheritance to satisfy pylint. Otherwise pylint throws `arguements-differ` error.

anomalib/data/btech.py
```diff
@@ -192,12 +193,12 @@ class BTech(AnomalibDataModule):
...
         transform_config_val: Config for pre-processing during validation.
         create_validation_set: Create a validation subset in addition to the train and test subsets
-        seed (Optional[int], optional): Seed used during random subset splitting.
+        seed (int | None, optional): Seed used during random subset splitting.
         test_split_mode (TestSplitMode): Setting that determines how the testing subset is obtained.
         test_split_ratio (float): Fraction of images from the train set that will be reserved for testing.
         val_split_mode (ValSplitMode): Setting that determines how the validation subset is obtained.
         val_split_ratio (float): Fraction of train or test images that will be reserved for validation.
-        seed (Optional[int], optional): Seed which may be set to a fixed value for reproducibility.
+        seed (int | None, optional): Seed which may be set to a fixed value for reproducibility.
```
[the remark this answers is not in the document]
samet-akcay: > My main concern is with the deletion of unused attributes with the del keyword, which feels a bit hacky. 

This is based on Google's Python style guidelines.
https://google.github.io/styleguide/pyguide.html#214-decision

anomalib/models/components/base/anomaly_module.py
samet-akcay: in almost all of the implementations, we use `self.loss` to compute the loss. For example, here
https://github.com/openvinotoolkit/anomalib/blob/ce70716dbcb8fca2e2b2437eed356c0d0d9f6ca3/anomalib/models/draem/lightning_model.py#L41

When we define this as `Tensor`, pylint and mypy complains, because we use this as `Callable`

samet-akcay: If PL uses this under the hood, we could then rename this to `self.loss_func`

anomalib/models/components/base/anomaly_module.py
samet-akcay: Again this is based on this style guide, but if we agree to use _batch_idx, then I'm fine. Of course we need to ensure if pylint is happy.
https://google.github.io/styleguide/pyguide.html#214-decision

anomalib/models/csflow/lightning_model.py
```diff
@@ -90,21 +92,21 @@ class CsflowLightning(Csflow):
...
             cross_conv_hidden_channels=hparams.model.cross_conv_hidden_channels,
             clamp=hparams.model.clamp,
             num_channels=3,
         )
-        self.hparams: Union[DictConfig, ListConfig]  # type: ignore
+        self.hparams: DictConfig | ListConfig  # type: ignore
         self.save_hyperparameters(hparams)
 
-    def configure_callbacks(self):
+    def configure_callbacks(self) -> list[EarlyStopping]:
```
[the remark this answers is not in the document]
samet-akcay: Yes! Thanks for spotting!

anomalib/models/components/base/anomaly_module.py
samet-akcay: The guideline states the following though
> Other common forms of suppressing this warning include using ‘_’ as the identifier for the unused argument or prefixing the argument name with ‘unused_’, or assigning them to ‘_’. These forms are allowed but no longer encouraged. These break callers that pass arguments by name and do not enforce that the arguments are actually unused.

anomalib/utils/cli/cli.py
```diff
@@ -47,19 +49,19 @@ class AnomalibCLI(LightningCLI):
 
     def __init__(  # pylint: disable=too-many-function-args
         self,
-        model_class: Optional[Union[Type[LightningModule], Callable[..., LightningModule]]] = None,
-        datamodule_class: Optional[Union[Type[LightningDataModule], Callable[..., LightningDataModule]]] = None,
-        save_config_callback: Optional[Type[SaveConfigCallback]] = SaveConfigCallback,
+        model_class: type[LightningModule] | Callable[..., LightningModule] | None = None,
```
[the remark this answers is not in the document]
samet-akcay: That's PL code. I guess that's their preference. Not sure about the actual intention

setup.py
samet-akcay: `refurb` suggests this change.

`[FURB101]: Use `y = Path(x).read_text()` instead of `with open(x, ...) as f: y = f.read()``

anomalib/data/btech.py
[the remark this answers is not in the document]
djdameln: Alright, in that case I'm fine with it. It still feels a bit hacky but maybe I just need to get used to it :)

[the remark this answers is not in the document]
samet-akcay: Yeah, initially looked a bit weird to me too :D

anomalib/utils/cli/cli.py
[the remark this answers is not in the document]
samet-akcay: When I removed `type` I cannot silence some mypy issues. I guess we might need to explicitly specify it here.

setup.py
samet-akcay: This is an automatic change by `pyupgrade`, meaning that we don't need `"r"` in recent versions of python.

anomalib/utils/cli/cli.py
```diff
@@ -147,7 +149,7 @@ def __set_default_root_dir(self) -> None:
         # If `resume_from_checkpoint` is not specified, it means that the project has not been created before.
         # Therefore, we need to create the project directory first.
         if config.trainer.resume_from_checkpoint is None:
-            root_dir = config.trainer.default_root_dir if config.trainer.default_root_dir else "./results"
+            root_dir = config.trainer.default_root_dir or "./results"
```
djdameln, line 152: cool, I was not aware of this syntax

djdameln (approved):
Thanks, just one minor comment, which could probably be ignored

anomalib/utils/cli/cli.py
samet-akcay: me neither :)

anomalib/models/csflow/torch_model.py
```diff
@@ -400,50 +400,59 @@ def __init__(
...
-                Node(
-                    [nodes[-1].out0, nodes[-1].out1, nodes[-1].out2],
-                    ParallelGlowCouplingLayer,
-                    {
-                        "clamp": self.clamp,
-                        "subnet_args": {
-                            "channels_hidden": self.cross_conv_hidden_channels,
-                            "kernel_size": self.kernel_sizes[coupling_block],
-                        },
+            nodes.extend([permute_node])
```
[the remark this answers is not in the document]
samet-akcay: I feel so too, but just did it this way to silence refurb :)

PR #874, opened 2023-01-30
Link dataset path locally in workflow

ashwinvaidya17:
# Description

- Since nbsphinx runs the notebooks, it uses the local dataset path. This is because only the tests get path either locally or from the environment variable. This PR links the dataset path from the environment variable to the local folder.

# Runs

- This works: https://github.com/openvinotoolkit/anomalib/actions/runs/4042706269/jobs/6950734804. The issue was that we have disabled write permissions so this workflow explicitly enables it for this job.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -7,6 +7,9 @@ on:
       - ".github/**" # Ignore changes towards the .github directory
   workflow_dispatch: # run on request (no need for PR)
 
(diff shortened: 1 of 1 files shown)


samet-akcay (approved):
Thanks! and welcome back :D

samet-akcay (approved):

djdameln (approved):

PR #892, opened 2023-02-06
🩹 Patch requirements

ashwinvaidya17:
# Description

- Move extra requirements to extras.txt
- Check in each logger


samet-akcay (approved):
Thanks for your effort

PR #907, opened 2023-02-11
Mvtec 3d

alexriedel1:
# Description
- I have added support for the MVTec3D dataset
- added support for 3D anomaly detection folder datasets

## Checklist


alexriedel1:
Ooh sorry, theres still some trash in it

samet-akcay:
Thanks @alexriedel1! Great PR. We will review it ASAP

anomalib/post_processing/visualizer.py
```diff
@@ -161,6 +161,7 @@ def _visualize_full(self, image_result: ImageResult) -> np.ndarray:
             visualization.add_image(image=image_result.segmentations, title="Segmentation Result")
         elif self.task == TaskType.CLASSIFICATION:
             visualization.add_image(image_result.image, title="Image")
+            visualization.add_image(image_result.heat_map, "Predicted Heat Map")
```
djdameln, line 164: Is this intentional? If so, can you explain this change?

anomalib/data/mvtec_3d.py
```diff
@@ -0,0 +1,293 @@
...
+        super().__init__(task=task, transform=transform)
+
+        self.root_category = Path(root) / Path(category)
+        self.split = split
+
+    def _setup(self) -> None:
+        self.samples = make_mvtec_3d_dataset(self.root_category, split=self.split, extensions=IMG_EXTENSIONS)
+
+
+class MVTec3D(AnomalibDataModule):
```
djdameln, line 207: Can you add an entry for this module in `anomalib.data.get_datamodule` so that the dataset can be used from the `config.yaml`?

djdameln (changes requested):
Thanks @alexriedel1! This is something that has been on our list for ages, but never got around to. Overall the PR looks good, but as you already mentioned, there seem to be some leftovers from development/debugging. If you could clean those up and address these other minor comments we'd be happy to add this to the library.

anomalib/post_processing/visualizer.py
alexriedel1: My intention here is that even in a classification task, the model might predict a heatmap that can be useful to the user.

anomalib/data/utils/image.py
```diff
@@ -206,6 +207,24 @@ def read_image(path: str | Path, image_size: int | tuple[int, int] | None = None
     return image
 
 
+def read_depth_image(path: str | Path) -> np.ndarray:
```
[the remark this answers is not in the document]
alexriedel1: I cannot think about many more depth related utility functions. there will probably function for pre-processing depth images for certain model types but these will be in `anomalib/pre_processing` I guess.

djdameln (commented):
Thanks for updating the PR. Just a few more comments, which I missed in my initial review

anomalib/post_processing/visualizer.py
djdameln: I see. I agree it would be useful to show the heatmap for classification models, but not all classification models produce a heatmap, so this might break. So we need to add a check here to make sure that the heatmap is available:
```suggestion
            if hasattr(image_result, "heat_map"):
                visualization.add_image(image_result.heat_map, "Predicted Heat Map")
```

djdameln (approved):
Looks good to me. Thanks again!

anomalib/data/folder_3d.py
```diff
@@ -0,0 +1,445 @@
...
+        # path is absolute; return unmodified
+        path = folder
+    # path is relative.
+    elif root is None:
+        # no root provided; return absolute path
+        path = folder.resolve()
+    else:
+        # root provided; prepend root and return absolute path
+        path = (Path(root) / folder).resolve()
+    return path
```
samet-akcay, line 93: These functions are also defined in `folder.py`. Would it be an idea to move these to somewhere like `anomalib.data.utils.path`?

anomalib/data/folder_3d.py
```diff
@@ -0,0 +1,445 @@
...
+
+
+def make_folder_dataset(
```
samet-akcay, line 96: How about naming this function something like `make_folder3d_dataset` or similar to avoid any possible confusion? Or is this already clear enough? thoughts? @alexriedel1, @djdameln, @ashwinvaidya17

anomalib/data/folder_3d.py
```diff
@@ -0,0 +1,445 @@
...
+        samples.loc[samples.label == "abnormal", "depth_path"] = samples.loc[
+            samples.label == "abnormal_depth"
+        ].image_path.values
+
+        if normal_test_dir is not None:
+            samples.loc[samples.label == "normal_test", "depth_path"] = samples.loc[
+                samples.label == "normal_test_depth"
+            ].image_path.values
+
+        # make sure all every rgb image has a corresponding depth image and that the file exists
```
samet-akcay, line 193:
```suggestion
        # make sure every rgb image has a corresponding depth image and that the file exists
```

notebooks/100_datamodules/103_folder.ipynb
```diff
@@ -11,7 +11,7 @@
   },
```
samet-akcay, line 11: Are the changes in this notebook intentional? If so, not sure if `C:\\Users\\REH\\anomalib\\datasets\\MVTec\\hazelnut\\test\\crack\\` makes sense?

samet-akcay (changes requested):
@alexriedel1, thank you for your patience and for this great PR! Nice and clean code! 

I only have few comments.

alexriedel1:
now its working :)
```
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...............................................................Passed
check for added large files..............................................Passed
debug statements (python)................................................Passed
detect private key.......................................................Passed
black....................................................................Passed
ruff.....................................................................Passed
mypy.....................................................................Passed
nbqa-black...............................................................Passed
nbqa-ruff................................................................Passed
Lint Dockerfiles.........................................................Passed
```

samet-akcay (approved):

PR #921, opened 2023-02-20
Switch to src layout

samet-akcay:
# Description

- This PR changes the project structure to src layout. The overall tree structure will be as follows:

```bash
├── CHANGELOG.md
├── CITATION.cff
├── CODE_OF_CONDUCT.md
├── CONTRIBUTING.md
├── datasets
├── docs
│   ├── anomalib.png
│   ├── Makefile
│   └── source
├── LICENSE
├── MANIFEST.in
├── notebooks
│   ├── 000_getting_started
│   ├── 100_datamodules
│   ├── 200_models
│   ├── 300_benchmarking
│   ├── 400_openvino
│   └── README.md
├── pyproject.toml
├── README.md
├── requirements
│   ├── base.txt
│   ├── dev.txt
│   ├── docs.txt
│   ├── extras.txt
│   └── openvino.txt
├── setup.py
├── src
│   └── anomalib
├── tests
│   ├── helpers
│   ├── __init__.py
│   ├── nightly
│   └── pre_merge
├── third-party-programs.txt
├── tools
│   ├── benchmarking
│   ├── hpo
│   ├── inference
│   ├── test.py
│   └── train.py
└── tox.ini
```

This official python [link](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/#:~:text=The%20src%20layout%20requires%20installation,installation%20is%20used%20for%20testing) outlines the main advantages of this layout. Some of these are as follows:

> - The src layout requires installation of the project to be able to run its code, and the flat layout does not.
>This means that the src layout involves an additional step in the development workflow of a project (typically, an [editable installation](https://setuptools.pypa.io/en/latest/userguide/development_mode.html) is used for development and a regular installation is used for testing).
>
> - The src layout helps prevent accidental usage of the in-development copy of the code.
This is relevant since the Python interpreter includes the current working directory as the first item on the import path. This means that if an import package exists in the current working directory with the same name as an installed import package, the variant from the current working directory will be used. This can lead to subtle misconfiguration of the project’s packaging tooling, which could result in files not being included in a distribution.
>
> - The src layout helps avoid this by keeping import packages in a directory separate from the root directory of the project, ensuring that the installed copy is used.
>
> - The src layout helps enforce that an [editable installation](https://setuptools.pypa.io/en/latest/userguide/development_mode.html) is only able to import files that were meant to be importable.
This is especially relevant when the editable installation is implemented using a [path configuration file](https://docs.python.org/3/library/site.html#index-2) that adds the directory to the import path.
The flat layout would add the other project files (eg: README.md, tox.ini) and packaging/tooling configuration files (eg: setup.py, noxfile.py) on the import path. This would make certain imports work in editable installations but not regular installations.

## Changes

## Checklist


pyproject.toml
```diff
@@ -73,6 +73,18 @@ follow_imports = "skip"
 follow_imports_for_stubs = true
 
 
+# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+# PYTEST CONFIGURATION                                                        #
```
djdameln, line 77: Are the changes to this file related to switching to src layout or is this something separate?

tox.ini
```diff
@@ -32,7 +32,7 @@ deps =
 commands =
     ; 1. Run Coverage First.
     coverage erase
-    coverage run --include=anomalib/* -m pytest {posargs:tests/pre_merge/} -ra --showlocals
+    coverage run --include=anomalib/* -m pytest {posargs:tests/pre_merge/}
```
ashwinvaidya17, line 35: How does this work now? Should it be `--include=src/anomalib/*`

ashwinvaidya17 (commented):
Thanks. One question

pyproject.toml
samet-akcay: it's separate, but would be good to have. VSCode would automatically detect the tests

tox.ini
samet-akcay: Let me double check

ashwinvaidya17 (approved):
Thanks

PR #870, opened 2023-01-27
Bump up PyTorch Lightning version to v1.9.*

samet-akcay:
# What does this PR do?
- Bump up PyTorch Lightning version to 1.9.0

# Why is this change needed?
- anomalib currently support PL 1.6.* and CLI has new functionalities after v1.7.0. In addition, there are some security related issues, which v1.9.* addresses.

## Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor (non-breaking change which refactors the code base)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist

- [x] My code follows the [pre-commit style and check guidelines](https://github.com/openvinotoolkit/anomalib/blob/main/CONTRIBUTING.md) of this project.
- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing tests pass locally with my changes
- [ ] I have added a summary of my changes to the [CHANGELOG](https://github.com/openvinotoolkit/anomalib/blob/main/CHANGELOG.md) (not for minor changes, docs and tests).

PR #928, opened 2023-02-24
📝Add Dobot Notebooks

ashwinvaidya17:
# Description

- Add notebooks for Dobot


requirements/extras.txt
```diff
@@ -3,3 +3,4 @@ gradio>=2.9.4
 tensorboard
 wandb==0.12.17
 GitPython
```
djdameln, line 5: Would it be an idea to remove this from Anomalibs requirements and add it as a prerequisite in the README of the notebooks instead?

ashwinvaidya17: I think these were moved https://github.com/openvinotoolkit/anomalib/pull/870/files#diff-3d0f7ff059ea9ec68b83f0d1a6eee038151708acf96afe054eb44a9d92786f80

ashwinvaidya17: Oh you mean GitPython. Yeah let's do that

ashwinvaidya17: How about requirements/notebooks.txt?

paularamo (approved):

PR #946, opened 2023-03-03
Custom Loops POC

ashwinvaidya17:
# Description

- First draft for custom loops.
- I have skipped pre-merge checks for now.


anomalib/training/strategies/default/test.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # with a single dataloader don't pass a 2D list | Taken from base method
+        output_or_outputs: EPOCH_OUTPUT | list[EPOCH_OUTPUT] = (
+            outputs[0] if len(outputs) > 0 and self.num_dataloaders == 1 else outputs
+        )
+        self.trainer.post_processor.update_metrics(
+            self.trainer.lightning_module.image_metrics,
+            self.trainer.lightning_module.pixel_metrics,
+            output_or_outputs,
+        )
```
samet-akcay, line 86: I guess this will change in the upcoming PRs, right?

anomalib/training/strategies/default/validate.py
```diff
@@ -0,0 +1,77 @@
...
+from __future__ import annotations
+
+from functools import lru_cache
+from typing import Any, List
+
+from pytorch_lightning.loops.dataloader.evaluation_loop import EvaluationLoop
+from pytorch_lightning.loops.epoch.evaluation_epoch_loop import EvaluationEpochLoop
+from pytorch_lightning.utilities.types import EPOCH_OUTPUT, STEP_OUTPUT
+
+import anomalib.training as core
```
samet-akcay, line 15: Why this? I found this a bit confusing tbh

anomalib/training/strategies/default/validate.py
```diff
@@ -0,0 +1,77 @@
...
+        """The ``on_validation_batch_end`` hook.
+
+        Args:
+            output: The output of the performed step
+            batch: The input batch for the step
+            batch_idx: The index of the current batch
+            dataloader_idx: Index of the dataloader producing the current batch
+        """
+        super()._on_evaluation_batch_end(output, **kwargs)
+        self.trainer.normalizer.update_metrics(output)
```
samet-akcay, line 41: What is the difference between `self.trainer.post_processor.update_metrics` vs `self.trainer.normalizer.update_metrics`?

anomalib/training/utils/normalizer.py
samet-akcay, line 1: Would it be an idea to place this to `anomalib.post_processing`?

anomalib/training/utils/post_processor.py
samet-akcay, line 1: Same as above, would this be part of `anomalib.post_processing`?

samet-akcay: Or are these too `Trainer` specific?

anomalib/utils/cli/cli.py
```diff
@@ -165,25 +162,24 @@ def __set_callbacks(self) -> None:
...
+        # if normalization:
+        #     if normalization == "min_max":
+        #         callbacks.append(MinMaxNormalizationCallback())
+        #     elif normalization == "cdf":
+        #         callbacks.append(CdfNormalizationCallback())
+        #     else:
+        #         raise ValueError(
+        #             f"Unknown normalization type {normalization}.
+        #  \n" "Available types are either None, min_max or cdf"
+        #         )
```
samet-akcay, line 175: We are not going to remove these?

samet-akcay (commented):
I've gone through my first round. Will do another one later.

anomalib/utils/cli/cli.py
ashwinvaidya17: I just commented these for the pre-commit to work. I will remove these in the CLI PR

anomalib/training/strategies/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Custom loops for different training strategies."""
+"""Training strategies for Anomalib models."""
```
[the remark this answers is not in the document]
ashwinvaidya17: Missed this

anomalib/training/strategies/__init__.py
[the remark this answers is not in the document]
ashwinvaidya17: What about we make it more explicit by calling it `learning_strategies` or `training_strategies`?

anomalib/training/strategies/default/test.py
ashwinvaidya17: Yes

anomalib/training/strategies/default/validate.py
ashwinvaidya17: Maybe I should make it more explicit. The `post_processor` updates the threshold metrics whereas the `normalizer` updates the min-max/cdf metrics. How about I rename it to `post_processor.update_threshold_metrics`?

djdameln:
@ashwinvaidya17 I can't run the code. It's complaining about missing keys in the config. Can you push the config file? Also it would be good to pull main to update to the src layout

samet-akcay:
> It's complaining about missing keys in the config. Can you push the config file? Also it would be good to pull main to update to the src layout

Good thing that Github recognises the files being moved to `src`, so luckily no conflicts.

djdameln:
> Good thing that Github recognises the files being moved to src, so luckily no conflicts.

Unfortunately, Github does not move the review comments, so I have to manually move all my pending review comments :(

ashwinvaidya17:
> > Good thing that Github recognises the files being moved to src, so luckily no conflicts.
> 
> Unfortunately, Github does not move the review comments, so I have to manually move all my pending review comments :(

Oh no!

src/anomalib/training/utils/metrics_manager.py
```diff
@@ -0,0 +1,60 @@
...
+            logger.warning(
+                "Cannot perform pixel-level evaluation when task type is classification. "
+                "Ignoring the following pixel-level metrics: %s",
+                self.pixel_metric_names,
+            )
+        else:
+            pixel_metric_names = self.pixel_metric_names
+
+        anomaly_module.image_metrics = create_metric_collection(image_metric_names, "image_")
+        anomaly_module.pixel_metrics = create_metric_collection(pixel_metric_names, "pixel_")
```
djdameln, line 46: Same as for the thresholds, do we need to keep these as attributes of the model, or could we just keep them here in this class? evaluation metrics do not need to persist after re-loading, so I don't see any reason to keep them in the model.

src/anomalib/training/utils/metrics_manager.py
```diff
@@ -0,0 +1,60 @@
...
+    def update_metrics(self, anomaly_module: AnomalyModule, outputs: EPOCH_OUTPUT | List[EPOCH_OUTPUT] | STEP_OUTPUT):
+        if isinstance(outputs, list):
+            for output in outputs:
+                self.update_metrics(anomaly_module, output)
+        else:
+            anomaly_module.image_metrics.cpu()
+            anomaly_module.image_metrics.update(outputs["pred_scores"], outputs["label"].int())
+            if "mask" in outputs.keys() and "anomaly_maps" in outputs.keys():
+                anomaly_module.pixel_metrics.cpu()
+                anomaly_module.pixel_metrics.update(outputs["anomaly_maps"], outputs["mask"].int())
```
djdameln, line 60: Did you try updating the metrics in the `_evaluation_step_end` step, instead of `_evaluation_epoch_end`? This would be more in line with the intended use of torchmetrics objects. Previously we had some issues with normalization which forced us to update the metrics at the end of the epoch, but now that we have more control over the order in which post-processing steps are executed, we could try this again.

djdameln (commented):
Thanks for the huge effort. Overall it's starting to shape up.

One of my main concerns is the coupling between thresholding, normalization and post-processing. If we could separate the responsibilities of the different classes a bit better there it would be a huge improvement.

Another thing to consider is which modules we store in the lightning module. For example, do we want to keep the evaluation metrics as an attribute of the model? Previously we had no choice, because the model itself was responsible for computing the metrics. Now that we have a dedicated class for this we could reconsider. Similar with the threshold and normalization stats, though for these metrics it's a bit more tricky because these need to be preserved when saving and loading the model.

src/anomalib/training/trainer.py
```diff
@@ -0,0 +1,96 @@
...
+        fn = self.state.fn
+
+        self.strategy.barrier("pre_setup")
+
+        if self.datamodule is not None:
+            self._call_lightning_datamodule_hook("setup", stage=fn)
+
+        # Setup required classes before callbacks and lightning module
+        self.post_processor.setup(self.lightning_module)
+        self.normalizer.setup()
```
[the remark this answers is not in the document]
ashwinvaidya17: We don't have access to the `lightning_module` until it is passed to `fot`, `test`, etc. I'll move the ones that do not depend on the `lightning_module` to constructor. There might be a better approach than calling setup. I'll investigate.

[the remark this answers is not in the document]
djdameln: If we keep the metrics as attributes of the `MetricsManager`, and the thresholds as attributes of the `PostProcessor`, would we still need access to the lightning module? If not, this would be another advantage of keeping the metrics in the 'manager' classes instead of the model.

src/anomalib/training/learning_strategies/default/predict.py
```diff
@@ -0,0 +1,94 @@
...
+        Everything is same except for the custom methods called after trainer's predict step.
+
+        Note:
+            Update methods only between the commented block.
+
+        Warning:
+            The order of callbacks and lightning module's ``predict_step`` has been swapped. Originally callbacks were
+            called before the lightning step. But, we need to compute masks and labels before visualizer callback is
+            called. So, in this case the lightning module's ``predict_step`` is called first, then custom methods for
+            post-processing, and finally, the callbacks are called.
```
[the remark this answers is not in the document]
ashwinvaidya17: The problem is that the `lightning_module_hook` gets the anomaly masks that are used by the `post_processor`. So, we cannot call it before the `self.trainer._call_lightning_module_hook`. But since visualizer is the bottleneck here we can consider moving the visualizer callback into custom loops.

src/anomalib/training/learning_strategies/default/test.py
```diff
@@ -0,0 +1,87 @@
...
+        super()._on_evaluation_batch_end(output, **kwargs)
+
+    @lru_cache(1)
+    def _should_track_batch_outputs_for_epoch_end(self) -> bool:
+        """Track batch outputs for epoch end.
+
+        If this is not overridden, the outputs are not collected if the model does not have a ``test_step_end``
+        method. This ensures that the outputs are collected even if the model does not have a ``test_step_end`` method.
+        """
+        return True
```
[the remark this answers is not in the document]
ashwinvaidya17: Now that I think about it, I think we can remove this. Originally this was because the `test_step_end` is not called when the lightning module's `test_step` is not present. But if the `test_step` is not overriden then we don't get any outputs in `test_step_end` so it is pretty useless anyways. I'll remove this.

src/anomalib/training/learning_strategies/default/predict.py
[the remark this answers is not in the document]
djdameln: But the only constraint here is that the post processor and normalizer should be called before the callback hooks right? So the order in my suggestion should address the visualization issue and also keep the ordering of lightning module vs callbacks consistent.

src/anomalib/training/learning_strategies/default/test.py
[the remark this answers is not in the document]
djdameln: I feel it would be better to do it in `test_step_end`. What do you mean with `the outputs are deleted`? It's not possible to test with a PL model without implementing the `test_step_end` method?

src/anomalib/trainer/utils/normalizer.py
```diff
@@ -0,0 +1,137 @@
...
+            elif "box_scores" in outputs:
+                self.normalization_metrics(torch.cat(outputs["box_scores"]))
+            elif "pred_scores" in outputs:
+                self.normalization_metrics(outputs["pred_scores"])
+            else:
+                raise ValueError(
+                    "No values found for normalization, provide anomaly maps, bbox scores, or image scores"
+                )
+        elif self.normalization_method == NormalizationMethod.CDF:
+            self._standardize_batch(outputs)
```
djdameln, line 83: Not sure if this class should hold normalization-method-specific logic.

djdameln (changes requested):
Thanks, getting there. My most important concern is that the thresholding mechanism breaks when normalization is disabled.

src/anomalib/models/components/base/anomaly_module.py
```diff
@@ -87,21 +64,6 @@ def predict_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> A
         del batch_idx, dataloader_idx  # These variables are not used.
 
         outputs: Tensor | dict[str, Any] = self.validation_step(batch)
```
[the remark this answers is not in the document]
ashwinvaidya17: Yeah, that is better. My only concern is that it will end up touching a lot more files. If it does not create extra overhead for you guys then I can make this change.

src/anomalib/trainer/utils/__init__.py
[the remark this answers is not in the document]
ashwinvaidya17: Maybe `managers`?

src/anomalib/models/components/base/anomaly_module.py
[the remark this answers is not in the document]
djdameln: I feel it would be an improvement. Of course, we could also do it in a separate PR. @samet-akcay curious about your opinion on this.

src/anomalib/trainer/utils/metrics_manager.py
```diff
@@ -0,0 +1,154 @@
...
+    def compute(self, outputs: EPOCH_OUTPUT | List[EPOCH_OUTPUT]) -> None:
+        """Computes the metrics.
+
+        1. Update metrics
+        2. Compute metrics
+
+        Args:
+            outputs (EPOCH_OUTPUT): Epoch outputs
+        """
+        self._update_metrics(outputs)
```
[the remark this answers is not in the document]
ashwinvaidya17: The threshold is computed at the end of the epoch. The metrics then use this threshold when compute is called. When I call `update` at the end of the batch and call `compute` at the end of the epoch, I get poor F1 scores.

src/anomalib/trainer/utils/normalizer.py
ashwinvaidya17: Should I add methods such as `normalize_batch` inside `post_processing/normalization/cdf.py`. I can then directly call those methods here. Not sure if that's the right place for this.

src/anomalib/trainer/utils/metrics_manager.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def set_threshold(self) -> None:
+        """Sets threshold."""
+
+        if (
+            self.trainer.state.stage in (TrainerFn.TESTING, TrainerFn.PREDICTING)
+            and self.trainer.normalizer.normalization_method != NormalizationMethod.NONE
+        ):
+            image_metrics_threshold = 0.5
+            pixel_metrics_threshold = 0.5
```
djdameln, line 89: Could you add a comment here, explaining why we set the thresholds to 0.5?

src/anomalib/trainer/utils/metrics_manager.py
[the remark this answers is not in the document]
djdameln: Hmm, I see. It's because the F1Score updates the tp, fp, tn, fn counts during the update call, and for this it uses the threshold. So it seems there is no way around updating the metrics at the end of the epoch, at least for the validation sequence. For the test sequence we could still do it at the end of each batch. I feel it would be a bit better in line with the intended use of the torchmetrics classes. The drawback would be that we'll have slightly different implementation betwen val and test.

src/anomalib/trainer/utils/normalizer.py
djdameln: Not sure what would be the best solution here. It's just that I don't like seeing normalization method specific logic in the normalizer. Because this would mean that adding a new normalization type could involve changing the normalizer. Ideally this should not be necessary. We could consider addressing this separately to speed things up.

src/anomalib/trainer/utils/__init__.py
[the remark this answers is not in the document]
djdameln: @samet-akcay any ideas about the naming here?

samet-akcay (commented):
Getting there...

I think we could consider merging this one. It is not possible to fix all the changes in a single pr. there are already follow-up PRs that are on the queue.

src/anomalib/trainer/loops/predict.py
```diff
@@ -0,0 +1,90 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any
+
+import torch
+from lightning_fabric.utilities import move_data_to_device
+from pytorch_lightning.loops.dataloader.prediction_loop import PredictionLoop
+from pytorch_lightning.loops.epoch.prediction_epoch_loop import PredictionEpochLoop
+
+import anomalib.trainer as core
```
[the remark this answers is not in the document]
ashwinvaidya17: Since I need to refer to the trainer here and predict is called inside trainer, it leads to circular import. I took this inspiration from pytorch lightning. They use `import pytorch_lightning as pl` which does the same thing and circumvents the circular import issue. But maybe we can name it something different.

PR #954, opened 2023-03-09
Enable bandit scanning

yunchu:
# Description

- Enabled bandit code scanning by adding jobs in code_scan workflow
- Configuration file for the bandit was imported/added from IPAS as a default
- Corresponding tox testenv also added
- The report will be posted as the artifact to the workflow

## Changes

## Checklist


.github/workflows/code_scan.yml
```diff
@@ -7,6 +7,26 @@ on:
     - cron: "0 18 * * 1-5"
 
 jobs:
+  Bandit:
+    runs-on: ubuntu-20.04
```
ashwinvaidya17, line 11: Do we need this? The CI already runs on ubuntu 20 image. If we add this it might unnecessarily pull the image and given the proxy settings, pulling images generally fails stating "too many requests"

ipas_default.config
```diff
@@ -0,0 +1,410 @@
+
```
ashwinvaidya17, line 1: Is it possible to move this to some other folder instead of root?

tox.ini
```diff
@@ -97,3 +97,12 @@ commands =
...
     rm snyk-req.txt
+
+[testenv:bandit-scan]
+skip_install = true
+deps =
+    bandit
+allowlist_externals =
+    bandit
+commands =
+    - bandit -r -c {toxinidir}/ipas_default.config {toxinidir}/ -f txt -o {toxworkdir}/bandit-report.txt
```
ashwinvaidya17, line 108: Maybe we can place the config in `.ci`. We can then use `{toxinidir}/.ci/ipas_default.config`

ashwinvaidya17 (commented):
Thanks. I have a few comments

ipas_default.config
samet-akcay: Rather than having this config in the root or somewhere else, would it be possible to move this `pyproject.toml` file?

We ideally want to collect all the configuration within `pyproject.toml` file. I can see from the bandit [documentation](https://bandit.readthedocs.io/en/latest/config.html), it seems to be possible: 

```
[tool.bandit]
exclude_dirs = ["tests", "path/to/file"]
tests = ["B201", "B301"]
skips = ["B101", "B601"]

[tool.bandit.any_other_function_with_shell_equals_true]
no_shell = [
  "os.execl",
  ...
```

And to run bandir based on the configurations from `pyproject.toml` file

```
bandit -c pyproject.toml -r .
```

.github/workflows/code_scan.yml
yunchu: it's about using the Github-hosted runner instance instead of Self-hosted one. so we don't need to worry about the image pulling burden at all.
in case of Snyk, scanning request should be passed to the internal resource but it's not possible to access it from external (GH-hosted one) without having something like VPN. so that's the reason two jobs run on different machine at this point.

ipas_default.config
yunchu: yes, it can. can you point out the proper target?

tox.ini
yunchu: that's clear. will update.

samet-akcay: Same as the above comment. I think we should have something like 

```
bandit -r -c {toxinidir}/pyproject.toml ...
```

ipas_default.config
yunchu: @samet-akcay Yes, I know that but my concern is on that I've copied the bandit config file from IPAS (Intel Product Assurance and Security) which contains minimum requirements for specific task in the SDL (internal release process), so I think, it would be better to keep it as same as original to track the updating their side and applying them to our projects efficiently. how do you think?

yunchu: here is the internal repo which has configuration: https://github.com/intel-innersource/applications.security.bandit-config

ashwinvaidya17 (approved):
Thanks

PR #969, opened 2023-03-18
Fixed outdated info in readme

blaz-r:
# Description

I fixed links in the readme that were not working. 
* Most of them were broken due to move of `anomalib` into `src` directory.
* There was also a link to hardware section of the documentation that no longer exists so I changed that as well.
* Since some of the links were relative and some used entire url, I also unified them all to be relative if referring to the file from the repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,11 +44,11 @@ Anomalib is a deep learning library that aims to collect state-of-the-art anomal
 
 # Getting Started
 
-To get an overview of all the devices where `anomalib` as been tested thoroughly, look at the [Supported Hardware](https://openvinotoolkit.github.io/anomalib/#supported-hardware) section in the documentation.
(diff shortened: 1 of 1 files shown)


blaz-r:
Also fixed commands:
* added `src` before anomalib in path
* `run` in weight path
* --meta_data changed to --metadata

samet-akcay (approved):
Thanks for the fix @blaz-r

blaz-r:
@samet-akcay one thing I noticed is that meta_data was changed to metadata, so I fixed that as I said before. But in version 4.0.0 there still is meta_data used. 
Is my change correct to reflect the current state on main branch or should it be as it is v4.0.0?

samet-akcay:
it's correct. Those changes will be in effect in v0.5.0

PR #983, opened 2023-03-26
WIP: Add kolektor dataset

### LINKED ISSUE #964: [Task]: Implement Kolektor dataset

Add Kolektor dataset: https://www.vicos.si/resources/kolektorsdd/

Implement dataloader to support Kolektor dataset.

### Additional context

_No response_

Ravindu987:
# Description

- Work in progress
- Added the dataset classes, dataloaders and other necessary functions to load `Kolektor Dataset`
- Due to the nature of the dataset, used unconventional methods which needs to be discussed
- Requires ImageIO and Numpy libraries

- Fixes #964 

## Changes

## Checklist

### DIFF
--- a/src/anomalib/data/kolektor.py
+++ b/src/anomalib/data/kolektor.py
@@ -19,6 +19,7 @@ from anomalib.data.utils import (
 )
 
 
+# Check if a mask shows defects
 def is_good(path):
     img_arr = imread(path)
     if np.all(img_arr == 0):
@@ -33,10 +34,11 @@ def make_kolektor_dataset(
 ) -> DataFrame:
 
     root = Path(root)
+
+    # Get list of images and masks
     samples_list = [
         (str(root),) + f.parts[-2:] for f in root.glob(r"**/*") if f.suffix == ".jpg"
     ]
-
     masks_list = [
         (str(root),) + f.parts[-2:] for f in root.glob(r"**/*") if f.suffix == ".bmp"
     ]
@@ -44,36 +46,42 @@ def make_kolektor_dataset(
     if not samples_list:
         raise RuntimeError(f"Found 0 images in {root}")
 
+    # Create dataframes
     samples = DataFrame(samples_list, columns=["path", "item", "image_path"])
     masks = DataFrame(masks_list, columns=["path", "item", "image_path"])
 
+    # Modify image_path column by converting to absolute path
     samples["image_path"] = samples.path + "/" + samples.item + "/" + samples.image_path
     masks["image_path"] = masks.path + "/" + masks.item + "/" + masks.image_path
 
+    # Sort samples by image path
     samples = samples.sort_values(by="image_path", ignore_index=True)
     masks = masks.sort_values(by="image_path", ignore_index=True)
 
+    # Add mask paths for sample images
     samples["mask_path"] = masks.image_path.values
 
+    # Use is_good func to configure the label
     samples["label"] = samples["mask_path"].apply(is_good)
-    samples.loc[(samples.label), "label"] = "Good"
+    samples.loc[(samples.label == True), "label"] = "Good"
     samples.loc[(samples.label == False), "label"] = "Bad"
 
+    # Add label indexes
     samples.loc[(samples.label == "Good"), "label_index"] = 0
     samples.loc[(samples.label == "Bad"), "label_index"] = 1
     samples.label_index = samples.label_index.astype(int)
 
-    # samples = samples.sort_values(by=["label", "image_path"], ignore_index=True)
-
+    # Add all 'Bad' samples to test set
     samples.loc[(samples.label == "Bad"), "split"] = "test"
 
+    # Divide 'good' images to train/test on 0.8/0.2 ratio
     train_samples, test_samples = train_test_split(
         samples[samples.label == "Good"], train_size=0.8, test_size=0.2, random_state=42
     )
-
     samples.loc[train_samples.index, "split"] = "train"
     samples.loc[test_samples.index, "split"] = "test"
 
+    # Reorder columns
     samples = samples[
         ["path", "item", "split", "label", "image_path", "mask_path", "label_index"]
     ]
@@ -87,6 +95,7 @@ def make_kolektor_dataset(
               folder follow the same naming convention as the anomalous images in the dataset (e.g. image: '000.png', \
               mask: '000.png' or '000_mask.png')."
 
+    # Get the dataframe for the required split
     if split:
         samples = samples[samples.split == split].reset_index(drop=True)
 
--- a/src/anomalib/data/kolektor.py
+++ b/src/anomalib/data/kolektor.py
@@ -1,17 +1,25 @@
 from __future__ import annotations
 
 from pathlib import Path
+import albumentations as A
 
 from pandas import DataFrame
 from imageio import imread
 from sklearn.model_selection import train_test_split
 
-from anomalib.data.task_type import TaskType
-from anomalib.data.utils import Split
 import numpy as np
 
+from anomalib.data.task_type import TaskType
+from anomalib.data.utils import (
+    InputNormalizationMethod,
+    Split,
+    TestSplitMode,
+    ValSplitMode,
+    get_transforms,
+)
 
-def isGood(path):
+
+def is_good(path):
     img_arr = imread(path)
     if np.all(img_arr == 0):
         return True
@@ -47,7 +55,7 @@ def make_kolektor_dataset(
 
     samples["mask_path"] = masks.image_path.values
 
-    samples["label"] = samples["mask_path"].apply(isGood)
+    samples["label"] = samples["mask_path"].apply(is_good)
     samples.loc[(samples.label), "label"] = "Good"
     samples.loc[(samples.label == False), "label"] = "Bad"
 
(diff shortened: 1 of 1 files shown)


src/anomalib/data/kolektor.py
```diff
@@ -0,0 +1,179 @@
+from __future__ import annotations
```
ashwinvaidya17, line 1: The file should include the source to the original dataset and a license header. You can have a look here https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/data/mvtec.py

src/anomalib/data/kolektor.py
```diff
@@ -0,0 +1,179 @@
...
+            transform=transform_train,
+            split=Split.TRAIN,
+            root=root,
+        )
+        self.test_data = KolektorDataset(
+            task=task,
+            transform=transform_eval,
+            split=Split.TEST,
+            root=root,
+        )
```
ashwinvaidya17, line 179: This is missing `prepare_data` method. Have a look at https://github.com/openvinotoolkit/anomalib/blob/27adbaeb7da406c52bceba06dbc875ba3c0e955a/src/anomalib/data/mvtec.py#L258 to see an example implementation

src/anomalib/data/kolektor.py
```diff
@@ -0,0 +1,179 @@
...
+    ]
+
+    # assert that the right mask files are associated with the right test images
+    assert (
+        samples.loc[samples.label_index == 1]
+        .apply(lambda x: Path(x.image_path).stem in Path(x.mask_path).stem, axis=1)
+        .all()
+    ), "Mismatch between anomalous images and ground truth masks. Make sure the mask files in 'ground_truth' \
+              folder follow the same naming convention as the anomalous images in the dataset (e.g. image: '000.png', \
+              mask: '000.png' or '000_mask.png')."
```
ashwinvaidya17, line 96: Are you sure this message is correct?

ashwinvaidya17 (changes requested):
Thanks for the efforts! I understand this is WIP, I've still gone ahead and added a few comments. Also, can you also accress the issues raised by Codacy.

Ravindu987:
Hi @ashwinvaidya17 
Thank you very much for your feedback.
In the latest commits I resolved three of the issues you raised.

Regarding the `prepare_dataset` there's an issue. The download link for the dataset doesn't provide an extension for the downloaded file. Instead it sends the extension via the response header. This works well with a browser but the `urlretrieve` method used in the `download_and_extract` method downloads the file without extension and therefore cannot extract.

We have two options to resolve this
1. Instead of `download_and_extract` use a custom function to download and set the extension.
2. Update the `download_and_extract` function to receive the response header and set the extension. If we choose this we'll have to update other `make_dataset` methods accordingly.

Please let me know your thoughts on this.

Also as I have informed @samet-akcay , I had to pause the work on this until I get a green light on the current logical implementation ( As you might have seen, I had to classify images on the runtime instead of loading from directories because of the nature of the dataset). Please let me know whether the logic is good to go so that I can finish up.

Ravindu987:
Hey guys
I've gone ahead and completed writing docstrings and implemented a custom function to download and extract the dataset.
Since I've tested it locally I removed the WIP tag. Please have a look.

Could you refer me a documentation on the test convention used here. Once I get an insight on this I can finish it as well.

Thanks

src/anomalib/data/kolektor.py
```diff
@@ -0,0 +1,315 @@
...
+        )
+        logger.info("Checking the hash of the downloaded file.")
+        hash_check(downloaded_file_path, file_hash)
+
+    logger.info("Extracting dataset into given folder")
+    with ZipFile(downloaded_file_path, "r") as zip_file:
+        zip_file.extractall(root)
+
+    logger.info("Cleaning up files.")
+    (downloaded_file_path).unlink()
```
samet-akcay, line 82: There is a function to download and extract data [here](https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/data/utils/download.py#L220). Is there a specific difference between your implementation and this implementation? If not, this would be a duplication.

src/anomalib/data/kolektor.py
```diff
@@ -0,0 +1,315 @@
...
+    def prepare_data(self) -> None:
+        """Download the dataset if not available."""
+        if (self.root).is_dir():
+            logger.info("Found the dataset.")
+        else:
+            download_and_extract_kolektor(
+                "https://go.vicos.si/kolektorsdd",
+                self.root,
+                "2b094030343c1cd59df02203ac6c57a0",
+            )
```
samet-akcay, line 315: If you define `DOWNLOAD_INFO` variable as given below

```python
DOWNLOAD_INFO = DownloadInfo(
    name="kolektor",
    url="https://go.vicos.si/kolektorsdd",
    hash="2b094030343c1cd59df02203ac6c57a0",
)
```

Then it would be possible to use `anomalib.data.download.download_and_extract` function as:

```python
download_and_extract(self.root, DOWNLOAD_INFO)
```

This would be inline with the rest of the datamodule implementations. For reference, here are the some of the implementations:
[mvtec-ad](https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/data/mvtec.py#L52), [avenue](https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/data/avenue.py#L44), [btech](https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/data/btech.py#L38), [mvtec-3d](https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/data/mvtec_3d.py#L49)...

samet-akcay (changes requested):
Thanks @Ravindu987 for your great contribution! I have a comment regarding the use of download and extract function.

src/anomalib/data/kolektor.py
```diff
@@ -0,0 +1,315 @@
...
+    ValSplitMode,
+    get_transforms,
+)
+from anomalib.data.utils.download import hash_check
+
+logger = logging.getLogger(__name__)
+
+
+# Check if a mask shows defects
+def is_good(path):
```
samet-akcay, line 48: Not sure if this name is descriptive enough. Maybe your comment could be used as the name of the function such as `is_mask_anomalous or something similar?

src/anomalib/data/kolektor.py
```diff
@@ -0,0 +1,315 @@
...
+    img_arr = imread(path)
```
samet-akcay, line 49: is there any specific reason for using `imageio.imread` function? Is this in anomalib's list of requirements?

src/anomalib/data/kolektor.py
Ravindu987: There's a slight difference as I've mentioned [here](https://github.com/openvinotoolkit/anomalib/pull/983#issuecomment-1498479796) The `download_and_extract` function in `utils` downloads the Kolektor dataset without an extension and therefore it can't be extracted. This happens since the provided link for the dataset doesn't specify the extension which is required by the `download_and_extract` function. In this implementation I've hardcoded the .zip extension in line 65.

src/anomalib/data/kolektor.py
Ravindu987: This links with the previous issue which makes the `download_and_extract` function not usable here.

src/anomalib/data/kolektor.py
Ravindu987: Actually there was no specific reasons. Pillow or opencv seemed overkill for simply loading the image. Please let me know if I should use any other library.

Ravindu987:
Thank you very much @samet-akcay  for the review. I resolved one requested change and answered your questions as well. Please let me know your thoughts and any further changes.

src/anomalib/data/kolektor.py
samet-akcay: Maybe we could update the existing code a little bit. How about the following:

Replace
https://github.com/openvinotoolkit/anomalib/blob/d81dbb5b3dfd831576d137fb93881b34254a7fff/src/anomalib/data/utils/download.py#L25-L31

with:
```python
@dataclass
class DownloadInfo:
    """Info needed to download a dataset from a url."""

    name: str
    url: str
    hash: str
    filename: str | None = None
```

And replace
https://github.com/openvinotoolkit/anomalib/blob/d81dbb5b3dfd831576d137fb93881b34254a7fff/src/anomalib/data/utils/download.py#L229-L230

with,
```python
    # save the compressed file in the specified root directory, using the same file name as on the server
    if info.filename:
        downloaded_file_path = root / info.filename
    else:
        downloaded_file_path = root / info.url.split("/")[-1]
```

Then, I think `anomalib`'s download function should work. You wouldn't need to overwrite it.

Let me know if it works. Cheers!

Ravindu987: Thank you for the great suggestion. I've implemented and tested the change and it works fine. It has been updated in the latest commits.

src/anomalib/data/kolektor.py
Ravindu987: I didn't thought it would make much of a time difference here. In that case it would be good to switch. I'll do a small time comparison and switch to opencv if there's a significant difference.

Ravindu987: Opencv gave around a 25% decrease in time usage so I switched to opencv in the latest commit.

samet-akcay (approved):
lgtm, thanks a lot for your contribution!

Ravindu987:
Thanks for the review guys. 
The Codacy Static Code Analysis fails because  the `init` function in the DataModule class has too many arguments. I see no way around this as other implementations have the same issue.
Not sure why the Tox checks are failing. When I ran all pre-merge checks locally they all passed. Any ide on this?

samet-akcay:
> Thanks for the review guys. The Codacy Static Code Analysis fails because the `init` function in the DataModule class has too many arguments. I see no way around this as other implementations have the same issue. Not sure why the Tox checks are failing. When I ran all pre-merge checks locally they all passed. Any ide on this?

It's a problem with our CI that we need to fix. You don't need to do anything else at this point. Thank again for your contribution

Ravindu987:
> > Thanks for the review guys. The Codacy Static Code Analysis fails because the `init` function in the DataModule class has too many arguments. I see no way around this as other implementations have the same issue. Not sure why the Tox checks are failing. When I ran all pre-merge checks locally they all passed. Any ide on this?
> 
> It's a problem with our CI that we need to fix. You don't need to do anything else at this point. Thank again for your contribution

Noted with thanks. Hope to contribute more in the future.

PR #990, opened 2023-03-29
Set transformations from the config file

alexriedel1:
This PR allows to define image augmentations and transforms that are part of the albumentations module to be defined in the model config file. 

This becomes possible now:

```
dataset:
  name: mvtec
  format: mvtec
  path: ./datasets/MVTec
  task: segmentation
  category: bottle
  train_batch_size: 32
  test_batch_size: 32
  num_workers: 8
  image_size: 256 # dimensions to which images are resized (mandatory)
  center_crop: 224 # dimensions to which images are center-cropped after resizing (optional)
  normalization: imagenet # data distribution to which the images will be normalized: [none, imagenet]
  transform_config:
    train:
      Resize:
        height: 256
        width: 256
        always_apply: true
      Normalize:
        mean: [0.485, 0.456, 0.406]
        std: [0.229, 0.224, 0.225]
    eval:
      Resize:
        height: 256
        width: 256
        always_apply: true
      Normalize:
        mean: [0.485, 0.456, 0.406]
        std: [0.229, 0.224, 0.225]
```

It not only helps readability but also allows for image augmentations to be used in HPO sweeps


ashwinvaidya17 (approved):
Thanks! This is very useful

src/anomalib/utils/hpo/runners.py
djdameln, line 1: Can you explain the changes to this file?

alexriedel1: If you perform HPO sweeps on colab, the GPU memory isn't freed after each run and accumulates over time. To solve this, the model needs to be deleted and torch cache has to be cleared:
https://stackoverflow.com/a/57569472/11951277

samet-akcay (approved):

samet-akcay:
@alexriedel1, not sure you noticed, but pre-commit checks failed due to the following error

```
src/anomalib/utils/hpo/runners.py:6:1: I001 [*] Import block is un-sorted or un-formatted
Found 1 error.
```

alexriedel1:
> @alexriedel1, not sure you noticed, but pre-commit checks failed due to the following error
> 
> ```
> src/anomalib/utils/hpo/runners.py:6:1: I001 [*] Import block is un-sorted or un-formatted
> Found 1 error.
> ```

yess, working on it :)

sudhanv09:
Ah my bad i was on the release version, now I have installed this dev version, it works. Thanks

PR #996, opened 2023-03-31
[Custom loops POC] Support model loading

ashwinvaidya17:
# Description

- Since normalization metrics is stored in the anomaly module, this PR ensures that it is loaded and assigned correctly when the model is loaded from the weights file.
- Refactors the NormalizationMetrics


samet-akcay (commented):
gone through my first round of review. 
- Loading weights: as discussed, loading the object directly could potentially avoid all this manual loading. (Not 100% sure if it worked)
- Normalization: Some minor comments here and there

src/anomalib/trainer/loops/one_class/validate.py
```diff
@@ -26,7 +26,7 @@ def _evaluation_step_end(self, *args, **kwargs) -> STEP_OUTPUT | None:
         if outputs is not None:
             self.trainer.post_processor.apply_predictions(outputs)
             self.trainer.thresholder.update(outputs)
-            self.trainer.normalizer.update_metrics(outputs)
+            self.trainer.normalizer.update(outputs)
```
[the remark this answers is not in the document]
ashwinvaidya17: How about `update_normalization_statistics`?

src/anomalib/trainer/loops/one_class/test.py
```diff
@@ -48,6 +48,7 @@ def __init__(self) -> None:
 
     def on_run_start(self, *args, **kwargs) -> None:
         """Can be used to call setup."""
+        self.trainer.thresholder.initialize()
         self.trainer.metrics_manager.initialize()
```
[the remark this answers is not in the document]
ashwinvaidya17: Yeah, consistency would be better. I am fine with `metrics`.

src/anomalib/models/components/base/anomaly_module.py
djdameln, line 1: While I understand why these changes are needed, I feel they clutter the anomaly module, which was previously cleaned up so nicely by your previous PR. Maybe we can think of an alternative design that does not involve complicated logic for loading the model. Would there be other ways in which we can load the normalization stats and thresholds from the checkpoint without including it in the model? (The model itself does not use these attributes anyway, we only have it here for persistence after re-loading).

djdameln: As discussed, we could override the `save_checkpoint` method in the `AnomalibTrainer`.

djdameln (changes requested):

src/anomalib/trainer/utils/normalizer/base.py
```diff
@@ -0,0 +1,48 @@
...
+        return self.trainer.lightning_module
+
+    @property
+    def metric(self) -> Metric:
+        """Assigns metrics if not available."""
+        # get metric from anomaly_module
+        if not hasattr(self.trainer.lightning_module, "normalization_metrics"):
+            self.trainer.lightning_module.normalization_metrics = self.metric_class()
+        # Every time metric is called, it is moved to the cpu
+        return self.trainer.lightning_module.normalization_metrics.cpu()
```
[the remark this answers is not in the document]
ashwinvaidya17: I think this is somethig might have to address later. We move the metrics and outputs to cpu at a bunch of places. We have this `PostProcessor._outputs_to_cpu(outputs)` in post-processor. I guess the original idea was to avoid oom issues by moving metrics computation to the cpu. But in this case we can look into adding a new key to the config to allow gpu computation.

ashwinvaidya17:
I've updated the PR with partial changes while we finalize the naming convention regarding the normalizer and metrics manager.

samet-akcay (approved):
I think I'm alright with the overall approach for now. Thanks for your effort

src/anomalib/trainer/utils/checkpoint_connector.py
```diff
@@ -0,0 +1,59 @@
...
+            checkpoint["normalization_metric"] = self.trainer.normalizer.metric
+        return checkpoint
+
+    def restore_training_state(self) -> None:
+        """Loads the thresholding and normalization classes from the checkpoint."""
+        super().restore_training_state()
+        self._load_metrics()
+
+    def _load_metrics(self) -> None:
+        """Loads the thersholding and normalization classes from the checkpoint."""
```
djdameln, line 50: typo: `thresholding`

src/anomalib/deploy/export.py
djdameln, line 1: Not sure if it's in the scope of this PR, but export functionality is broken, because it tries to retrieve threshold information from the lightning model, which is now stored in trainer.

djdameln (commented):
Thanks, I'm happy with the changes. Just one concern about the export, but maybe you were planning to address this in a different PR.

PR #1001, opened 2023-04-05
Remove config from torch inferencer

samet-akcay:
# Description

- Remove `config` from `TorchInferencer`

## Changes

## Checklist


src/anomalib/models/cfa/config.yaml
```diff
@@ -62,7 +62,7 @@ logging:
   log_graph: false # Logs the model graph to respective logger.
 
 optimization:
-  export_mode: null #options: onnx, openvino
+  export_mode: null # options: torch, onnx, openvino
```
djdameln, line 65: What if a user would like to export to both torch and openvino/onnx format?

src/anomalib/deploy/inferencers/torch_inferencer.py
```diff
@@ -29,38 +25,20 @@ class TorchInferencer(Inferencer):
     """PyTorch implementation for the inference.
 
     Args:
-        config (str | Path | DictConfig | ListConfig): Configurable parameters that are used
-            during the training stage.
-        model_source (str | Path | AnomalyModule): Path to the model ckpt file or the Anomaly model.
-        metadata_path (str | Path, optional): Path to metadata file. If none, it tries to load the params
-                from the model state_dict. Defaults to None.
+        path (str | Path | AnomalyModule): Path to the model ckpt file or the Anomaly model.
```
djdameln, line 28: This needs to be updated (I think we only support `str` and `Path` now, right?)

src/anomalib/deploy/inferencers/torch_inferencer.py
```diff
@@ -122,18 +94,7 @@ def pre_process(self, image: np.ndarray) -> Tensor:
...
-
-        image_size = (self.config.dataset.image_size[0], self.config.dataset.image_size[1])
-        center_crop = self.config.dataset.get("center_crop")
-        if center_crop is not None:
-            center_crop = tuple(center_crop)
-        normalization = InputNormalizationMethod(self.config.dataset.normalization)
-        transform = get_transforms(
-            config=transform_config, image_size=image_size, center_crop=center_crop, normalization=normalization
-        )
+        transform = A.from_dict(self.metadata["transform"])
```
djdameln, line 97: This way, we instantiate the transforms object with every call. Maybe it would be better to load the transforms in the constructor and store them as `self.transform`

djdameln (changes requested):
Thanks, my main question is do we want to support exporting to both torch and openvino format within the same run. In that case we would need to make some changes to how we configure the export.

src/anomalib/models/cfa/config.yaml
samet-akcay: This will change with the new CLI. In this case, we would only add a temporary change to this.

djdameln: Alright, let's leave this for later then

djdameln (approved):
Thanks!

PR #1023, opened 2023-04-21
Configure reference frame for multi-frame video clips

djdameln:
# Description

- In multi-frame video pipelines, we need to set which frame acts as the reference frame, meaning the ground truth information of that frame will be used to represent the (multi-frame) video clip.
- This PR makes this configurable from the config. The user may elect to take the GT info from the first frame, last frame, mid frame or all frames combined.

## Changes

## Checklist

### DIFF
--- a/src/anomalib/data/base/video.py
+++ b/src/anomalib/data/base/video.py
@@ -38,6 +38,7 @@ class AnomalibVideoDataset(AnomalibDataset, ABC):
         transform (A.Compose): Albumentations Compose object describing the transforms that are applied to the inputs.
         clip_length_in_frames (int): Number of video frames in each clip.
         frames_between_clips (int): Number of frames between each consecutive video clip.
+        target_frame (VideoTargetFrame): Specifies the target frame in the video clip, used for ground truth retrieval
     """
 
     def __init__(
--- a/src/anomalib/data/base/video.py
+++ b/src/anomalib/data/base/video.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from abc import ABC
+from enum import Enum
 from typing import Callable
 
 import albumentations as A
@@ -17,6 +18,18 @@ from anomalib.data.utils import ValSplitMode, masks_to_boxes
 from anomalib.data.utils.video import ClipsIndexer
 
 
+class VideoTargetFrame(str, Enum):
+    """Target frame for a video-clip.
+
+    Used in multi-frame models to determine which frame's ground truth information will be used.
+    """
+
+    FIRST = "first"
+    LAST = "last"
+    MID = "mid"
+    ALL = "all"
+
+
 class AnomalibVideoDataset(AnomalibDataset, ABC):
     """Base video anomalib dataset class.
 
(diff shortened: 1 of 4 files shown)


src/anomalib/data/base/video.py
```diff
@@ -68,6 +88,28 @@ def _setup_clips(self) -> None:
             frames_between_clips=self.frames_between_clips,
         )
 
+    def _select_targets(self, item):
+        if self.target_frame == VideoTargetFrame.FIRST:
+            idx = 0
+        elif self.target_frame == VideoTargetFrame.LAST:
+            idx = -1
+        elif self.target_frame == VideoTargetFrame.MID:
+            idx = int(self.clip_length_in_frames / 2)
```
samet-akcay, line 97: Is this how it is usually handled in the literature? Would it be an idea to just use an `index` argument to specify a specific frame rather than limiting it to first, mid and last?

src/anomalib/post_processing/visualizer.py
```diff
@@ -88,11 +88,13 @@ def visualize_batch(self, batch: dict) -> Iterator[np.ndarray]:
...
             Generator that yields a display-ready visualization for each image.
         """
-        batch_size, _num_channels, height, width = batch["image"].size()
+        batch_size = batch["image"].shape[0]
         for i in range(batch_size):
             if "image_path" in batch:
+                height, width = batch["image"].shape[-2:]
                 image = read_image(path=batch["image_path"][i], image_size=(height, width))
             elif "video_path" in batch:
+                _, height, width, _ = batch["original_image"].shape
```
samet-akcay, line 97: Just to follow the same convention with Lines 91 and 94, would it be an idea to do something like `height, width = batch["original_image"].shape[1:3]`

samet-akcay: Or, Lines 91 and 94 could possibly use Line 97's convention?

samet-akcay (commented):
Thanks, only few comments

src/anomalib/data/base/video.py
```diff
@@ -28,7 +41,12 @@ class AnomalibVideoDataset(AnomalibDataset, ABC):
     """
```
ashwinvaidya17, line 41: The doc string needs to be updated.

ashwinvaidya17 (commented):
Thanks. One minor comment

src/anomalib/data/base/video.py
djdameln: One option would be to use -1 for last frame, and -2 for all frames. Would that be an idea?

samet-akcay: This is probably an edge case. I'm not sure if a user would use a random index in the first place. 

If last frame is most common, then we could maybe stick to your approach. -2 for all might be confusing to users.

samet-akcay (approved):

src/anomalib/data/base/video.py
ashwinvaidya17: I feel passing an index might be confusing. However, the only solution I can think of is to support `VideoTargetFrame.index` like this https://stackoverflow.com/questions/55996561/support-unknown-values-in-python-enums. So , we can have `VideoTargetFrame.20` to return frame number 20. But might be a bit hacky.

ashwinvaidya17 (approved):
Thanks. Looks good to me

PR #1038, opened 2023-04-25
Make anomalib PEP 561 compliant for mypy

### LINKED ISSUE #1015: [Task]: Make anomalib PEP 561 compliant for mypy

When installing anomalib as a package, mypy starts to complain about missing type stubs. Seeing that anomalib is mostly (if not fully) typed, it would be nice to make the package PEP561 compliant (currently I ignore `anomalib` for mypy)

Make the package PEP 561 compliant. Could potentially be as easy as adding a `py.typed` at the root of the shipped package directory. Note that I am not an expert in building/shipping python packages, so I might also be wrong on this

### Additional context

https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

WenjingKangIntel:
# Description

This PR aims to make anomalib PEP 561 compliant for mypy:
1. Add py.typed file in package directory.
2. Include py.typed in MANIFEST.in.
reference:
https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

If we have a `test_mypy.py` for testing with the following content:
`
from anomalib.config import get_configurable_parameters, update_nncf_config
`
Before this PR, when running: `mypy test_pep561.py`, it will generate the following error.
```
test_pep561.py:1: error: Skipping analyzing "anomalib.config": module is installed, but missing library stubs or py.typed marker  [import]
test_pep561.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 1 source file)
```
After this PR, it will have the following output:
`Success: no issues found in 1 source file
`
This shows that this PR can make anomalib PEP 561 compliant for mypy.
- Fixes #1015

## Changes

## Checklist

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,4 @@
 recursive-include requirements *
+
+# Include marker file for PEP 561
+include src/anomalib/py.typed


samet-akcay (approved):
lgtm, thanks for the contribution!

ashwinvaidya17 (commented):
Do you think we need to also include a line in `setup.py`. Something like
```python
 package_data={"anomalib": ["py.typed"]},
```

WenjingKangIntel:
> Do you think we need to also include a line in `setup.py`. Something like
> 
> ```python
>  package_data={"anomalib": ["py.typed"]},
> ```

Hi, Vaidya,
From the link:
https://setuptools.pypa.io/en/latest/userguide/datafiles.html#summary
package_data specifies additional patterns to match files that may or may not be matched by MANIFEST.in. The results turn out the content in MANIFEST.in is matched.

Some more information about:
https://packaging.python.org/en/latest/guides/using-manifest-in/
all files specified by the package_data and data_files setup() arguments and the commands in MANIFEST.in (if such a file exists) are executed in order to add and remove further files to and from the sdist.

Conclusion:
Please double check. From my understanding, adding "py.typed" to package_data is not needed.

ashwinvaidya17:
> > Do you think we need to also include a line in `setup.py`. Something like
> > ```python
> >  package_data={"anomalib": ["py.typed"]},
> > ```
> 
> Hi, Vaidya, From the link: https://setuptools.pypa.io/en/latest/userguide/datafiles.html#summary package_data specifies additional patterns to match files that may or may not be matched by MANIFEST.in. The results turn out the content in MANIFEST.in is matched.
> 
> Some more information about: https://packaging.python.org/en/latest/guides/using-manifest-in/ all files specified by the package_data and data_files setup() arguments and the commands in MANIFEST.in (if such a file exists) are executed in order to add and remove further files to and from the sdist.
> 
> Conclusion: Please double check. From my understanding, adding "py.typed" to package_data is not needed.

I tried using both `python -m build` and `python setup.py sdist bdist_wheel` to generate distribution package. When I use `python -m build`, the `py.typed` file is present in the distribution however, if I use `python setup.py ...` then this file is missing. So, when we upload to pip, the inclusion of this file depends on what method the pip uploader uses to generate the distribution. Not sure what's the standard approach to generate distributions. @samet-akcay any thoughts on this?

samet-akcay:
> > > Do you think we need to also include a line in `setup.py`. Something like
> > > ```python
> > >  package_data={"anomalib": ["py.typed"]},
> > > ```
> > 
> > 
> > Hi, Vaidya, From the link: https://setuptools.pypa.io/en/latest/userguide/datafiles.html#summary package_data specifies additional patterns to match files that may or may not be matched by MANIFEST.in. The results turn out the content in MANIFEST.in is matched.
> > Some more information about: https://packaging.python.org/en/latest/guides/using-manifest-in/ all files specified by the package_data and data_files setup() arguments and the commands in MANIFEST.in (if such a file exists) are executed in order to add and remove further files to and from the sdist.
> > Conclusion: Please double check. From my understanding, adding "py.typed" to package_data is not needed.
> 
> I tried using both `python -m build` and `python setup.py sdist bdist_wheel` to generate distribution package. When I use `python -m build`, the `py.typed` file is present in the distribution however, if I use `python setup.py ...` then this file is missing. So, when we upload to pip, the inclusion of this file depends on what method the pip uploader uses to generate the distribution. Not sure what's the standard approach to generate distributions. @samet-akcay any thoughts on this?

Maybe we could verify this by uploading to test-pypi?
https://packaging.python.org/en/latest/guides/using-testpypi/

samet-akcay:
We could merge this once the CI passes.

ORippler:
I just did an install of `anomalib==0.5.0` from pypi and `py.typed` was not present in the anomalib folder, so it's not yet fully compliant.

PR #1052, opened 2023-04-27
Enable AnomalyScoreMeanVarThreshold based on three-sigma 

### LINKED ISSUE #1027: [Task]: Enabling Unsupervised Workflow: Propose a new thresholding approach
### Anomaly Detection
Anomaly detection is the process of identifying data points, events, or observations within a dataset that significantly deviate from the normal or expected behavior. Anomalies may be caused by a variety of factors, including failures, fraud, or unusual behavior.

Anomaly detection is an essential task in numerous industries, including cybersecurity, finance, healthcare, and manufacturing. It can be used to detect fraud in financial transactions, to identify anomalies in medical data for early disease diagnosis, to detect flaws in manufacturing processes, and to monitor traffic for security threats.

### Anomalib
Anomalib is a deep learning library that aims to collect the best anomaly detection algorithms for testing on both public and private datasets. Anomalib offers several ready-to-use implementations of anomaly detection algorithms described in recent research, as well as a set of tools that make it easier to build and use custom models. The library has a strong focus on image-based anomaly detection, where the goal of the algorithm is to identify anomalous images or anomalous pixel regions within images in a dataset.

## The thresholding problem

Anomaly detection models in Anomalib are trained only on normal images. During inference, the models are tasked with distinguishing anomalous samples from normal samples. The task is similar to a classical binary classification problem, but instead of generating a class label and a confidence score, Anomalib models generate an anomaly score, which quantifies the distance of the sample to the distribution of normal samples seen during training. The range of possible anomaly score values is unbounded and may differ widely between models and/or datasets, which makes it challenging to set a good threshold for mapping the raw anomaly scores to a binary class label (normal vs. anomalous).

## Desired solution
Anomalib currently has an [adaptive thresholding mechanism](https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/utils/metrics/anomaly_score_threshold.py) in place which aims to address the thresholding problem. The adaptive thresholding mechanism computes the F1 score over a validation set for a range of thresholds. The final threshold value is the threshold value that results in the highest F1 score. A major drawback of this approach is that the validation set is required to contain anomalous samples, which might not always be available in real world anomaly detection problems. 

**The goal of this hackathon is to design a fully unsupervised thresholding mechanism that does not rely on anomalous samples.** 

#### Possible approaches
- One possibility is to generate fake anomalous images for the validation set, and then use Anomalib's existing adaptive thresholding mechanism. A simple version of this solution, which makes use of random Perlin noise masks, is already [implemented](https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/data/synthetic.py) in Anomalib. Different noise methods or generative models could possibly improve the results of this method.
- Use the characteristics of the anomaly score distribution of the normal training images to set a confidence interval.
- Other, better, solutions may exist. Be creative!

### Additional context

_No response_

WenjingKangIntel:
# Description

1. Use train data as validation data during validation and collect all the predictions
2. Based on the predictions above, perform three sigma rule to determine the final threshold: `mean - 3 * standard_deviation` or` mean + 3 * standard_deviation`, +/- is determined by config. Currently, only the + scenario should be used, because currently this repo determines the result via checking whether the predictions is **larger** than the threshold.

- Fixes #1027

## Changes

## Checklist

### DIFF
--- a/src/anomalib/utils/metrics/anomaly_score_threshold.py
+++ b/src/anomalib/utils/metrics/anomaly_score_threshold.py
@@ -81,8 +81,8 @@ class AnomalyScoreMeanVarThreshold(AnomalyScoreThreshold):
     MEANVAR_MODE = MeanVarianceMode.ADD
 
     def __init__(self, default_value: float = 0.5, **kwargs) -> None:
-        self.meanvar_mode = self.MEANVAR_MODE
-        self.std_dev_ratio = self.STD_DEV_RATIO
+        self.meanvar_mode = kwargs.get("meanvar_mode", self.MEANVAR_MODE)
+        self.std_dev_ratio = kwargs.get("std_dev_ratio", self.STD_DEV_RATIO)
         super().__init__(default_value=default_value, **kwargs)
 
     def compute(self) -> Tensor:
--- a/src/anomalib/utils/metrics/anomaly_score_threshold.py
+++ b/src/anomalib/utils/metrics/anomaly_score_threshold.py
@@ -81,8 +81,8 @@ class AnomalyScoreMeanVarThreshold(AnomalyScoreThreshold):
     MEANVAR_MODE = MeanVarianceMode.ADD
 
     def __init__(self, default_value: float = 0.5, **kwargs) -> None:
-        self.meanvar_mode: MeanVarianceMode
-        self.std_dev_ratio: float
+        self.meanvar_mode = self.MEANVAR_MODE
+        self.std_dev_ratio = self.STD_DEV_RATIO
         super().__init__(default_value=default_value, **kwargs)
 
     def compute(self) -> Tensor:
--- a/src/anomalib/utils/metrics/anomaly_score_threshold.py
+++ b/src/anomalib/utils/metrics/anomaly_score_threshold.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import warnings
+from enum import Enum
 
 import torch
 from torch import Tensor
@@ -50,7 +51,6 @@ class AnomalyScoreThreshold(PrecisionRecallCurve):
                 "to poor predictions. For a more reliable adaptive threshold computation, please add some anomalous "
                 "images to the validation set."
             )
-
         precision, recall, thresholds = super().compute()
         f1_score = (2 * precision * recall) / (precision + recall + 1e-10)
         if thresholds.dim() == 0:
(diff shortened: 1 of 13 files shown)


WenjingKangIntel:
test on:
model: padim
config:
```
dataset:
  name: mvtec
  format: mvtec
  path: /home/kangwenj/hackathon/data/mvtec_anomaly_detection
  category: bottle
  task: segmentation
val_split_mode: same_as_train

threshold:
  method: meanvar
  std_dev_ratio: 3
  meanvar_mode: add
```
result:
| Test metric        | DataLoader 0|
| ----------- | ----------- |
| image_AUROC          | 0.9460317492485046     |
| image_F1Score         | 0.9402984976768494|
| pixel_AUROC          | 0.9814534187316895|
| pixel_F1Score         | 0.5805442333221436|

if using adaptive and val_split_mode: same_as_test:
result:
| Test metric        | DataLoader 0|
| ----------- | ----------- |
| image_AUROC          | 0.9936507940292358     |
| image_F1Score         | 0.9763779044151306|
| pixel_AUROC          | 0.983005702495575|
| pixel_F1Score         |0.72203129529953|


Actually, increasing std_dev_ratio to 10, can get a better result for image_AUROC/image_F1Score:
       image_AUROC          0.9936507940292358
      image_F1Score         0.9677419662475586
but the pixel_AUROC/pixel_F1Score is not good:
     pixel_AUROC          0.7939819097518921
      pixel_F1Score         0.3183630406856537

There might be two different sweet spots for image/pixel threshold.

src/anomalib/utils/metrics/anomaly_score_threshold.py
```diff
@@ -60,3 +60,51 @@ def compute(self) -> Tensor:
...
+class AnomalyScoreMeanVarThreshold(AnomalyScoreThreshold):
+    """Anomaly Score Threshold.
+
+    This class computes/stores the threshold that determines the anomalous label
+    given anomaly scores. The computation is based on the mean and variance.
+
+    """
+
+    STD_DEV_RATIO = 0.1
+    MEANVAR_MODE = MeanVarianceMode.ADD
```
ashwinvaidya17, line 81: Is it possible to avoid global variables? Can we add it to the `__init__` args?

src/anomalib/utils/metrics/anomaly_score_threshold.py
```diff
@@ -60,3 +60,51 @@ def compute(self) -> Tensor:
...
+            warnings.warn(
+                "The validation set does not contain any anomalous images. As a result, the adaptive threshold will "
+                "take the value of the highest anomaly score observed in the normal validation images, which may lead "
+                "to poor predictions. For a more reliable adaptive threshold computation, please add some anomalous "
+                "images to the validation set."
+            )
+        mean_pred = torch.mean(torch.cat(self.preds, dim=0))
+        std_pred = torch.std(torch.cat(self.preds, dim=0))
+        if self.meanvar_mode == MeanVarianceMode.MINUS:
+            self.value = mean_pred - self.std_dev_ratio * std_pred
```
ashwinvaidya17, line 107: As you mentioned in your description this does not make sense in the anomaly context. Maybe we shouldn't support `minus` operation.

ashwinvaidya17 (changes requested):
Thanks. It looks good overall. Just minor comments. Before this can be merged we need to discuss the design for adding more thresholding methods internally.

src/anomalib/utils/metrics/anomaly_score_threshold.py
WenjingKangIntel: We may add it as a placeholder. If in the future, if we use negative outputs of the model, then minus mode can be used?

src/anomalib/utils/metrics/anomaly_score_threshold.py
WenjingKangIntel: fixed.

PR #1061, opened 2023-04-28
Feature: Use Perlin ROI to create noise on to the object of interest

leonardloh:
# Description

- Our aim is to generate fake anomalous images for the validation set, and then use Anomalib's existing adaptive thresholding mechanism. But the current Perlin Noise augmentation will cover the whole image instead of the object of interest.
- To automatically segment out the ROI Mask of the object given an image and then implement Perlin Noise onto the object only.

Code changes:

- Subclass `Augmenter` to create another class call `PerlinROIAugmenter`
- Create PerlinROI methods in the `PerlinROIAugmenter`
- Enable `Dataset` to access synthetic type (default: Perlin) from `config.yaml`

## Changes

## Checklist


src/anomalib/models/padim/config.yaml
```diff
@@ -13,9 +13,11 @@ dataset:
   transform_config:
     train: null
     eval: null
-  test_split_mode: from_dir # options: [from_dir, synthetic]
+  test_split_mode: synthetic # options: [from_dir, synthetic]
+  test_synthetic_type: perlin_roi
```
ashwinvaidya17, line 17: Maybe these should live under a separate key. Since `synthetic` is optional `test_synthetic_mode` probably shouldn't be at the same level as other keys.
How about?

```yaml
test_split:
    mode: synthetic/from_dir
    optional_key1: ...
    optional_key2: ...
    ...
```

tests/pre_merge/datasets/test_synthetic_data.py
```diff
@@ -30,24 +30,53 @@ def get_folder_dataset():
...
-        synthetic_dataset = SyntheticAnomalyDataset.from_dataset(folder_dataset)
+        synthetic_dataset = SyntheticAnomalyDataset.from_dataset(folder_dataset, TestSyntheticType.PERLIN)
         return synthetic_dataset
 
     return make
 
 
 @pytest.fixture(autouse=True)
-def synthetic_dataset_from_samples():
+def make_perlinroi_synthetic_dataset():
```
ashwinvaidya17, line 45: Since both the methods are very similar, we can use `mark.parameterize` https://docs.pytest.org/en/6.2.x/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions

src/anomalib/data/utils/augmenter.py
```diff
@@ -168,3 +170,75 @@ def augment_batch(self, batch: Tensor) -> tuple[Tensor, Tensor]:
...
+            if torch.rand(1) > self.p_anomalous:  # include normal samples
+                perturbations_list.append(torch.zeros((channels, height, width)))
+                masks_list.append(torch.zeros((1, height, width)))
+            else:
+                anomaly_source_path = (
+                    random.sample(self.anomaly_source_paths, 1)[0] if len(self.anomaly_source_paths) > 0 else None
+                )
+                perturbation, mask = self.generate_perturbation(height, width, anomaly_source_path)
+                perturbations_list.append(Tensor(perturbation).permute((2, 0, 1)))
+                masks_list.append((Tensor(mask).permute((2, 0, 1)) * thresh_batch))
```
ashwinvaidya17, line 202: Except for `thresh_batch` here this is similar to the base method. Maybe we can just add an if statement in the base class and add this line there.

src/anomalib/data/utils/augmenter.py
```diff
@@ -168,3 +170,75 @@ def augment_batch(self, batch: Tensor) -> tuple[Tensor, Tensor]:
...
+        else:
+            raise ValueError("Beta must be either float or tuple of floats")
+
+        augmented_batch = batch * (1 - masks) + (beta) * (perturbations * masks) + (1 - beta) * batch * (masks)
+
+        return augmented_batch, masks
+
+    def gaussian_blur(self, image: Tensor, transform: A.Compose) -> tuple[Tensor, Tensor]:
+        """Generate Gaussian Blur.
+
```
ashwinvaidya17, line 222: Can you also describe the intention behind performing gaussian blur and finding out the thresholds based on it.

src/anomalib/data/utils/augmenter.py
```diff
@@ -168,3 +170,75 @@ def augment_batch(self, batch: Tensor) -> tuple[Tensor, Tensor]:
...
+            - Augmented image to which anomalous perturbations have been added.
+            - Ground truth masks corresponding to the anomalous perturbations.
+        """
+        gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+        blurred = cv2.GaussianBlur(gray, (5, 5), 0)
+        (T, thresh) = cv2.threshold(blurred, 0, 255, cv2.THRESH_BINARY | cv2.THRESH_OTSU)
+
+        cv2.bitwise_not(thresh)
+
+        if thresh[0, 0] & thresh[-1, -1] & thresh[-1, 0] & thresh[0, -1] == 255:
```
ashwinvaidya17, line 238: Can you explain this line? From what I understand this inverts the image if the background is white

src/anomalib/data/utils/augmenter.py
```diff
@@ -168,3 +170,75 @@ def augment_batch(self, batch: Tensor) -> tuple[Tensor, Tensor]:
         augmented_batch = batch * (1 - masks) + (beta) * perturbations + (1 - beta) * batch * (masks)
 
         return augmented_batch, masks
+
+
+class PerlinROIAugmenter(Augmenter):
+    def augment_batch(self, batch: Tensor, thresh_batch: Tensor | None = None) -> tuple[Tensor, Tensor]:
```
ashwinvaidya17, line 176: Can you add a short description on how this works. Something like, extracts foreground object based on binary thresholding and limits perlin noise augmentation to that region.

ashwinvaidya17 (changes requested):
This is a great addition! Thanks. The only thing I am concerned about is the keys in config. It might require some design. But it is something we will have to discuss internally.

src/anomalib/models/padim/config.yaml
kxtan: Hi Ashwin, thanks for the suggestion. We will work on the changes and keep you updated.

tests/pre_merge/datasets/test_synthetic_data.py
kxtan: Noted, we will work on the changes. Thanks again Ashwin

src/anomalib/data/folder_3d.py
```diff
@@ -292,8 +294,10 @@ class Folder3D(AnomalibDataModule):
             during validation.
             Defaults to None.
         test_split_mode (TestSplitMode): Setting that determines how the testing subset is obtained.
+        test_synthetic_type (TestSyntheticType): Method for generating synthetic test data.
```
djdameln, line 297: I don't know if it makes sense to expose these parameters to the 3d datasets. The anomalous augmentations are only applied to the images, not to the 3d point cloud data.

djdameln (commented):
Thanks. This could be a useful addition to our Perlin noise mechanism. Overall I have the same comments as Ashwin, in that some more explanation about the methods used to obtain the final noise masks could be provided in the docstrings.

src/anomalib/models/padim/config.yaml
leonardloh: hi @ashwinvaidya17 can we proceed with this design? We would like your approval before we make any changes as many codes will be impacted.

```
test_split:  
    mode: synthetic # options: [from_dir, synthetic]
      type: perlin_roi
    ratio: 0.2
```

src/anomalib/data/utils/augmenter.py
kxtan: we renamed the function `generate_roi` as it does more than gaussian blur. The intention is to generate the region of interest.

src/anomalib/data/utils/augmenter.py
kxtan: good idea, added this and removed the child method

src/anomalib/data/utils/augmenter.py
kxtan: added as the docstring for this class

src/anomalib/data/utils/augmenter.py
kxtan: similar to your understanding: if all four corners of the image are white, we invert the mask.
we added this as a comment in the code base as well

src/anomalib/data/folder_3d.py
kxtan: thanks @djdameln, we removed it from this dataset. please let us know if there are more.

src/anomalib/data/utils/augmenter.py
```diff
@@ -126,11 +127,12 @@ def generate_perturbation(
 
         return perturbation, mask
 
-    def augment_batch(self, batch: Tensor) -> tuple[Tensor, Tensor]:
+    def augment_batch(self, batch: Tensor, thresh_batch: Tensor | None = None) -> tuple[Tensor, Tensor]:
         """Generate anomalous augmentations for a batch of input images.
 
         Args:
             batch (Tensor): Batch of input images
+            thresh_batch (Tensor | None, optional): Batch of
```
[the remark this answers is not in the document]
kxtan: added comment and refactored the class based on @ashwinvaidya17's suggestions

PR #1073, opened 2023-05-05
EfficientAD

alexriedel1:
# Description

Hey, 
this is the work in progress for implementing EfficientAD based on the implementations https://github.com/nelson1425/EfficientAD/blob/main/efficientad.py and https://github.com/rximg/EfficientAD/tree/main

Right now the performance doesn't meet the paper results but I'm on it :)

- Fixes # (issue)

## Changes

## Checklist


samet-akcay:
Wow @alexriedel1, this is great contribution! Thanks a lot!

samet-akcay:
> Right now I'm searching for a way to iterate through the validation dataloader at the very last epoch and save the model via the PL Lightning ModelCheckpointer after that. Does anyone have a hint on how to do this?

Regarding your question, the quickest way to do this would be something like the following. (Haven't checked if it works, though. You might want to double check)

```python
class EfficientAD(AnomalyModule):
    """PL Lightning Module for the EfficientAD algorithm."""
    ...

    def on_validation_epoch_end(self) -> None:
        if self.current_epoch == self.trainer.max_epochs:   # or `self.trainer.max_epochs - 1`? Need to check
            for i, data in enumerate(self.trainer.datamodule.val_dataloader()):
                print("You could save the model here.")
```

alexriedel1:
> > Right now I'm searching for a way to iterate through the validation dataloader at the very last epoch and save the model via the PL Lightning ModelCheckpointer after that. Does anyone have a hint on how to do this?
> 
> Regarding your question, the quickest way to do this would be something like the following. (Haven't checked if it works, though. You might want to double check)
> 
> ```python
> class EfficientAD(AnomalyModule):
>     """PL Lightning Module for the EfficientAD algorithm."""
>     ...
> 
>     def on_validation_epoch_end(self) -> None:
>         if self.current_epoch == self.trainer.max_epochs:   # or `self.trainer.max_epochs - 1`? Need to check
>             for i, data in enumerate(self.trainer.datamodule.val_dataloader()):
>                 print("You could save the model here.")
> ```

Thanks! Yes I now went for on validation start, to have the quantiles calculated for the actual final validation


https://github.com/openvinotoolkit/anomalib/blob/2eb33bf7c33cf8439e12d630304f7f4318926f67/src/anomalib/models/efficientad/lightning_model.py#L165-L172

samet-akcay:
Yeah, this looks good!

src/anomalib/models/efficientad/config.yaml
```diff
@@ -0,0 +1,103 @@
+dataset:
+  name: mvtec
+  format: mvtec
+  path: ./datasets/MVTec
+  category: bottle
+  task: segmentation
+  train_batch_size: 1
+  eval_batch_size: 1
```
nelson1425, line 8: Evaluation can be made faster with larger batch size

src/anomalib/models/efficientad/config.yaml
```diff
@@ -0,0 +1,103 @@
...
+  enable_checkpointing: true
+  default_root_dir: null
+  gradient_clip_val: 0
+  gradient_clip_algorithm: norm
+  num_nodes: 1
+  devices: 1
+  enable_progress_bar: true
+  overfit_batches: 0.0
+  track_grad_norm: -1
+  check_val_every_n_epoch: 1 # Don't validate before extracting features.
```
nelson1425, line 73: Training of EfficientAD is steps instead of epochs, so validate every 1000 steps?

```
  check_val_every_n_epoch: null
  val_check_interval: 1000 # Validate every 1000 steps
```

src/anomalib/models/efficientad/config.yaml
```diff
@@ -0,0 +1,103 @@
...
+  fast_dev_run: false
+  accumulate_grad_batches: 1
+  max_epochs: 5
```
nelson1425, line 76:
```
  max_epochs: -1
  min_epochs: null
  max_steps: 70000 # EfficientAD paper recommend 70000 steps
  min_steps: 70000
```

src/anomalib/models/efficientad/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""Deep Feature Kernel Density Estimation model."""
```
nelson1425, line 1:
```
"""EfficientAD: Accurate Visual Anomaly Detection at Millisecond-Level Latencies.
https://arxiv.org/pdf/2303.14535.pdf
"""
```

src/anomalib/models/efficientad/config.yaml
```diff
@@ -0,0 +1,103 @@
+dataset:
+  name: mvtec
+  format: mvtec
+  path: ./datasets/MVTec
+  category: bottle
+  task: segmentation
+  train_batch_size: 1
+  eval_batch_size: 1
+  num_workers: 0
```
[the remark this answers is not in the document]
alexriedel1: well depends on the cpu..

src/anomalib/models/efficientad/torch_model.py
```diff
@@ -0,0 +1,350 @@
...
+
+        if self.training:
+            # Student loss
+            d_hard = torch.quantile(distance_st, 0.999)
+            loss_hard = torch.mean(distance_st[distance_st >= d_hard])
+            student_imagenet_output = self.student(batch_imagenet)[:, :self.teacher_out_channels, :, :]
+            loss_st = loss_hard + (1 / (c * h * w)) * torch.sum(torch.pow(student_imagenet_output, 2))
+
+            # Autoencoder and Student AE Loss
+            aug_img = self.choose_random_aug_image(batch)
```
nelson1425, line 315: If the have imagenet normalization, then we must undo imagenet normalization, augment with PIL and redo imagenet normalization. Like this:


```python
class EfficientAdModel(nn.Module):
    def __init__(
        self,
        model_size: Literal['small', 'medium']):
        super().__init__()
        
        # create models
        self._out_channels = 384
        if size == 'small':
            self.teacher = get_pdn_small(self._out_channels)
            self.student = get_pdn_small(2 * self._out_channels)
        elif size == 'medium':
            self.teacher = get_pdn_medium(self._out_channels)
            self.student = get_pdn_medium(2 * self._out_channels)
        else:
            raise Exception()
        # state_dict = torch.load(config.weights, map_location='cpu')
        # teacher.load_state_dict(state_dict)
        self.teacher_norm = nn.BatchNorm2d(num_features=self._out_channels, momentum=None, affine=False)
        self.autoencoder = get_autoencoder(self._out_channels)
        
        self.autoencoder_transform = transforms.Compose([
            # undo the imagenet normalization
            transforms.Normalize(mean = [0., 0., 0.],
                                 std = [1/0.229, 1/0.224, 1/0.225]),
            transforms.Normalize(mean = [-0.485, -0.456, -0.406],
                                 std = [ 1., 1., 1. ]),
            transforms.ToPILImage(),
            # augmentation
            transforms.RandomChoice([
                transforms.ColorJitter(brightness=0.2),
                transforms.ColorJitter(contrast=0.2),
                transforms.ColorJitter(saturation=0.2)
            ]),
            # back to pytorch
            transforms.ToTensor(),
            transforms.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225])        
        ])
        
    def forward(self, images: Tensor) -> tuple[Tensor, Tensor]:
        with torch.no_grad():
            teacher_output = self.teacher(images)
            # update the teacher normalization for the first 100 batches.
            if self.training and self.teacher_norm.num_batches_tracked < 100:
                self.teacher_norm.train()
            else:
                self.teacher_norm.eval()
            teacher_output = self.teacher_norm(teacher_output)
            
        student_output = self.student(images)
        
        if self.training:
            # during training, augment the image for the autoencoder and apply
            # the teacher and student to the augmented image also.
            images_for_ae = self.autoencoder_transform(images)
            with torch.no_grad():
                teacher_output_for_ae = self.teacher(images_for_ae)
                # do not update teacher normalization on augmented
                self.teacher_norm.eval()
                teacher_output_for_ae = self.teacher_norm(teacher_output_for_ae)
            student_output_for_ae = self.student(images_for_ae)
        else:
            # during inference, no augmentation so directly use teacher and student output
            images_for_ae = images
            teacher_output_for_ae = teacher_output
            student_output_for_ae = student_output
        autoencoder_output = self.autoencoder(images_for_ae)
        
        # split the student output
        student_output = student_output[:, :self._out_channels]
        student_output_for_ae = student_output_for_ae[:, self._out_channels:]
        
        # compute losses for training, anomaly maps for inference
        distance_st = (teacher_output - student_output) ** 2
        distance_stae = (autoencoder_output - student_output_for_ae)**2
        if self.training:
            # student loss
            d_hard = torch.quantile(distance_st, q=0.999)
            loss_hard = torch.mean(distance_st[distance_st >= d_hard])
            
            # autoencoder loss
            distance_ae = (teacher_output_for_ae - autoencoder_output) ** 2
            loss_ae = torch.mean(distance_ae)
            
            # autoencoder student loss
            loss_stae = torch.mean(distance_stae)
            
            # return losses
            output = {'loss_student_hard': loss_hard,
                      'loss_autoencoder': loss_ae,
                      'loss_autoencoder_student': loss_stae}
        else:
            # compute anomaly map of student and anomaly map of autoencoder
            maps_st = torch.mean(distance_st, dim=1, keepdim=True)
            maps_ae = torch.mean(distance_stae, dim=1, keepdim=True)
            
            # merge maps
            anomaly_maps = 0.5 * maps_st + 0.5 * maps_ae
            anomaly_maps = torch.nn.functional.pad(anomaly_maps, (4, 4, 4, 4))
            anomaly_scores = torch.amax(anomaly_maps, dim=(1, 2, 3))
            output = anomaly_maps, anomaly_scores
        return output
```

src/anomalib/models/efficientad/torch_model.py
```diff
@@ -0,0 +1,350 @@
...
+            loss_stae = torch.mean(distance_stae)
+
+            return (loss_st, loss_ae, loss_stae)
+
+        else:
+            distance_stae = torch.pow(ae_output - student_output_ae, 2)
+
+            map_st = torch.mean(distance_st, dim=1, keepdim=True)
+            map_stae = torch.mean(distance_stae, dim=1, keepdim=True)
+
```
nelson1425, line 338: For accurate segmentation, maps need to be padded again after this, see https://github.com/nelson1425/EfficientAD/blob/a44786ccddd3697b54903394ee1361f38e65de16/efficientad.py#L204

src/anomalib/models/efficientad/config.yaml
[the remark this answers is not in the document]
nelson1425: Yes, what is the most common default value of the other trainable methods of anomalib? Probably makes sense to use consistent default values.

rximg:
Hi @alexriedel1, I have run @nelson1425's version and noticed that the Lhard loss in my version is decreasing more slowly than in nelson1425's. I have checked for differences between the two versions, but have not found any obvious differences. I suspect that the problem may be occurring during the process of distilling the student network.

alexriedel1:
> 



> Hi @alexriedel1, I have run @nelson1425's version and noticed that the Lhard loss in my version is decreasing more slowly than in nelson1425's. I have checked for differences between the two versions, but have not found any obvious differences. I suspect that the problem may be occurring during the process of distilling the student network.

In this pull request I have added both models and yes, the difference is caused by the different pre trained students 
image

src/anomalib/models/efficientad/torch_model.py
alexriedel1: Oh right great point! I'm trying to find another fix for this problem though

src/anomalib/models/efficientad/torch_model.py
alexriedel1: Alright it's now added

src/anomalib/models/efficientad/torch_model.py
alexriedel1: I moved the imagenet normalization to the model definition

rximg:
Hi @alexriedel1, I have fixed some bugs in gradient updating and model evaluation status, and updated the results of MVTec AD which closely match the official ones.https://github.com/rximg/EfficientAD/commit/03ecb3ef100b5186594b2ee4ce5c684ce9457291#diff-7a26d02d1b7f4bc9d8fee30e821584c68b76e21f0f0adec99b69c82f009f5c82

src/anomalib/models/efficientad/torch_model.py
nelson1425: When reproducing the paper the padding of the output was necessary for the segmentation metrics to be the same. Without the padding of the output the segmentation metrics were very bad. I think it is because if we do not pad the output then the resizing of the anomaly map will displace some of the anomaly scores. Similar to when we run an edge filter with a filter size of 11x11 across a 100x100 image. This will output a 90x90 image. If we resize the 90x90 image to 100x100 again then the edge detections will be displaced. Instead we need to pad the image with 5 pixels on each border.

samet-akcay:
Hi @alexriedel1, is this ready for review? Cheers!

alexriedel1:
> Hi @alexriedel1, is this ready for review? Cheers!
I'm doing some last performance tests and will tell you when it's done :)


One thing that bothers me is that EfficientAD needs to be trained with ImageNet images. I created a subset of ImageNet (more than enough to train EfficientAD) and uploaded it to my Google Drive. Thats why anomalib needs gdown as a package. 

Do you know any way to upload this Imagenet subset to a safer and more consistent place than my personla Goole Drive?

samet-akcay:
> Do you know any way to upload this Imagenet subset to a safer and more consistent place than my personal Google Drive? It's roughly 3GB

One possibility would be to publish it in the Releases similar to [this](https://github.com/openvinotoolkit/anomalib/releases/tag/rkde-weights). There could be some problems though
- File size limit is 2GB ([link](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases#storage-and-bandwidth-quotas))
- License could also be an issue.

alexriedel1:
> > Do you know any way to upload this Imagenet subset to a safer and more consistent place than my personal Google Drive? It's roughly 3GB
> 
> One possibility would be to publish it in the Releases similar to [this](https://github.com/openvinotoolkit/anomalib/releases/tag/rkde-weights). There could be some problems though
> 
> * File size limit is 2GB ([link](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases#storage-and-bandwidth-quotas))
> * ImageNet License could also be an issue.

OK I see. I guess reducing the imagenet suibset to 2GB would be possible, about the license I'm not sure how to deal with it. 

**Anyway, the commit can be reviewed now.** 

I didn't run benchmarks for all MVTec AD categories though. 
On the bottle class, the following results are obtained using the pre-trained model by @rximg 
https://wandb.ai/alexried/mvtec/reports/image_AUROC-image_F1Score-pixel_AUROC-pixel_F1Score-23-05-12-17-12-27---Vmlldzo0MzQ0MzI0

The authors state that padding doesn't affect the performance but only speeds up training. However I realized that using a padded pre-trained model and removing padding in the actual EfficientAD training causes a drop in pixel F1 score. 

As you can see in the training log, the model converges relatively fast and shows good performance. Maybe training for 70,000 epochs (which is 325 epochs on MVTec bottle) isn't even necessary. 



The imagenet-pretrained teacher model is crucial for the performance. Using @rximg pre-trained model gave the best performance compared to some models trained by me using albumentation instead of torchvision transforms..

samet-akcay:
Would it be an idea to use fastai's Imagenette dataset, which is a subsample of the original ImageNet dataset.
https://github.com/fastai/imagenette

There is a direct [link](https://s3.amazonaws.com/fast-ai-imageclas/imagenette2.tgz) in the readme to download the dataset.

alexriedel1:
> Contributor

Thanks @samet-akcay, imagenette works just as well :)
https://api.wandb.ai/links/alexried/v5k5ws5o

So, please have a look over the code :)

src/anomalib/models/efficientad/pre_trained/teacher_medium_nelson1425.pth
djdameln, line 1: How did you obtain these weights? Did you train them or yourself or retrieve from a public source? If the latter, is there any licensing involved?

src/anomalib/models/efficientad/lightning_model.py
```diff
@@ -0,0 +1,248 @@
...
+        pre_trained_dir: str,
+        image_size: list,
+        model_size: str = "M",
+        lr: float = 0.0001,
+        weight_decay: float = 0.00001,
+        padding: bool = False,
+    ) -> None:
+        super().__init__()
+
+        self.category = category_name
```
djdameln, line 63: This attribute does not seem to be accessed anywhere. Could it be removed?

src/anomalib/models/efficientad/lightning_model.py
```diff
@@ -0,0 +1,248 @@
...
+                A.ToGray(p=0.3),  # converting it to gray scale with a probability of 0.3
+                A.CenterCrop(image_size[0], image_size[1]),  # and cropping the center 256 × 256 pixels
+                A.ToFloat(always_apply=False, p=1.0, max_value=255),
+                ToTensorV2(),
+            ]
+        )
+
+        self.prepare_imagenet_data()
+        imagenet_dataset = ImageFolder(self.imagenet_dir, transform=TransformsWrapper(t=self.data_transforms_imagenet))
+        self.imagenet_loader = DataLoader(imagenet_dataset, batch_size=1, shuffle=True, pin_memory=True)
```
djdameln, line 86: would hard-coding the batch size here lead to problems when Anomalib's training batch size is set to >1?

src/anomalib/models/efficientad/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""EfficientAD: Accurate Visual Anomaly Detection at Millisecond-Level Latencies.
+https://arxiv.org/pdf/2303.14535.pdf
+"""
+
+# Copyright (C) 2022 Intel Corporation
```
ashwinvaidya17, line 5:
```suggestion
# Copyright (C) 2023 Intel Corporation
```

src/anomalib/models/efficientad/torch_model.py
```diff
@@ -0,0 +1,314 @@
...
+        teacher_path (Path): path of pre-trained teacher model.
+        teacher_out_channels (int): number of convolution output channels of the pre-trained teacher model
+        model_size (str): size of teacher and student model
+    """
+
+    def __init__(
+        self,
+        teacher_path: Path,
+        teacher_out_channels: int,
+        input_size: list,
```
samet-akcay, line 192: Is this annotation correct? `input_size: list,`

src/anomalib/models/efficientad/torch_model.py
```diff
@@ -0,0 +1,314 @@
...
+
+# Copyright (C) 2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import logging
+import random
+from pathlib import Path
+from typing import Union
```
samet-akcay, line 11:
```suggestion

```

src/anomalib/models/efficientad/torch_model.py
```diff
@@ -0,0 +1,314 @@
...
+        teacher_path: Path,
+        teacher_out_channels: int,
+        input_size: list,
+        model_size="M",
+        padding=False,
+    ) -> None:
+        super().__init__()
+
+        self.teacher: Union[PDN_M, PDN_S]
+        self.student: Union[PDN_M, PDN_S]
```
samet-akcay, line 199:
```suggestion
        self.teacher: PDN_M | PDN_S
        self.student: PDN_M | PDN_S
```

src/anomalib/models/efficientad/lightning_model.py
```diff
@@ -0,0 +1,248 @@
...
+    """PL Lightning Module for the EfficientAD algorithm."""
+
+    def __init__(
+        self,
+        category_name: str,
+        teacher_file_name: str,
+        teacher_out_channels: int,
+        pre_trained_dir: str,
+        image_size: list,
+        model_size: str = "M",
```
ashwinvaidya17, line 56: Should we make this an Enum? It will be easier to see the available options then.

samet-akcay (commented):
@alexriedel1, this is great contribution! I only have some minor comments here and there and a question:

Do you think this model would be ONNX friendly since the model graph might need to contain some image transforms using albumentations?

src/anomalib/models/efficientad/lightning_model.py
```diff
@@ -0,0 +1,248 @@
...
+            distance = (y - channel_mean) ** 2
+            means_distance.append(torch.mean(distance, dim=[0, 2, 3]))
+
+        channel_var = torch.mean(torch.stack(means_distance), dim=0)[None, :, None, None]
+        channel_std = torch.sqrt(channel_var)
+        return {"mean": channel_mean, "std": channel_std}
+
+    @torch.no_grad()
+    def map_norm_quantiles(self, dataloader: DataLoader) -> dict[str, Tensor]:
+        """Calculate 90% andf 99.5% quantiles of the student and autoencoder feature maps.
```
ashwinvaidya17, line 130:
```suggestion
        """Calculate 90% and 99.5% quantiles of the student and autoencoder feature maps.
```

src/anomalib/models/efficientad/lightning_model.py
```diff
@@ -0,0 +1,248 @@
...
+          Dictionary containing anomaly maps.
+        """
+        del args, kwargs  # These variables are not used.
+
+        batch["anomaly_maps"] = self.model(batch["image"])["anomaly_map_combined"]
+
+        return batch
+
+
+class EfficientadLightning(EfficientAD):
```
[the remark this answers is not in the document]
ashwinvaidya17: We might have to change how we load the lightning model in our `models/__init__.py`

src/anomalib/models/efficientad/torch_model.py
```diff
@@ -0,0 +1,314 @@
...
+import torch.nn.functional as F
+from torch import Tensor, nn
+from torchvision import transforms
+
+logger = logging.getLogger(__name__)
+
+
+def imagenet_norm_batch(x):
+    mean = torch.tensor([0.485, 0.456, 0.406])[None, :, None, None].cuda()
+    std = torch.tensor([0.229, 0.224, 0.225])[None, :, None, None].cuda()
```
ashwinvaidya17, line 23: I'd be in favour of using `to(x.device)` as some users might want to run it on the CPU

ashwinvaidya17 (changes requested):
Thanks for the efforts! Very clean code. I have some minor comments
We should add this model to the changelog as well. I am in favour of add the weights to a release and downloading them when required. I am concerned about the size of the git tree growing later on if the binaries change.

lorenzolat1:
> @alexriedel1, this is great contribution! I only have some minor comments here and there and a question:
> 
> Do you think this model would be ONNX friendly since the model graph might need to contain some image transforms using albumentations?

Hi! I tried to generate an ONNX model from this implementation of EfficientAD in anomalib and, actually, the answer is no. I don't know if it's an opset type issue, but the inference process (with CUDA provider) requires 0.5 seconds to process an image (clearly, something is wrong, the model should be very light and fast). Looking at the model's graph I can't see anything strange, the model is "simple" (teacher + student + autoencoder) and operations too (convolutions, subs, resizes).

src/anomalib/models/efficientad/pre_trained/teacher_medium_nelson1425.pth
alexriedel1: The model weights come from this repo: https://github.com/nelson1425/EfficientAD
and from this repo https://github.com/rximg/EfficientAD/tree/main

alexriedel1:
> > @alexriedel1, this is great contribution! I only have some minor comments here and there and a question:
> > Do you think this model would be ONNX friendly since the model graph might need to contain some image transforms using albumentations?
> 
> Hi! I tried to generate an ONNX model from this implementation of EfficientAD in anomalib and, actually, the answer is no. I don't know if it's an opset type issue, but the inference process (with CUDA provider) requires 0.5 seconds to process an image (clearly, something is wrong, the model should be very light and fast). Looking at the model's graph I can't see anything strange, the model is "simple" (teacher + student + autoencoder) and operations too (convolutions, subs, resizes).

What timing do you expect on your machine? @samet-akcay the model graph doesn't actually need any transforms different from the other models pre-processing pipelines at inference time.

alexriedel1:
One thing that would be really valuable would be if someone is able to run all mvtec classes for a model benchmark.

lorenzolat1:
> > > @alexriedel1, this is great contribution! I only have some minor comments here and there and a question:
> > > Do you think this model would be ONNX friendly since the model graph might need to contain some image transforms using albumentations?
> > 
> > 
> > Hi! I tried to generate an ONNX model from this implementation of EfficientAD in anomalib and, actually, the answer is no. I don't know if it's an opset type issue, but the inference process (with CUDA provider) requires 0.5 seconds to process an image (clearly, something is wrong, the model should be very light and fast). Looking at the model's graph I can't see anything strange, the model is "simple" (teacher + student + autoencoder) and operations too (convolutions, subs, resizes).
> 
> What timing do you expect on your machine? @samet-akcay the model graph doesn't actually need any transforms different from the other models pre-processing pipelines at inference time.

Solved, there was a problem with the kind of algorithm for the convolution's operation in the CUDAProvider for OnnxRuntime, setting "cudnn_conv_algo_search" to "DEFAULT" in the provider options solved the timing problem. Using a GTX 1060 I got 88ms of inference time for an image, which is a reasonable timing. It's not as the 5.3ms for the A100 as the paper but it's still a good result. I think the onnx model works well!

alexriedel1:
> > > > @alexriedel1, this is great contribution! I only have some minor comments here and there and a question:
> > > > Do you think this model would be ONNX friendly since the model graph might need to contain some image transforms using albumentations?
> > > 
> > > 
> > > Hi! I tried to generate an ONNX model from this implementation of EfficientAD in anomalib and, actually, the answer is no. I don't know if it's an opset type issue, but the inference process (with CUDA provider) requires 0.5 seconds to process an image (clearly, something is wrong, the model should be very light and fast). Looking at the model's graph I can't see anything strange, the model is "simple" (teacher + student + autoencoder) and operations too (convolutions, subs, resizes).
> > 
> > 
> > What timing do you expect on your machine? @samet-akcay the model graph doesn't actually need any transforms different from the other models pre-processing pipelines at inference time.
> 
> Solved, there was a problem with the kind of algorithm for the convolution's operation in the CUDAProvider for OnnxRuntime, setting "cudnn_conv_algo_search" to "DEFAULT" in the provider options solved the timing problem. Using a GTX 1060 I got 88ms of inference time for an image, which is a reasonable timing. It's not as the 5.3ms for the A100 as the paper but it's still a good result. I think the onnx model works well!

Great! Do you also get reasonable anomaly detection Performance results?

lorenzolat1:
> > > > > @alexriedel1, this is great contribution! I only have some minor comments here and there and a question:
> > > > > Do you think this model would be ONNX friendly since the model graph might need to contain some image transforms using albumentations?
> > > > 
> > > > 
> > > > Hi! I tried to generate an ONNX model from this implementation of EfficientAD in anomalib and, actually, the answer is no. I don't know if it's an opset type issue, but the inference process (with CUDA provider) requires 0.5 seconds to process an image (clearly, something is wrong, the model should be very light and fast). Looking at the model's graph I can't see anything strange, the model is "simple" (teacher + student + autoencoder) and operations too (convolutions, subs, resizes).
> > > 
> > > 
> > > What timing do you expect on your machine? @samet-akcay the model graph doesn't actually need any transforms different from the other models pre-processing pipelines at inference time.
> > 
> > 
> > Solved, there was a problem with the kind of algorithm for the convolution's operation in the CUDAProvider for OnnxRuntime, setting "cudnn_conv_algo_search" to "DEFAULT" in the provider options solved the timing problem. Using a GTX 1060 I got 88ms of inference time for an image, which is a reasonable timing. It's not as the 5.3ms for the A100 as the paper but it's still a good result. I think the onnx model works well!
> 
> Great! Do you also get reasonable anomaly detection Performance results?

Unfortunately I used a custom dataset, therefore I haven't a global reference of the anomaly detection's performance for the model. Compared to other models I got good results with my dataset in anomaly detection, I will test the MVTec's dataset.

alexriedel1:
> > > > > > @alexriedel1, this is great contribution! I only have some minor comments here and there and a question:
> > > > > > Do you think this model would be ONNX friendly since the model graph might need to contain some image transforms using albumentations?
> > > > > 
> > > > > 
> > > > > Hi! I tried to generate an ONNX model from this implementation of EfficientAD in anomalib and, actually, the answer is no. I don't know if it's an opset type issue, but the inference process (with CUDA provider) requires 0.5 seconds to process an image (clearly, something is wrong, the model should be very light and fast). Looking at the model's graph I can't see anything strange, the model is "simple" (teacher + student + autoencoder) and operations too (convolutions, subs, resizes).
> > > > 
> > > > 
> > > > What timing do you expect on your machine? @samet-akcay the model graph doesn't actually need any transforms different from the other models pre-processing pipelines at inference time.
> > > 
> > > 
> > > Solved, there was a problem with the kind of algorithm for the convolution's operation in the CUDAProvider for OnnxRuntime, setting "cudnn_conv_algo_search" to "DEFAULT" in the provider options solved the timing problem. Using a GTX 1060 I got 88ms of inference time for an image, which is a reasonable timing. It's not as the 5.3ms for the A100 as the paper but it's still a good result. I think the onnx model works well!
> > 
> > 
> > Great! Do you also get reasonable anomaly detection Performance results?
> 
> Unfortunately I used a custom dataset, therefore I haven't a global reference of the anomaly detection's performance for the model. Compared to other models I got good results with my dataset in anomaly detection, I will test the MVTec's dataset.

Ok great, just wanted to make sure that nothing is lost during onnx conversion and the models works in principle

paularamo:
I have tried this using Notebooks section, I got an error with 001_getting_started notebook. 
image

`---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[14], line 5
      2 config.optimization.export_mode = "openvino"
      4 # Get the model and callbacks
----> 5 model = get_model(config)
      6 callbacks = get_callbacks(config)

File ~\Intel\tests\anomalib_effad_env\lib\site-packages\anomalib\models\__init__.py:106, in get_model(config)
    103     model = getattr(module, f"{_snake_to_pascal_case(config.model.name)}Lightning")(config)
    105 else:
--> 106     raise ValueError(f"Unknown model {config.model.name}!")
    108 if "init_weights" in config.keys() and config.init_weights:
    109     model.load_state_dict(load(os.path.join(config.project.path, config.init_weights))["state_dict"], strict=False)

ValueError: Unknown model efficientad!`

src/anomalib/models/efficientad/lightning_model.py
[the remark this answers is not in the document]
samet-akcay: yeah, true. Anway, we'll deprecate and remove `EfficientADLightning`, hopefully, in v1.0. We could address that then.

src/anomalib/models/efficientad/pre_trained/teacher_medium_nelson1425.pth
samet-akcay, line 1: @alexriedel1, would it be possible to choose one of the weight pairs? How will users know which weights pair they use? (either rximg or nelson)

samet-akcay: Apart from this comment, I think I'm happy with the overall shape/quality of the PR

alexriedel1: Yes! I want to train another teacher model myself in the next week using albumentation transforms hoping it will further increase EfficientADs accuracy. After that I will push the model with the highest accuracy to this PR

alexriedel1:
> I have tried this using Notebooks section and also OpenVINO model. I got an error with 001_getting_started notebook. image
> 
> `--------------------------------------------------------------------------- ValueError Traceback (most recent call last) Cell In[14], line 5 2 config.optimization.export_mode = "openvino" 4 # Get the model and callbacks ----> 5 model = get_model(config) 6 callbacks = get_callbacks(config)
> 
> File ~\Intel\tests\anomalib_effad_env\lib\site-packages\anomalib\models__init__.py:106, in get_model(config) 103 model = getattr(module, f"{_snake_to_pascal_case(config.model.name)}Lightning")(config) 105 else: --> 106 raise ValueError(f"Unknown model {config.model.name}!") 108 if "init_weights" in config.keys() and config.init_weights: 109 model.load_state_dict(load(os.path.join(config.project.path, config.init_weights))["state_dict"], strict=False)
> 
> ValueError: Unknown model efficientad!`

Guess you didn't pull from this PR? I have no issues loading the model

src/anomalib/models/efficientad/pre_trained/teacher_medium_nelson1425.pth
samet-akcay: Thanks! I guess we will need to wait for this to merge this PR, right?

alexriedel1: yes please

paularamo:
> > I have tried this using Notebooks section and also OpenVINO model. I got an error with 001_getting_started notebook. image
> > `--------------------------------------------------------------------------- ValueError Traceback (most recent call last) Cell In[14], line 5 2 config.optimization.export_mode = "openvino" 4 # Get the model and callbacks ----> 5 model = get_model(config) 6 callbacks = get_callbacks(config)
> > File ~\Intel\tests\anomalib_effad_env\lib\site-packages\anomalib\models__init__.py:106, in get_model(config) 103 model = getattr(module, f"{_snake_to_pascal_case(config.model.name)}Lightning")(config) 105 else: --> 106 raise ValueError(f"Unknown model {config.model.name}!") 108 if "init_weights" in config.keys() and config.init_weights: 109 model.load_state_dict(load(os.path.join(config.project.path, config.init_weights))["state_dict"], strict=False)
> > ValueError: Unknown model efficientad!`
> 
> Guess you didn't pull the right branch? I have no issues loading the model

Yes, It is working installing from your source files. Thanks

alexriedel1:
I have trained a teacher model using albumentation transform and it's giving even better results than the previous model. I added the model weights and removed the old ones. Everything should be in place now and ready for merging. What about adding the model to tests etc.?

nelson1425:
@alexriedel1 very well done thank you for your work! Can you enter the Mvtec AD results in the tables in https://github.com/openvinotoolkit/anomalib/blob/main/README.md ?

alexriedel1:
> @alexriedel1 very well done thank you for your work! Can you enter the Mvtec AD results in the tables in https://github.com/openvinotoolkit/anomalib/blob/main/README.md ?

Hey, I dont have the ressources to run the anomalib implementation on all mvtec AD classes so I cannot add the results for this implementation of EfficientAD. If there is someone here who could do a benchmark (https://github.com/openvinotoolkit/anomalib/tree/main/tools/benchmarking) with EfficientAD that would be great

samet-akcay:
> > @alexriedel1 very well done thank you for your work! Can you enter the Mvtec AD results in the tables in https://github.com/openvinotoolkit/anomalib/blob/main/README.md ?
> 
> Hey, I dont have the ressources to run the anomalib implementation on all mvtec AD classes so I cannot add the results for this implementation of EfficientAD. If there is someone here who could do a benchmark (https://github.com/openvinotoolkit/anomalib/tree/main/tools/benchmarking) with EfficientAD that would be great

I'll try to run a benchmark this week

src/anomalib/models/efficientad/pre_trained/pretrained_teacher_medium.pth
samet-akcay, line 1: What would be the best action to take here? Should we keep them here or release them as assets in the releases section? @alexriedel1, @djdameln, @ashwinvaidya17 ?

ashwinvaidya17: My main concern in that the git tree might become too big. I am in favour of releasing them as assets. We can add a readme with acknowledgement and license. Also, release will give us a permalink that we can refer to in the code.

alexriedel1: Sure, I will add the download function once the weights are released as assets

samet-akcay: Here is the link to the weights
https://github.com/openvinotoolkit/anomalib/releases/tag/efficientad_pretrained_weights

@alexriedel1, can you please double check if they work ok? Let me know if there is something wrong with the link. Thanks!

ashwinvaidya17 (approved):
Thanks!

alexriedel1:
Im having a hard time getting the precommit hooks prettier and markdownlint to run on my windows machine ☹️ I will give it another try though today

samet-akcay:
> Im having a hard time getting the precommit hooks prettier and markdownlint to run on my windows machine ☹️ I will give it another try though today

We had this feedback from someone else too. Wasn't aware that pre-commit wasn't working well on a windows machine. Or maybe our configuration is problematic. We might need to investigate.

Meanwhile I could run the checks on your PR

djdameln (approved):
Thanks a lot for this contribution!

watertianyi:
@alexriedel1 @samet-akcay 
I saw that you used the https://github.com/nelson1425/EfficientAD code to improve it
What I want to know is:
1. Did you use pretraining.py in nelson1425 to train the teacher weights?
2. Then the training code in anomalib is equivalent to the efficientad.py training code in nelson1425, right?
3. If I use the teacher weights trained by pretraining.py in nelson1425, can they be used in anomalib?
4. I only used 200 positive samples of my own data to train efficientad in anomalib. The classification effect was image_AUROC 70% and image_F1Score 60%.
If segmentation training is used, pixel_AUROC is 70% and pixel_F1Score is 30%. Is there any way to improve the above?
5. Also, when the above classification model is tested, the defect intensity of the heat map of the abnormal data is not obvious or the performance is not obvious. Is there any way to improve it?
6. Although the segmentation model can reflect the heat map defects of abnormal data, it is often segmented incorrectly?

PR #1080, opened 2023-05-09
Modify codecov upload

ashwinvaidya17:
# Description

- Removed the `-Z` flag as it fails the CI if codecov upload fails
- Don't pass token if it is not passed to the action.


samet-akcay (approved):
Thanks for addressing this

PR #1087, opened 2023-05-11
🚚 [Custom Loops] Thresholding refactor

ashwinvaidya17:
# Description

- Refactors thresholding in custom loops.
- [ ] TODO: Need to make similar changes to other configs if this design is approved


src/anomalib/models/padim/config.yaml
```diff
@@ -36,9 +36,12 @@ model:
 
 post_processing:
   normalization_method: min_max # <null, min_max, cdf>
-  threshold_method: adaptive # options: [adaptive, manual]
-  manual_image_threshold: null
-  manual_pixel_threshold: null
+  image_threshold:
+    class_path: AdaptiveScoreThreshold # see anomalib.utils.metrics.thresholding for more options
+    init_args: null
```
samet-akcay, line 41: when there is no `init_args` to pass, do we still need to pass it as `null`. What is the behaviour when we just do the following:

```
  image_threshold:
    class_path: AdaptiveScoreThreshold # see anomalib.utils.metrics.thresholding for more options
```

src/anomalib/trainer/trainer.py
```diff
@@ -32,18 +32,16 @@ class AnomalibTrainer(Trainer):
         Refer to PyTorch Lightning's Trainer for a list of parameters for details on other Trainer parameters.
 
     Args:
-        threshold_method (ThresholdMethod): Thresholding method for normalizer.
+        imag_threshold_method (dict): Thresholding method. If None, adaptive thresholding is used.
```
samet-akcay, line 35:
```suggestion
        image_threshold_method (dict): Thresholding method. If None, adaptive thresholding is used.
```

src/anomalib/trainer/utils/thresholder.py
```diff
@@ -23,37 +21,18 @@ class Thresholder:
 
     Args:
         trainer (trainer.AnomalibTrainer): Trainer object
-        threshold_method (ThresholdMethod): Thresholding method to use. Defaults to ``ThresholdMethod.ADAPTIVE``.
-        manual_image_threshold (Optional[float]): Image threshold in case manual threshold is used. Defaults to None.
-        manual_pixel_threshold (Optional[float]) = Pixel threshold in case manual threshold is used. Defaults to None.
+        imag_threshold_method (BaseAnomalyScoreThreshold): Thresholding method. If None, adaptive thresholding is used.
```
samet-akcay, line 24:
```suggestion
        image_threshold_method (BaseAnomalyScoreThreshold): Thresholding method. If None, adaptive thresholding is used.
```

src/anomalib/trainer/utils/thresholder.py
```diff
@@ -86,17 +73,45 @@ def update(self, outputs: STEP_OUTPUT) -> None:
...
+        if (
+            self.trainer.task_type != TaskType.CLASSIFICATION
+            and "anomaly_maps" in outputs.keys()
+            and "mask" in outputs.keys()
+        ):
             pixel_metric.cpu()
+            # TODO this should use bounding boxes for detection task type
             pixel_metric.update(outputs["anomaly_maps"], outputs["mask"].int())
+
+    def _get_threshold_method(self, threshold_method: dict | None) -> BaseAnomalyScoreThreshold:
```
samet-akcay, line 89: how about naming this `def _get_thresholder`? It is a bit confusing to have the same name as the arg? Thoughts?

src/anomalib/utils/metrics/thresholding/adaptive_score_threshold.py
```diff
@@ -11,16 +11,14 @@
 from torch import Tensor
 from torchmetrics import PrecisionRecallCurve
 
+from .base import BaseAnomalyScoreThreshold
 
-class AnomalyScoreThreshold(PrecisionRecallCurve):
-    """Anomaly Score Threshold.
 
+class AdaptiveScoreThreshold(PrecisionRecallCurve, BaseAnomalyScoreThreshold):
```
samet-akcay, line 17: Is this method also called as `AdaptiveScoreThreshold` in the Mvtec paper?
What if we had another method that adaptively computes the threshold?

src/anomalib/utils/metrics/thresholding/base.py
```diff
@@ -0,0 +1,33 @@
...
+
+class BaseAnomalyScoreThreshold(Metric, ABC):
+    """Base class for thresholding metrics."""
+
+    def __init__(self, **kwargs) -> None:
+        super().__init__(**kwargs)
+
+    def compute(self) -> Any:
+        """Compute the threshold
+        Returns:
```
samet-akcay, line 22:
```suggestion
        
        Returns:
```

src/anomalib/utils/metrics/thresholding/base.py
```diff
@@ -0,0 +1,33 @@
...
+            Value of the optimal threshold.
+        """
+        raise NotImplementedError("Subclass of BaseAnomalyScoreThreshold must implement the compute method")
+
+    def update(self, *args, **kwargs) -> None:
+        """Update the metric state
+        Args:
```
samet-akcay, line 29:
```suggestion
        
        Args:
```

src/anomalib/utils/metrics/thresholding/manual_threshold.py
```diff
@@ -0,0 +1,36 @@
...
+
+import torch
+
+from .base import BaseAnomalyScoreThreshold
+
+
+class ManualThreshold(BaseAnomalyScoreThreshold):
+    def __init__(self, default_value: float = 0.5, **kwargs) -> None:
+        """Initialize Manual Threshold.
+        Args:
```
samet-akcay, line 16:
```suggestion
        
        Args:
```

src/anomalib/utils/metrics/thresholding/manual_threshold.py
```diff
@@ -0,0 +1,36 @@
...
+            default_value (float, optional): Default threshold value. Defaults to 0.5.
+        """
+        super().__init__(**kwargs)
+        self.add_state("value", default=torch.tensor(default_value), persistent=True)
+        self.value = torch.tensor(default_value)
+
+    def compute(self) -> torch.Tensor:
+        """Compute the threshold.
+        In case of manual thresholding, the threshold is already set and does not need to be computed.
+        Returns:
```
samet-akcay, line 26:
```suggestion
        
        Returns:
```

samet-akcay, line 26:
```suggestion
        
        Returns:
```

src/anomalib/utils/metrics/thresholding/manual_threshold.py
```diff
@@ -0,0 +1,36 @@
...
+        Args:
+            default_value (float, optional): Default threshold value. Defaults to 0.5.
+        """
+        super().__init__(**kwargs)
+        self.add_state("value", default=torch.tensor(default_value), persistent=True)
+        self.value = torch.tensor(default_value)
+
+    def compute(self) -> torch.Tensor:
+        """Compute the threshold.
+        In case of manual thresholding, the threshold is already set and does not need to be computed.
```
samet-akcay, line 25:
```suggestion
        
        In case of manual thresholding, the threshold is already set and does not need to be computed.
```

src/anomalib/utils/metrics/thresholding/manual_threshold.py
```diff
@@ -0,0 +1,36 @@
...
+        Returns:
+            torch.Tensor: Value of the optimal threshold.
+        """
+        return self.value
+
+    def update(self, *args, **kwargs) -> None:
+        """Do nothing.
+        Args:
```
samet-akcay, line 33:
```suggestion
        
        Args:
```

samet-akcay (changes requested):
It looks good to me overall. It would be good to instantiate these components via a class path. 

I only have few minor comments.

src/anomalib/utils/metrics/thresholding/adaptive_score_threshold.py
ashwinvaidya17: They don't have this but yeah the naming is quite broad. How about F1AdaptiveThreshold?

src/anomalib/utils/metrics/thresholding/base.py
ashwinvaidya17: I thought we had a pre-commit hook to check docstrings.

samet-akcay: yeah, weird. It should've picked it up

src/anomalib/utils/metrics/thresholding/adaptive_score_threshold.py
samet-akcay: @djdameln, would be good to get your thoughts here too

src/anomalib/trainer/utils/thresholder.py
djdameln: I agree that the current naming is confusing, but I feel that `_get_thresholder` is also confusing, because the class itself is also called `Thresholder`. Not sure what would be the best option.

djdameln: Maybe `_get_threshold_metric`. After all, the returned instance is a subclass of `Metric`

src/anomalib/trainer/utils/thresholder.py
```diff
@@ -86,17 +73,45 @@ def update(self, outputs: STEP_OUTPUT) -> None:
...
+        """Get threshold method.
+
+        Args:
+            threshold_method (dict | None): Threshold method. Defaults to AdaptiveScoreThreshold.
+
+        Returns:
+            Instantiated threshold method.
+        """
+        thresholder: BaseAnomalyScoreThreshold
```
djdameln, line 98: Following the naming discussion above, this variable could be called `threshold_metric`

src/anomalib/utils/metrics/thresholding/adaptive_score_threshold.py
djdameln: I would prefer `F1Adaptivethreshold`

samet-akcay: Alright let's go with `F1Adaptivethreshold` then

src/anomalib/trainer/utils/thresholder.py
```diff
@@ -86,17 +70,45 @@ def update(self, outputs: STEP_OUTPUT) -> None:
...
             pixel_metric.update(outputs["anomaly_maps"], outputs["mask"].int())
+
+    def _get_threshold_metric(self, threshold_method: dict | None) -> BaseAnomalyScoreThreshold:
+        """Get threshold method.
+
+        Args:
+            threshold_method (dict | None): Threshold method. Defaults to F1AdaptiveThreshold.
+
+        Returns:
+            Instantiated threshold method.
```
samet-akcay, line 93:
```suggestion
            Instantiated threshold metric.
```

src/anomalib/trainer/utils/thresholder.py
```diff
@@ -86,17 +70,45 @@ def update(self, outputs: STEP_OUTPUT) -> None:
...
+            self.trainer.task_type != TaskType.CLASSIFICATION
+            and "anomaly_maps" in outputs.keys()
+            and "mask" in outputs.keys()
+        ):
             pixel_metric.cpu()
+            # TODO this should use bounding boxes for detection task type
             pixel_metric.update(outputs["anomaly_maps"], outputs["mask"].int())
+
+    def _get_threshold_metric(self, threshold_method: dict | None) -> BaseAnomalyScoreThreshold:
+        """Get threshold method.
```
samet-akcay, line 87:
```suggestion
        """Get instantiated threshold metric.
```

src/anomalib/utils/metrics/thresholding/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""Thresholding metrics"""
+
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+from .base import BaseAnomalyScoreThreshold
```
samet-akcay, line 6: would it be an idea to follow a similar naming convention to `F1AdaptiveThreshold` and `ManualThreshold` and name the base class as `BaseAnomalyScoreThreshold`? @djdameln @ashwinvaidya17 ?

samet-akcay (commented):
Final bits...

src/anomalib/trainer/trainer.py
```diff
@@ -32,18 +32,16 @@ class AnomalibTrainer(Trainer):
...
+        pixel_threshold_method (dict): Thresholding method. If None, adaptive thresholding is used.
         normalization_method (NormalizationMethod): Normalization method
-        manual_image_threshold (float | None): If threshold method is manual, this needs to be set. Defaults to None.
-        manual_pixel_threshold (float | None): If threshold method is manual, this needs to be set. Defaults to None.
     """
 
     def __init__(
         self,
-        threshold_method: ThresholdMethod = ThresholdMethod.ADAPTIVE,
+        image_threshold: dict | None = None,  # TODO change from dict to BaseAnomalyScoreThreshold in CLI
```
samet-akcay, line 42:
```suggestion
        image_threshold: dict | None = None,  # TODO change from dict to BaseAnomalyThreshold in CLI
```

src/anomalib/trainer/utils/thresholder.py
```diff
@@ -23,48 +21,34 @@ class Thresholder:
 
     Args:
         trainer (trainer.AnomalibTrainer): Trainer object
-        threshold_method (ThresholdMethod): Thresholding method to use. Defaults to ``ThresholdMethod.ADAPTIVE``.
-        manual_image_threshold (Optional[float]): Image threshold in case manual threshold is used. Defaults to None.
-        manual_pixel_threshold (Optional[float]) = Pixel threshold in case manual threshold is used. Defaults to None.
+        image_threshold_method (BaseAnomalyScoreThreshold): Thresholding method. If None, adaptive thresholding is used.
+        pixel_threshold_method (BaseAnomalyScoreThreshold): Thresholding method. If None, adaptive thresholding is used.
```
samet-akcay, line 25:
```suggestion
        image_threshold_method (BaseAnomalyThreshold): Thresholding method. If None, adaptive thresholding is used.
        pixel_threshold_method (BaseAnomalyThreshold): Thresholding method. If None, adaptive thresholding is used.
```

src/anomalib/trainer/utils/thresholder.py
```diff
@@ -86,17 +70,45 @@ def update(self, outputs: STEP_OUTPUT) -> None:
...
-            self._update_thresholds(self.trainer.image_threshold, self.trainer.pixel_threshold, outputs)
-
-    @staticmethod
-    def _update_thresholds(
-        image_metric: AnomalyScoreThreshold,
-        pixel_metric: AnomalyScoreThreshold,
-        outputs: STEP_OUTPUT,
-    ) -> None:
+        image_metric = self.trainer.image_threshold
+        pixel_metric = self.trainer.pixel_threshold
```
samet-akcay, line 74: can you remind me why we assign threshold to metric variable?

samet-akcay (commented):
thanks, final comments hopefully:)

src/anomalib/trainer/utils/thresholder.py
ashwinvaidya17: Yeah I think image_threshold and pixel_threshold might be better names. I guess initially the `_update_thresholds` method also computed the metric values.

samet-akcay (approved):

src/anomalib/trainer/utils/thresholder.py
```diff
@@ -85,17 +69,45 @@ def update(self, outputs: STEP_OUTPUT) -> None:
...
+                    threshold_module = import_module("anomalib.utils.metrics.thresholding")
+                    _threshold_class = getattr(threshold_module, _class_path)
+            except (AttributeError, ModuleNotFoundError) as exception:
+                raise ValueError(f"Threshold class {_class_path} not found") from exception
+
+            init_args = threshold_method.get("init_args")
+            init_args = init_args if init_args is not None else {}
+            threshold_metric = _threshold_class(**init_args)
+
+        return threshold_metric
```
djdameln, line 113: I'm guessing this will change with the new CLI? Then I'll ignore this part

djdameln (changes requested):

src/anomalib/trainer/utils/thresholder.py
ashwinvaidya17: Yeah. I'll add TODO to this class

src/anomalib/trainer/connectors/thresholding_connector.py
```diff
@@ -22,48 +20,29 @@ class ThresholdingConnector:
...
     def initialize(self) -> None:
-        """Assigns pixel and image thresholds to the model.
-
-        This allows us to export the metrics along with the torch model.
-        """
-
-        if self.threshold_method == ThresholdMethod.MANUAL:
-            self.trainer.pixel_threshold.value = torch.tensor(self.manual_pixel_threshold).cpu()
-            self.trainer.image_threshold.value = torch.tensor(self.manual_image_threshold).cpu()
+        """Assigns pixel and image thresholds to the Anomalib trainer."""
```
djdameln, line 40: Is this docstring still accurate? Thresholds are no longer assigned to trainer right?

djdameln (commented):
I wonder if the design could be simplified further. For example, what if we apply thresholding before normalization? In the validation loop, we could return the threshold value from ThresholdConnector.compute` in the validation loop, and pass it to `MetricsConnector.set_threshold`. This would have the following advantages:
- By explicitly passing the threshold values, the code becomes easier to follow.
- We only need to set the threshold once after validation, and when loading the model from file.
- We would not have to store the threshold value in the trainer, leading to a cleaner class implementation.
- The logic in `MetricsConnector.set_threshold` would be simplified.
- The value of the threshold would always be in the same (un-normalized) domain, making debugging easier.

Of course, we would have to confirm that everything still works as expected. We would also have to make some changes to the `MetricsConnector`. Specifically, the predictions would have to be collected after each step instead of at the end of the epoch.

ashwinvaidya17:
> I wonder if the design could be simplified further. For example, what if we apply thresholding before normalization? In the validation loop, we could return the threshold value from `ThresholdConnector.compute` in the validation loop, and pass it to `MetricsConnector.set_threshold`. This would have the following advantages:
> 
> * By explicitly passing the threshold values, the code becomes easier to follow.
> * We only need to set the threshold once after validation, and when loading the model from file.
> * We would not have to store the threshold value in the trainer, leading to a cleaner class implementation.
> * The logic in `MetricsConnector.set_threshold` would be simplified.
> * The value of the threshold would always be in the same (un-normalized) domain, making debugging easier.
> 
> Of course, we would have to confirm that everything still works as expected. We would also have to make some changes to the `MetricsConnector`. Specifically, the predictions would have to be collected after each step instead of at the end of the epoch.

The problem is that threshold is used by `post_processing_connector` and `normalizer`. Since it is accessed by a lot of modules, I thought it made sense to make it a part of the trainer instead of accessing it through `self.trainer.threshold_connector.image_threshold`. Basically remove one layer of access.

ashwinvaidya17:
Closing as we are moving to fabric

PR #1088, opened 2023-05-11
Fix mypy pep561

### LINKED ISSUE #1015: [Task]: Make anomalib PEP 561 compliant for mypy

WenjingKangIntel:
# Description

With this PR, when running 
`python setup.py install`, `python setup.py install sdist` or `python setup.py sdist bdist_wheel`, `py.typed` will be included in the build package.
Snipaste_2023-05-11_18-38-04

- Fixes #1015

## Changes

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -107,6 +107,7 @@ setup(
     packages=find_packages(where="src", include=["anomalib", "anomalib.*"]),
     install_requires=INSTALL_REQUIRES,
     extras_require=EXTRAS_REQUIRE,
+    include_package_data=True,
     package_data={"": ["config.yaml"]},
     entry_points={"console_scripts": ["anomalib=anomalib.utils.cli.cli:main"]},
 )


WenjingKangIntel:
@ashwinvaidya17

ashwinvaidya17:
Do we really need `sdist` and  `bdist_wheel` in the package?

WenjingKangIntel:
> Do we really need `sdist` and `bdist_wheel` in the package?

Without this pr, even pure `python setup.py install` will not include `py.typed` in the package. FYI.

WenjingKangIntel:
From this link:
https://setuptools.pypa.io/en/latest/userguide/datafiles.html#summary
include_package_data
Accept all data files and directories matched by [MANIFEST.in](https://packaging.python.org/en/latest/guides/using-manifest-in/) or added by a [plugin](https://setuptools.pypa.io/en/latest/userguide/extension.html#adding-support-for-revision-control-systems).

include_package_data needs to explicitly set to True.

setup.py
```diff
@@ -107,6 +107,7 @@ def get_required_packages(requirement_files: list[str]) -> list[str]:
     packages=find_packages(where="src", include=["anomalib", "anomalib.*"]),
     install_requires=INSTALL_REQUIRES,
     extras_require=EXTRAS_REQUIRE,
+    include_package_data=True,
     package_data={"": ["config.yaml"]},
```
ashwinvaidya17, line 111: From what I tested
```package_data={"": ["config.yaml"], "anomalib": ["py.typed"]},```
should be sufficient

ashwinvaidya17 (changes requested):

setup.py
WenjingKangIntel: Yes, I've tested it as well. It worked. But if we are going to use `MANIFEST.in`, we need to use the include_package_data=True.
If we are going to use `package_data={"": ["config.yaml"], "anomalib": ["py.typed"]},`, then we need to remove `MANIFEST.in`, because they have duplicated functionality, which is to include the same file.

ashwinvaidya17 (approved):

samet-akcay (approved):

PR #1118, opened 2023-06-09
Fix/load model

### LINKED ISSUE #1076: [Bug]: Version change causes torch_inference error
### Describe the bug

Hello @ashwinvaidya17 and @samet-akcay, I noticed that ```./tools/inference/torch_inference.py``` appears changed in the two versions as shown.
2971683342964_ pic

When configuring the __parser__```[ parser=ArgumentParser() ]```, the version of @ashwinvaidya17 is passed into config, while @samet-akcay, the latest version (and the one I'm using now), is not passed into config。
3011683349959_ pic

And there is a difference in the corresponding function ```load_model``` that loads the model

2991683343203_ pic
----
3001683343302_ pic

Because I am using the latest version of anomalib (samet-akcay)，When I use pytorch inference, it comes up with an error. The specific place where the error occurred I have located to be 
```
model = torch.load(path, map_location=self.device)["model"]
```

Its specific error message is:
```KeyError: 'model'```

So I used the following code to debug：
1CFF5722BAF8A8F9AB3C49903EC5289B

I found that this is indeed a mistake, there is no "model"

----
So I have the following questions:

1: Is this ```KeyError: 'model'``` problem a bug that does exist, as I see it specifically mentioned in the update file (#600, #601). How should I go about fixing it. 

2: I'm actually encapsulating my own program to use OnnxRuntime to implement the reasoning. I noticed that openvino's confidence score is calculated after getting a value from ```predictions=predictions[self.output_blob]``` . Do you have any suggestions on how I should get this similar result in OnnxRuntime by its inference?

I am looking forward to your reply, thank you very much!

### Dataset

Folder

### Model

PatchCore

torch_inference.py

### OS information

pytorch_1.12.1
CUDA_11.3

### Expected behavior

-

### Screenshots

_No response_

### Pip/GitHub

pip

_No response_

### Configuration YAML

```yaml
-
```

### Logs

```shell
-
```

### Code of Conduct

samet-akcay:
# Description

- Handle the exception if the `load_model` cannot load the torch weights.

- Fixes #1076 

## Changes

## Checklist


src/anomalib/deploy/inferencers/torch_inferencer.py
```diff
@@ -82,9 +82,16 @@ def load_model(self, path: str | Path) -> nn.Module:
...
 
-        model = torch.load(path, map_location=self.device)["model"]
-        model.eval()
-        return model.to(self.device)
+        try:
+            model = torch.load(path, map_location=self.device)["model"]
+            model.eval()
+            return model.to(self.device)
+        except Exception as e:
+            raise Exception(
```
djdameln, line 90: As suggested by Codacy, should we raise (and maybe catch) a more specific exception here?

PR #1146, opened 2023-06-21
squeeze output map

alexriedel1:
# Description

It can happen for some models, that the output anomaly map can have an extra dimension [8,1,256,256] whereas the ground truth mask has not. This causes an error.
To fix this, the squeeze function is added before calculating pixel metrix.

Also this PR makes a change on how to read augmentation config from the config file to make image size HPO sweeps more accesible. 

This PR uses an ImportWarning instead of an ImportError for missing OpenVINO install. Otherwise without OpenVINO installed anomalib will crash.

## Changes

## Checklist

- [x ] My code follows the [pre-commit style and check guidelines](https://github.com/openvinotoolkit/anomalib/blob/main/CONTRIBUTING.md) of this project.
- [x ] I have performed a self-review of my code
- [x ] I have commented my code, particularly in hard-to-understand areas


src/anomalib/models/efficientad/torch_model.py
```diff
@@ -123,7 +123,6 @@ class Decoder(nn.Module):
     def __init__(self, out_channels, padding, img_size, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         self.img_size = img_size
-        self.last_upsample = 64 if padding else 56
```
djdameln, line 126: Was this commited intentionally?

alexriedel1: yes this line is not used as the same variable is (correctly) defined one line after..

samet-akcay (approved):
Thanks @alexriedel1

PR #1147, opened 2023-06-21
OpenVINO inference redesign

ashwinvaidya17:
# Description

- Bbox detection using batch operations. For loops are removed as their ranges become constant when exported to onnx.
## Known limitations
- Currently, torch and openvino inferencers support only single image in a batch.
- Since connected_components is added to the onnx graph, loading the model takes quite some time.
- Code is a bit messy but I'll wait for the discussions before starting the refactor.

> _**Note:** Introduces breaking changes_


src/anomalib/deploy/model.py
```diff
@@ -0,0 +1,277 @@
...
+            pred_mask.device
+        )  # N-1 x 2 x H x W
+        masks_ = masks_.view(labels.size(0), 2, -1)  # N-1 x 2 x H*W
+        # # assign bounding boxes
+        bboxes[:, 2] = torch.max(masks_[:, 1], dim=1)[0]
+        bboxes[:, 3] = torch.max(masks_[:, 0], dim=1)[0]
+        masks_[masks_ == 0] = H + 1
+        bboxes[:, 0] = torch.min(masks_[:, 1], dim=1)[0]
+        bboxes[:, 1] = torch.min(masks_[:, 0], dim=1)[0]
+        return bboxes, boxes_per_image
```
djdameln, line 148: instead of returning `boxes_per_image`, maybe you could prepend the image idx to the box coordinates, like
```python
indices = torch.repeat_interleave(torch.arange(B), boxes_per_image)
torch.cat([indices.unsqueeze(1), bboxes], dim=1)
```
This leads to better interpretable outputs and might make further processing easier.
(Just a suggestion, your call)

djdameln (commented):
Do you think we should add the `batch_masks_to_boxes` method to `anomalib.data.utils.boxes`? This would keep the implementation of `ExportModel` a bit cleaner. We could potentially have it replace the other `masks_to_boxes` calls throughout the code if it's faster.

src/anomalib/deploy/model.py
```diff
@@ -0,0 +1,277 @@
...
+        """
+        height, width = pred_mask.shape[-2:]
+        pred_mask = pred_mask.view((-1, 1, height, width)).float()  # reshape to (B, 1, H, W) and cast to float
+
+        batch_comps = connected_components(pred_mask, num_iterations=1000).squeeze(1).int()
+        B, H, W = batch_comps.shape
+
+        # split the boxes later on based on this
+        # since torch.count_nonzero is not supported in onnx, we use torch.sum with non equality
+        boxes_per_image = torch.sum(torch.unique(batch_comps.view(B, -1), dim=1) != 0, dim=1) - 1
```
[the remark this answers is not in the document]
ashwinvaidya17: What was your input tensor?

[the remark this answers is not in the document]
djdameln: I ran a padim model with detection task type and replaced `masks_to_boxes` in the predict step ([here](https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/models/components/base/anomaly_module.py#L96)) with your new batched method.

src/anomalib/deploy/model.py
ashwinvaidya17: I get this error when I try to use `repeat_interleave`
```bash
TypeError: z_(): incompatible function arguments. The following argument types are supported:
    1. (self: torch._C.Node, arg0: str, arg1: torch.Tensor) -> torch._C.Node

Invoked with: %25415 : Tensor = onnx::Constant(), scope: __main__.Dummy::
```
I think this is because the inputs to `repeat_interleave` have a variable dimension.

djdameln: That's weird, `repeat_interleave` is also used in RKDE implementation, which does export to onnx without problems

ashwinvaidya17: When I explicitly add the dimenstion `indices.repeat_interleave(boxes_per_image, 0)` onnx export works but OpenVINO IR conversion fails.
```bash
RuntimeError: Check 'unknown_operators.empty()' failed at src/frontends/onnx/frontend/src/core/graph.cpp:213:
OpenVINO does not support the following ONNX operations: ConcatFromSequence, SequenceEmpty, SplitToSequence
```
But looks like the support is in progress https://github.com/openvinotoolkit/openvino/issues/15094

djdameln: Again, OpenVINO conversion works fine for RKDE. So it's a bit surprising that it doesn't work here.

samet-akcay (commented):
This was my review before you added your recent changes. I'll go for another round.

ashwinvaidya17:
Closing this as OpenVINO inference will be addressed later

PR #1152, opened 2023-06-26
🚜 Address bandit issues

samet-akcay:
# Description

Bandit scan raised certain security vulnerabilities. To address this, the PR adds the following changes:
- Address the security vulnerabilities
- Add `bandit` as a pre-commit hook
- Autoupdate the pre-commit hooks' versions.

## Changes

## Checklist


src/anomalib/utils/callbacks/nncf/callback.py
```diff
@@ -94,4 +94,5 @@ def on_train_end(self, trainer: pl.Trainer, pl_module: pl.LightningModule) -> No
         onnx_path = os.path.join(self.export_dir, "model_nncf.onnx")
         self.nncf_ctrl.export_model(onnx_path)
         optimize_command = "mo --input_model " + onnx_path + " --output_dir " + self.export_dir
-        os.system(optimize_command)
+        # this is a subprocess call to run openvino, so it is safe to use os.system
+        os.system(optimize_command)  # nosec
```
yunchu, line 98: If you want to exclude a line from the bandit scanning, the test ID that would be excluded from the line should be followed `# nosec` comment. e.g. # nosec B605
And it's not related with the change but AFAIK, os.system() could be replaced to use subprocess.call() and it's more clear way to use sub-process since the return of the os.system() is OS dependent. (https://docs.python.org/3/library/subprocess.html#replacing-os-system)

Issue #1177, opened 2023-07-13
[Bug]:  efficientad training its own dataset reports an error


laogonggong847:
### Describe the bug

I encountered an error while training my own dataset using EfficientAD. I only made modifications to the dataset section of the configuration file provided by the official EfficientAD repository. Based on the same modifications, I was able to train models like CFA and PatchCore successfully, but I encountered an error specifically when training EfficientAD.

The yaml file for my efficientad is as follows (I only changed the ```dataset``` section, the rest is consistent)

```
dataset:
  name: mydata
  format: folder
  path: ./MyDataset/HC_ZT_ROI
  normal_dir: normal #  name of the folder containing normal images.
  abnormal_dir: abnormal #  name of the folder containing abnormal images.
  normal_test_dir: null #  name of the folder containing normal test images.
  task: classification
  mask: null
  extensions: null
  train_batch_size: 32
  test_batch_size: 32
  num_workers: 0
  image_size: 500 # dimensions to which images are resized (mandatory)
  center_crop: null # dimensions to which images are center-cropped after resizing (optional)
  normalization: null # data distribution to which the images will be normalized: [none, imagenet]
  transform_config:
    train: null
    eval: null
  test_split_mode: from_dir # options: [from_dir, synthetic]
  test_split_ratio: 0.2 # fraction of train images held out testing (usage depends on test_split_mode)
  val_split_mode: same_as_test # options: [same_as_test, from_test, synthetic]
  val_split_ratio: 0.5 # fraction of train/test images held out for validation (usage depends on val_split_mode)
  tiling:
    apply: false
    tile_size: null
    stride: null
    remove_border_count: 0
    use_random_tiling: False
    random_tile_count: 16
```
----
----

___I have tentatively determined that the cause of my error is the parameter ```"normalization"```___

When I when I set ```normalization: imagenet``` or ```normalization: none``` the exact error message is:
```
2023-07-13 13:56:58,938 - anomalib.models.efficientad.lightning_model - INFO - Load pretrained teacher model from pre_trained\efficientad_pretrained_weights\pretrained_teacher_small.pth
Traceback (most recent call last):
  File "E:/Code/Anomalib/0.6.0/anomalib-0-6-0/tools/train.py", line 82, in <module>
    train(args)
  File "E:/Code/Anomalib/0.6.0/anomalib-0-6-0/tools/train.py", line 59, in train
    model = get_model(config)
[...]

PR #1182, opened 2023-07-14
Minor fixes

alexriedel1:
# Description

It can happen for some models, that the output anomaly map can have an extra dimension [8,1,256,256] whereas the ground truth mask has not. This causes an error. To fix this, the squeeze function is added before calculating pixel metrix.

This PR uses an ImportWarning instead of an ImportError for missing OpenVINO install. Otherwise without OpenVINO installed anomalib will crash.

Also this PR makes a change on how to read augmentation config from the config file to make image size HPO sweeps more accesible.

Also the PR adresses https://github.com/openvinotoolkit/anomalib/issues/1177 and some other issues with the tensor shape for torch.quantile being too large in the efficientad torch model during training

- Fixes # (issue)

## Changes

## Checklist


src/anomalib/models/efficient_ad/torch_model.py
```diff
@@ -24,6 +24,16 @@ def imagenet_norm_batch(x):
     return x_norm
 
 
+def reduce_tensor_elems(tensor: torch.Tensor, device, max_input_size=2**24) -> torch.Tensor:
+    tensor = torch.flatten(tensor)
+    if len(tensor) > max_input_size:
+        # select a random subset with max_input_size elements.
+        perm = torch.randperm(len(tensor), device=device)
```
samet-akcay, line 31: As I mentioned below, instead of passing `device` as a function argument, we could get the device from the tensor itself. Something like 

```suggestion
        perm = torch.randperm(len(tensor), device=tensor.device)
```

samet-akcay (changes requested):
Thanks @alexriedel1, some minor comments only

blaz-r:
I have just one question. Does that extra dimension really cause issues for pixel metrics? I believe that I haven't encountered that, but it is possible that this is due to nature of ROC where entire tensor gets flattened anyway.

alexriedel1:
> I have just one question. Does that extra dimension really cause issues for pixel metrics? I believe that I haven't encountered that, but it is possible that this is due to nature of ROC where entire tensor gets flattened anyway.

I had some cases where the additional dimension in the heat maps caused an error.

alexriedel1:
> Thanks @alexriedel1, some minor comments only

done!

alexriedel1:
> > Thanks @alexriedel1, some minor comments only
> 
> done!

Codacy doesn't want regular formatted strings and doesn't want f-strings :((

alexriedel1:
ok now its done i guess :)

PR #1216, opened 2023-07-25
🚚 Refactor Benchmarking Script

ashwinvaidya17:
# Description

- While this started as an exercise to simplify the script, it probably has become more complicated. But the outputs look pretty now 🤷‍♂️
image

- All outputs are logged to `runs/benchmark.logs`.
- I've removed throughput computations as it needs to be rafactored and we don't use it anywhere.
- Also, I've tried to ensure that any exception does not break the benchmarking run so that the script keeps collecting results for the models that work.

> **Note:** wandb and comet uploads haven't been tested yet

## Changes

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,9 @@ target-version = "py38"
 # Allow imports relative to the "src" and "tests" directories.
 src = ["src", "tests"]
 
+[tool.ruff.isort]
+known-third-party = ["wandb", "comet_ml"]
+
 [tool.ruff.mccabe]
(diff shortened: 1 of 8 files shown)


djdameln (commented):
Do you think we could simplify the exception handling somehow? Catching the exceptions in various different places make it a bit hard to read the code. 

Also, I'm wondering if it would be possible to log the results to wandb as they are collected, instead of all at once at the end of the script. In some cases could be useful to see the results come in as they are generated.

samet-akcay (approved):

PR #1219, opened 2023-07-27
Add modelAPI related stuff

ashwinvaidya17:
# Description

- Adds support for the current version of modelAPI. This way the users can export the models from Anomalib and use it with cpp/python modelAPI wrappers. These are similar to the changes in OTX

### DIFF
--- a/src/anomalib/deploy/export.py
+++ b/src/anomalib/deploy/export.py
@@ -10,9 +10,11 @@ import subprocess  # nosec
 from enum import Enum
 from pathlib import Path
 from typing import Any
+from warnings import warn
 
 import numpy as np
 import torch
+from openvino.runtime import Core, serialize
 from torch import Tensor
 from torch.types import Number
 
@@ -113,7 +115,7 @@ def export(
         # Export model to onnx and convert to OpenVINO IR if export mode is set to OpenVINO.
         onnx_path = export_to_onnx(model, input_size, export_path)
         if export_mode == ExportMode.OPENVINO:
-            export_to_openvino(export_path, onnx_path)
+            export_to_openvino(export_path, onnx_path, metadata, input_size)
 
     else:
         raise ValueError(f"Unknown export mode {export_mode}")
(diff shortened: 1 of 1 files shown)


src/anomalib/deploy/export.py
```diff
@@ -156,12 +158,74 @@ def export_to_onnx(model: AnomalyModule, input_size: tuple[int, int], export_pat
...
+    _metadata[("model_info", "labels")] = ["Normal", "Anomaly"]
+    _metadata[("model_info", "image_shape")] = _serialize_list(input_size)
+
+    for k, data in _metadata.items():
+        model.set_rt_info(data, list(k))
+
+    tmp_xml_path = Path(Path(xml_file).parent) / "tmp.xml"
+    serialize(model, str(tmp_xml_path))
+    tmp_xml_path.rename(xml_file)
+    Path(str(tmp_xml_path.parent / tmp_xml_path.stem) + ".bin").unlink()
```
samet-akcay, line 227: can we have have something like this
```suggestion
    tmp_xml_path.with_suffix(".bin").unlink()
```

src/anomalib/deploy/export.py
```diff
@@ -156,12 +158,74 @@ def export_to_onnx(model: AnomalyModule, input_size: tuple[int, int], export_pat
...
+
+
+def _serialize_list(arr):
```
samet-akcay, line 230: would it be possible to annotate the variable and add docstring please

src/anomalib/deploy/export.py
```diff
@@ -113,7 +115,7 @@ def export(
         # Export model to onnx and convert to OpenVINO IR if export mode is set to OpenVINO.
         onnx_path = export_to_onnx(model, input_size, export_path)
         if export_mode == ExportMode.OPENVINO:
-            export_to_openvino(export_path, onnx_path)
+            export_to_openvino(export_path, onnx_path, metadata, input_size)
```
samet-akcay, line 118: would it be an idea to add `input_size` to `metadata`. Or would it be too much for a temporary solution?

ashwinvaidya17: I was thinking of adding it to metadata earlier but we don't need this in the onnx model and we might forget to remove it later when we refactor the openvino part.

src/anomalib/deploy/export.py
```diff
@@ -156,12 +158,82 @@ def export_to_onnx(model: AnomalyModule, input_size: tuple[int, int], export_pat
...
+    _metadata[("model_info", "labels")] = ["Normal", "Anomaly"]
+    _metadata[("model_info", "image_shape")] = _serialize_list(input_size)
+
+    for k, data in _metadata.items():
+        model.set_rt_info(data, list(k))
+
+    tmp_xml_path = Path(Path(xml_file).parent) / "tmp.xml"
+    serialize(model, str(tmp_xml_path))
+    tmp_xml_path.rename(xml_file)
+    tmp_xml_path.with_suffix(".bin").unlink()
```
djdameln, line 227: I think it would be good to add a comment explaining this section

src/anomalib/deploy/export.py
```diff
@@ -156,12 +158,83 @@ def export_to_onnx(model: AnomalyModule, input_size: tuple[int, int], export_pat
...
+
+    _metadata[("model_info", "reverse_input_channels")] = True
+    _metadata[("model_info", "model_type")] = "AnomalyDetection"
+    _metadata[("model_info", "labels")] = ["Normal", "Anomaly"]
+    _metadata[("model_info", "image_shape")] = _serialize_list(input_size)
+
+    for k, data in _metadata.items():
+        model.set_rt_info(data, list(k))
+
+    tmp_xml_path = Path(Path(xml_file).parent) / "tmp.xml"
```
samet-akcay, line 224:
```suggestion
    tmp_xml_path = Path(xml_file).parent / "tmp.xml"
```

samet-akcay (changes requested):
thanks, just a single line of change

djdameln (approved):

samet-akcay (approved):

PR #1247, opened 2023-08-09
metrics[0]: add per-image overlap (pimo) 

jpcbertoldo:
# Description

Previously in https://github.com/openvinotoolkit/anomalib/pull/1220

## Per-Image Overlap (PImO)

Adding the metric Per-Image Overlap (PImO) and its area under the curve (AUC).

At each given threshold:
- X-axis: in-image FPR average on normal images (equivalent to the set FPR of normal images).
- Y-axis: Overlap between the class 'anomalous' in the ground truth and the predicted masks (in-image TPR).

I choose the vocabulary 'overlap' insted of TPR because it echoes with the Per-Region Overlap (PRO).

---

## binclf curve

The per-image overlap (tpr) needs the binary classification matrix curve (binclf curve), which has to be computed _per image_ , while they must share the thresholds (Since the x-axis (FPR) is shared).

You'll find a weird-looking funciton using itertools to compute thos binclf curves, which i found to be much faster on CPU than the two alternatives in torchmetrics (the latter mentioned in the former's doc) --  image below shows some WIP comparisons; a proper version on the way.

image

However, there is a `tensor.cpu().numpy()` in there, and I only tested with tensors _already_ in the **C**PU. 
Perhaps when using tensors in the **G**PU that operation might make the fastest algorithm change?

- Fixes # (issue)

## Changes

## Checklist

- [X] My code follows the [pre-commit style and check guidelines](https://github.com/openvinotoolkit/anomalib/blob/main/CONTRIBUTING.md) of this project.


jpcbertoldo:
@samet-akcay could you rebase [openvinotoolkit:feature/pimo](https://github.com/openvinotoolkit/anomalib/tree/feature/pimo) to main and launch the workflow please?

jpcbertoldo:
image

there is an issue with the feature branch it seems, could you enable codacy?

jpcbertoldo:
image

This coverage issue it because I saved the notebook with the kernel name in my loca environment (`anomalib-dev-gsoc`). 
Can we ignore it for now and solve it in a later PR?

src/anomalib/utils/metrics/perimg/pimo.py
```diff
@@ -0,0 +1,211 @@
...
+              I.e. a metric of per-image average TPR.
+
+for shared fpr = mean( perimg fpr ) == set fpr
+    find the th = fpr^-1( MAX_FPR ) with a binary search on the pixels of the norm images
+    i.e. it's not necessary to compute the perimg fpr curves (tf. binclf curves) in advance
+for other shared fpr alternatives, it's necessary to compute the perimg fpr curves first anyway
+
+further: also choose the th upper bound to be the max score at normal pixels
+"""
+
```
samet-akcay, line 14: License header. You could do this later on as well, since this is going into the feature branch.

src/anomalib/utils/metrics/perimg/pimo.py
```diff
@@ -0,0 +1,211 @@
...
+    _validate_atleast_one_anomalous_image,
+    _validate_atleast_one_normal_image,
+)
+from .plot import (
+    plot_all_pimo_curves,
+)
+
+# =========================================== METRICS ===========================================
+
+PImOResult = namedtuple(
```
samet-akcay, line 37: do you have any specific motivation why use named tuple instead of data class. I was just curious :)  Just preference or any other motivation?

src/anomalib/utils/metrics/perimg/common.py
```diff
@@ -0,0 +1,114 @@
...
+    diffs = curve.diff()
+    diffs_valid = diffs if not nan_allowed else diffs[~torch.isnan(diffs)]
+
+    if (diffs_valid > 0).any():
+        raise ValueError(
+            "Expected argument `curve` to be monotonically decreasing, but got non-monotonically decreasing values."
+        )
+
+
+def _validate_perimg_rate_curves(curves: Tensor, nan_allowed: bool = False):
```
samet-akcay, line 40: matching snake_case convention
```suggestion
def _validate_per_img_rate_curves(curves: Tensor, nan_allowed: bool = False):
```

src/anomalib/utils/metrics/perimg/common.py
```diff
@@ -0,0 +1,114 @@
...
+
+    unique_values = torch.unique(image_classes)
+    if torch.any((unique_values != 0) & (unique_values != 1)):
+        raise ValueError(
+            "Expected argument `image_classes` to be a *binary* tensor with ground truth labels, "
+            f"but got a tensor with values {unique_values}."
+        )
+
+
+def _validate_atleast_one_anomalous_image(image_classes: Tensor):
```
samet-akcay, line 107: should we have this or keep it as is? Not sure
```suggestion
def _validate_at_least_one_anomalous_image(image_classes: Tensor):
```

src/anomalib/utils/metrics/perimg/common.py
```diff
@@ -0,0 +1,114 @@
...
+    if (image_classes == 1).sum() == 0:
+        raise ValueError("Expected argument at least one anomalous image, but found none.")
+
+
+def _validate_atleast_one_normal_image(image_classes: Tensor):
```
samet-akcay, line 112: same as above, not sure.
```suggestion
def _validate_at_least_one_normal_image(image_classes: Tensor):
```

samet-akcay (commented):
@jpcbertoldo, thanks for this! I just have minor comments, majority of which are naming-related.

src/anomalib/utils/metrics/__init__.py
```diff
@@ -20,9 +20,22 @@
 from .collection import AnomalibMetricCollection
 from .min_max import MinMax
 from .optimal_f1 import OptimalF1
+from .perimg import AUPImO, PerImageBinClfCurve, PImO
```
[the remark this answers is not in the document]
jpcbertoldo: I used `PerImageBinaryClassificationCurve` at first but it feels very wordy and some lines will become tough to read.
`binclf` is what I found in torchmetrics so i copied them.

`per_img` i think it's ok; put it in the list of refactors.

src/anomalib/utils/metrics/perimg/common.py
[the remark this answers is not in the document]
jpcbertoldo: > is `common.py` needed? Could these be in `__init__.py`. Or alternatively, can this be a `utils.py`?

In `__init__.py` it would create circular imports (in future PRs). 

`common.py` -> `utils.py` is fine IMO, but it will later have more things inside that are not really "utils". I'll add this to the refactors.

src/anomalib/utils/metrics/perimg/pimo.py
jpcbertoldo: Yes and more or less. 

It's possible to just unpack the things, like

```python
_, __, shared_fpr, tprs, image_classes = self.compute()
```.

Also it's just a tuple, I think it makes user code less dependent on in-`anomalib` classes. I tend to use it for this type of case where I actually just want to return many things but there are so many that it's easier to document it as a namedtuple. I noticed `scipy` does something similar (?).

Otherwise I'm a big fan of dataclass : )

File: src/anomalib/utils/metrics/perimg/common.py (hunk 3 above)
Comment (jpcbertoldo) [reply src/anomalib/utils/metrics/perimg/common.py:40]: refactor branch

File: src/anomalib/utils/metrics/perimg/common.py (hunk 4 above)
Comment (jpcbertoldo) [reply src/anomalib/utils/metrics/perimg/common.py:107]: refactor branch

File: src/anomalib/utils/metrics/perimg/common.py (hunk 5 above)
Comment (jpcbertoldo) [reply src/anomalib/utils/metrics/perimg/common.py:112]: refactor branch

**Comment by jpcbertoldo** [comment]
List of refactors before merging feature branch: 
https://github.com/jpcbertoldo/anomalib/blob/metrics/refactors/src/anomalib/utils/metrics/perimg/.refactors

Adds from this PR

```
1. rename `perimg` to `per_img`
  1.1 in functions
  1.2 the module
2. rename`common.py` to `utils.py`
3. add license header to files
4. remove `.ignore`
5. rename `atleast` to `at_least` (?)
```

src/anomalib/utils/metrics/perimg/common.py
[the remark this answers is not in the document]
jpcbertoldo: > > Can you add doctoring and other headers as well?
> 
> Sure. I'll add the docstrings, but what other headers specifically (apart from license)?

I reviewed and incremented module docstrings in a future PR.

https://github.com/jpcbertoldo/anomalib/pull/10/commits/621f58e22db21f156243fa88627fc611dfdc6069

jpcbertoldo:
@samet-akcay i put the naming issues in the todo list for the final refactor
i think we can merge this and move to the next PR?

Issue #1195, opened 2023-07-20
[Task]: Support list config for normal_dir and normal_test_dir of folder format


fanchuanster:
### What is the motivation for this task?

I am frustrated to merge multiple categories into a single normal dir (normal_test_dir as well), wondering it it makes sense to support list config for them?

dataset:
  name: model
  format: folder
  path: /app/data/trainData
  normal_dir: [folder1, folder2]
  normal_test_dir: [testfolder1, testfolder2]

this is the error it throws when it's a list config:

TypeError: expected str, bytes or os.PathLike object, not ListConfig

### Describe the solution you'd like

support list config for normal_dir and normal_test_dir of folder format

samet-akcay:
@djdameln, any thoughts?

harimkang:
I think it's a good feature for user convenience. I'll take a look and share my progress.

PR #1265, opened 2023-08-14
Add support for receiving folder dataset paths as a list

harimkang:
# Description

- Modified the Folder Dataset type to use the ListConfig type to accept paths as a List.
- Add a ListConfig to an existing type with an input of paths it can accept.

- PR for #1195 

```
# configs.yaml
dataset:
    name: model
    format: folder
    path: /app/data/trainData
    normal_dir: [folder1, folder2]
    normal_test_dir: [testfolder1, testfolder2]

# or we can also get the list format in the form below.
    normal_dir:
        - folder1
        - folder2
```

## Changes

## Checklist

- [X] My code follows the [pre-commit style and check guidelines](https://github.com/openvinotoolkit/anomalib/blob/main/CONTRIBUTING.md) of this project.

### DIFF
--- a/src/anomalib/data/folder.py
+++ b/src/anomalib/data/folder.py
@@ -8,7 +8,7 @@ This script creates a custom dataset from a folder.
 from __future__ import annotations
 
 from pathlib import Path
-from typing import List, Sequence
+from typing import Sequence
 
 import albumentations as A
 from pandas import DataFrame
@@ -54,22 +54,27 @@ def make_folder_dataset(
         DataFrame: an output dataframe containing samples for the requested split (ie., train or test)
     """
 
-    def _path_to_list_config_with_resolve_path(path: str | Path | Sequence[str | Path] | None) -> List[Path]:
-        """Function for changing path to List[Path].
+    def _resolve_path_and_convert_to_list(path: str | Path | Sequence[str | Path] | None) -> list[Path]:
+        """Convert path to list of paths.
         Args:
             path (str | Path | Sequence | None): Path to replace with Sequence[str | Path].
+        Examples:
+            >>> _path_to_list_config_with_resolve_path("dir")
+            [Path("path/to/dir")]
+            >>> _path_to_list_config_with_resolve_path(["dir1", "dir2"])
+            [Path("path/to/dir1"), Path("path/to/dir2")]
         Returns:
-            List[Path]: The result of path replaced by Sequence[str | Path].
+            list[Path]: The result of path replaced by Sequence[str | Path].
         """
         if isinstance(path, Sequence) and not isinstance(path, str):
             return [_resolve_path(dir_path, root) for dir_path in path]
         return [_resolve_path(path, root)] if path is not None else []
 
     # All paths are changed to the List[Path] type and used.
-    normal_dir = _path_to_list_config_with_resolve_path(normal_dir)
-    abnormal_dir = _path_to_list_config_with_resolve_path(abnormal_dir)
-    normal_test_dir = _path_to_list_config_with_resolve_path(normal_test_dir)
-    mask_dir = _path_to_list_config_with_resolve_path(mask_dir)
+    normal_dir = _resolve_path_and_convert_to_list(normal_dir)
+    abnormal_dir = _resolve_path_and_convert_to_list(abnormal_dir)
+    normal_test_dir = _resolve_path_and_convert_to_list(normal_test_dir)
+    mask_dir = _resolve_path_and_convert_to_list(mask_dir)
     assert len(normal_dir) > 0, "A folder location must be provided in normal_dir."
 
     filenames = []
--- a/src/anomalib/data/folder.py
+++ b/src/anomalib/data/folder.py
@@ -61,9 +61,9 @@ def make_folder_dataset(
         Returns:
             ListConfig: The result of path replaced by ListConfig.
         """
-        if not isinstance(path, ListConfig):
-            return ListConfig(content=[_resolve_path(path, root)] if path is not None else [])
-        return ListConfig(content=[_resolve_path(dir_path, root) for dir_path in path])
+        if isinstance(path, ListConfig):
+            return ListConfig(content=[_resolve_path(dir_path, root) for dir_path in path])
+        return ListConfig(content=[_resolve_path(path, root)] if path is not None else [])
 
     # All paths are changed to the ListConfig type and used.
     normal_dir = _path_to_list_config_with_resolve_path(normal_dir)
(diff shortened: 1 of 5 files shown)


harimkang:
Hi @djdameln, I want to write a unit test (and update documentation) for this, but I don't know what form it should take, can you give me some advice?

src/anomalib/data/folder.py
```diff
@@ -27,23 +28,23 @@
 
 
 def make_folder_dataset(
-    normal_dir: str | Path,
-    root: str | Path | None = None,
-    abnormal_dir: str | Path | None = None,
-    normal_test_dir: str | Path | None = None,
-    mask_dir: str | Path | None = None,
+    normal_dir: str | Path | ListConfig,
```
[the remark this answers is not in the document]
samet-akcay: I'm not sure if mypy would be happy with `List` though. Maybe we could try `Sequence[Path]`.

samet-akcay (changes requested):
Thanks a lot @harimkang. I only have a single comment regarding the type of the multiple paths.

src/anomalib/data/folder.py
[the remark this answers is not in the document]
harimkang: yes i agree with you. I'll modify it to the same type as Sequence[path].

djdameln:
> Hi @djdameln, I want to write a unit test (and update documentation) for this, but I don't know what form it should take, can you give me some advice?

For the datamodule tests you could have a look at this file: https://github.com/openvinotoolkit/anomalib/blob/main/tests/pre_merge/datasets/test_datamodule.py, where we test the full functionality of the different datamodules using several fixtures. You could use the fixtures to write a test where you instantiate the Folder datamodule using a list of paths. The test would look something like this:

```python
def test_folder_list_inputs(make_data_module):
    data_module = make_data_module("folder", abnormal_dir=["broken_large", "broken_small"])
    assert data_module
```

This would test if we can create a Folder datamodule using multiple abnormal directories instead of just one. Admittedly, creating the entire datamodule might add some unnecessary complexity, because we actually only need to test the `make_folder_dataset` function. We currently don't have unit tests specific for the `make_dataset` functions, but we could consider adding those, as it would allow us to test those functions directly and write much more specific tests. 

So, alternatively you could add a file called `test_make_dataset_functions` under `tests/pre_merge/datasets` and add a unit test for your newly added functionality there. We could then later append that file with other specific tests for the `make_dataset` functions.

For the documentation, you could add a short paragraph to this guide: https://github.com/openvinotoolkit/anomalib/blob/main/docs/source/how_to_guides/train_custom_data.rst, explaining that the `abnormal_dir` parameter also takes lists of directories, and provide an example with the hazelnut toy dataset (e.g. `abnormal_dir: [colour, crack]`).

src/anomalib/data/folder.py
```diff
@@ -52,11 +53,24 @@ def make_folder_dataset(
...
-    assert normal_dir.is_dir(), "A folder location must be provided in normal_dir."
+
+    def _path_to_list_config_with_resolve_path(path: str | Path | ListConfig | None) -> ListConfig:
+        """Function for changing path to ListConfig.
+        Args:
+            path (str | Path | ListConfig| None): Path to replace with ListConfig.
+        Returns:
+            ListConfig: The result of path replaced by ListConfig.
+        """
+        if isinstance(path, ListConfig):
```
djdameln, line 64: I think we should allow other data types here as well, since only supporting `ListConfig` is not very API-friendly. The following should probably work:

```python
>>> from anomalib.data import Folder
>>> data_module = Folder(
...     root="datasets/hazelnut_toy/",
...     normal_dir="good",
...     abnormal_dir=["colour", "crack"],
...     image_size=256,
... )
>>> data_module.setup()
```

djdameln (changes requested):

src/anomalib/data/folder.py
```diff
@@ -52,11 +53,24 @@ def make_folder_dataset(
     Returns:
         DataFrame: an output dataframe containing samples for the requested split (ie., train or test)
     """
-    normal_dir = _resolve_path(normal_dir, root)
-    abnormal_dir = _resolve_path(abnormal_dir, root) if abnormal_dir is not None else None
-    normal_test_dir = _resolve_path(normal_test_dir, root) if normal_test_dir is not None else None
-    mask_dir = _resolve_path(mask_dir, root) if mask_dir is not None else None
-    assert normal_dir.is_dir(), "A folder location must be provided in normal_dir."
+
+    def _path_to_list_config_with_resolve_path(path: str | Path | Sequence[str | Path] | None) -> List[Path]:
```
samet-akcay, line 57: How about a function something like: `_resolve_path_and_convert_to_list` or something similar. We don't convert to list_config anymore, so the current name might be confusing?

src/anomalib/data/folder.py
```diff
@@ -52,11 +53,24 @@ def make_folder_dataset(
...
+        """Function for changing path to List[Path].
+        Args:
+            path (str | Path | Sequence | None): Path to replace with Sequence[str | Path].
+        Returns:
+            List[Path]: The result of path replaced by Sequence[str | Path].
+        """
```
samet-akcay, line 63: I always find this sort of stuff easier to follow when there is example section in the docstring. How about adding one like the following:

```suggestion
        """Convert path to list of paths.

        Args:
            path (str | Path | Sequence | None): Path to replace with Sequence[str | Path].

        Examples:
            >>> _path_to_list_config_with_resolve_path("dir")
            [Path("path/to/dir")]

            >>> _path_to_list_config_with_resolve_path(["dir1", "dir2"])
            [Path("path/to/dir1"), Path("path/to/dir2")]

        Returns:
            list[Path]: The result of path replaced by Sequence[str | Path].
        """
```

src/anomalib/data/folder.py
samet-akcay, line 57:
```suggestion
    def _path_to_list_config_with_resolve_path(path: str | Path | Sequence[str | Path] | None) -> list[Path]:
```

samet-akcay (changes requested):
Thanks a lot @harimkang! Much better! Only a single comment left from my side.

src/anomalib/data/folder.py
harimkang: Thank you for the comment. I have fixed this. (to Sequence)

src/anomalib/data/folder.py
harimkang: Thank you for your suggestion. I have renamed this.

samet-akcay (approved):
Thanks @harimkang!

djdameln (approved):

PR #1300, opened 2023-08-24
🚜 Refactor padim and patchcore models

samet-akcay:
## Description

The aim of this PR was to properly handle the cases where there is no validation but test sets. However, it turns out that we might need additional changes, which could be handled by the custom trainer that will come with v1.0. I still wanted to push these changes thinking that these might improve the existing Padim Patchcore torch and lightning models. 

## Changes

<details>

- [ ] Documentation update

</details>

## Checklist

<details>

</details>


ashwinvaidya17 (approved):
Looks good to me

djdameln (changes requested):
I think we still need to perform the `is_fitted` checks in both `on_train_epoch_end` and `on_save_checkpoint`. Changing the terminology of `val_split_mode` and `test_split_mode` may help to prevent users from attempting a training with incompatible parameter configurations (such as `val_split_mode: none` and `threshold_method: adaptive`), but it won't address the original issue that the model cannot be exported when there is no validation set.

Here's a summary why I think we need to perform the check in each of these methods:

`on_validation_start`

We need the check here for training with a validation set, because Lightning performs the validation sequence within the training epoch (so before on_train_epoch_end). Without the check validation will be attempted with an unfitted model.

`on_save_checkpoint`

When training without a validation set, we need the check here to ensure that the model is fitted before exporting it.


`on_train_epoch_end`

When training without validation set, and immediately deploying the trained model afterwards without storing and loading to/from the file system. This will be mainly for API users, since our entrypoint scripts always re-load the model from the file system before evaluation/inference.

---

In addition, I think we do need to consider adding this to the base `AnomalyModule`. I'm mainly concerned that third party model implementers may not be aware that these checks are needed for membank-based models. Of course, we could (and probably should!) add a test case in the model tests for this, but it would be even better if we could abstract this from the concrete implementation altogether so that the implementer does not need to care about it at all. Adding this to the base class might require some design though, because the checks are only relevant for membank/feature-extraction based models.

---

Finally, in addition to [Padim, PatchCore], [AI-VAD, DFKDE, DFM, EfficientAD, RKDE] also fit some model components at the end of training, and thus may also need these checks. (Actually, this is about 50% of all models in Anomalib, which would warrant adding it to the base class).

src/anomalib/models/components/base/memory_bank_module.py
```diff
@@ -0,0 +1,38 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+
+from abc import abstractmethod
+
+import torch
+from torch import nn
+
+
+class MemoryBankMixin(nn.Module):
```
ashwinvaidya17, line 13: I think the reason this is not working that we probably need to inherit from `LightningModule` instead of `nn.Module`. But in that case I am not sure if this approach might work as we might have to implement configure_optimizers, and all those necessary methods.

PR #1307, opened 2023-08-28
🩹 Fix docs build

ashwinvaidya17:
# Description

- Fix docs build
- Example (https://github.com/ashwinvaidya17/anomalib/actions/runs/6001349893/job/16275342469)

Need to change settings though
image

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.10
+          python-version: "3.10"
 
       - name: Install requirements
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -8,66 +8,61 @@ on:
...
-            echo "Branch gh-pages already exists"
-          fi
-      - name: Commit docs to gh-pages branch
+      - name: Clean directory
         run: |
-          git fetch
-          git checkout gh-pages
           mkdir -p /tmp/docs_build
-          cp -r docs/build/html/* /tmp/docs_build/
+          cp -r docs/build/html/* /tmp/docs_build
```
samet-akcay, line 51: Can you explain this change?

ashwinvaidya17: Probably it was not present when I copied this command. I guess it should not affect copying the folder

samet-akcay: As far as I know `/` keeps the directory structure, while the latter just flattens the structure. If it works, I'm fine with it though

samet-akcay (approved):

PR #1311, opened 2023-08-30
[Custom Trainer] Refactor thresholding

ashwinvaidya17:
## Description

- Refactor thresholding to a callback
- Allows loading thresholding class


src/anomalib/utils/callbacks/thresholding.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def _outputs_to_cpu(self, output: STEP_OUTPUT):
+        if isinstance(output, dict):
+            for key, value in output.items():
+                output[key] = self._outputs_to_cpu(value)
+        elif isinstance(output, Tensor):
+            output = output.cpu()
+        return output
+
+    def _initialize_thresholds(
```
samet-akcay, line 93: can you logically order these private methods as well? For example, `_initalize_thresholds` could be the first one, right?

src/anomalib/utils/callbacks/thresholding.py
```diff
@@ -0,0 +1,159 @@
...
+        elif isinstance(threshold, (str, DictConfig, ListConfig)):
+            self._load_from_config(threshold)
+        else:
+            raise ValueError(f"Invalid threshold type {type(threshold)}")
+
+    def _load_from_config(self, threshold: DictConfig | str | ListConfig) -> None:
+        """Loads the thresholding class based on the config.
+
+        Example:
+            threshold: F1AdaptiveThreshold
```
samet-akcay, line 116: is it possible to do:

```yaml
threshold: adaptive
```

src/anomalib/utils/callbacks/thresholding.py
```diff
@@ -0,0 +1,159 @@
...
+            self.image_threshold = self._get_threshold_from_config(threshold)
+            self.pixel_threshold = self.image_threshold.clone()
+        elif isinstance(threshold, ListConfig):
+            self.image_threshold = self._get_threshold_from_config(threshold[0])
+            self.pixel_threshold = self._get_threshold_from_config(threshold[1])
+        else:
+            raise ValueError(f"Invalid threshold config {threshold}")
+
+    def _get_threshold_from_config(self, threshold: DictConfig | str) -> BaseAnomalyThreshold:
+        if isinstance(threshold, str):
```
samet-akcay, line 143: This function seems quite important. 

Can you add an `Example section here for a future reference? Like,

```python
def _get_threshold_from_config(self, threshold: DictConfig | str) -> BaseAnomalyThreshold:
    """
    Example:
         >>> _get_threshold_from_config(threshold="adaptive")
         ...
    """
```

src/anomalib/utils/metrics/thresholding/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""Thresholding metrics"""
+
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+from .base import BaseAnomalyThreshold
+from .f1adaptive_threshold import F1AdaptiveThreshold
```
samet-akcay, line 7: Which one would look better, `f1adaptive_threshold` or `f1_adaptive_threshold`?

src/anomalib/utils/sweep/helpers/callbacks.py
```diff
@@ -7,7 +7,7 @@
 from lightning.pytorch import Callback
 from omegaconf import DictConfig, ListConfig
 
-from anomalib.utils.callbacks import MetricsConfigurationCallback, PostProcessingConfigurationCallback
+from anomalib.utils.callbacks import MetricsConfigurationCallback
```
samet-akcay, line 10: Are we going to keep this callback?

samet-akcay (changes requested):
Thanks @ashwinvaidya17! Looking much better! Quite clean!

I have some minor comments here and there that we could discuss

src/anomalib/models/components/base/anomaly_module.py
```diff
@@ -38,9 +37,8 @@ def __init__(self) -> None:
         self.loss: nn.Module
         self.callbacks: list[Callback]
 
-        self.threshold_method: ThresholdMethod
-        self.image_threshold = AnomalyScoreThreshold().cpu()
-        self.pixel_threshold = AnomalyScoreThreshold().cpu()
+        self.image_threshold: BaseAnomalyThreshold
+        self.pixel_threshold: BaseAnomalyThreshold
```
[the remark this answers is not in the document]
ashwinvaidya17: Yeah that makes sense. `BaseAnomalyThreshold` is quite verbose. How about the last option. `BaseThreshold`

src/anomalib/models/components/base/anomaly_module.py
```diff
@@ -113,17 +108,11 @@ def on_test_batch_end(self, outputs, batch, batch_idx, dataloader_idx=0):
...
-            self.pixel_threshold.reset()
-
         self.image_metrics.reset()
         self.pixel_metrics.reset()
 
     def on_validation_epoch_end(self):
-        if self.threshold_method == ThresholdMethod.ADAPTIVE:
-            # Calculate thresholds before computing metrics.
-            self._compute_adaptive_threshold()
+        self._update_metrics_threshold()
```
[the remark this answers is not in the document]
ashwinvaidya17: I have actually removed this from the anomaly_module in the follow up PR. The method is intended to say update the threshold for metrics. `update_threshold_metrics` sounds like update the metrcs of the treshold.

src/anomalib/utils/metrics/thresholding/__init__.py
ashwinvaidya17: I like the second one

src/anomalib/utils/sweep/helpers/callbacks.py
ashwinvaidya17: No. I have removed this in the follow up PR

src/anomalib/utils/callbacks/thresholding.py
ashwinvaidya17: This refactor cannot handle adaptive. I can check against the lowercased substring but it might become ambigious if another threshold is added with adaptive in it

samet-akcay: Yeah that's fair. I just thought about backward compatibility perspective, but we're going to break it anyway

src/anomalib/utils/callbacks/thresholding.py
```diff
@@ -139,15 +130,34 @@ def _load_from_config(self, threshold: DictConfig | str | ListConfig) -> None:
...
+
+        Example:
+            >>> _get_threshold_from_config(F1AdaptiveThreshold)
+            or
+            >>> config = DictConfig({
+            ...                     "class_path": "ManualThreshold",
+            ...                     "init_args": {
+            ...                         "default_value": 0.7
+            ...                     }
+            ...                 })
```
samet-akcay, line 144:
```suggestion
            >>> config = DictConfig({
            ...    "class_path": "ManualThreshold",
            ...    "init_args": {"default_value": 0.7}
            ... })
```

src/anomalib/utils/callbacks/thresholding.py
```diff
@@ -139,15 +130,34 @@ def _load_from_config(self, threshold: DictConfig | str | ListConfig) -> None:
...
+            >>> __get_threshold_from_config(config)
+            or
+            >>> config = DictConfig({"class_path": "anomalib.utils.metrics.threshold.F1AdaptiveThreshold"})
+            >>> __get_threshold_from_config(config)
```
samet-akcay, line 148:
```suggestion
            >>> config = DictConfig({
            ...    "class_path": "anomalib.utils.metrics.threshold.F1AdaptiveThreshold"
            ... })
            >>> __get_threshold_from_config(config)
```

samet-akcay (changes requested):
Brilliant! Just a minor styling suggestion in one of the examples in docstring

samet-akcay (approved):

PR #1312, opened 2023-08-30
[Custom Trainer] Refactor Metrics

ashwinvaidya17:
## Description

- Move metrics configuration + anomaly module code into a single callback


src/anomalib/trainer/trainer.py
```diff
@@ -38,10 +40,16 @@ def __init__(
         | DictConfig
         | ListConfig
         | str = F1AdaptiveThreshold(),
+        task: TaskType = TaskType.SEGMENTATION,
+        image_metrics: list[str] | None = None,
+        pixel_metrics: list[str] | None = None,
```
samet-akcay, line 45: Would these be a `str` as well?

samet-akcay: eg., `Trainer(..., image_metrics="auc", ...)`

src/anomalib/utils/callbacks/metrics_manager.py
```diff
@@ -0,0 +1,173 @@
...
+        task (TaskType): Task type of the current run.
+        image_metrics (list[str] | None): List of image-level metrics.
+        pixel_metrics (list[str] | None): List of pixel-level metrics.
+    """
+
+    def __init__(
+        self,
+        task: TaskType = TaskType.SEGMENTATION,
+        image_metrics: list[str] | None = None,
+        pixel_metrics: list[str] | None = None,
```
samet-akcay, line 41: same as above, how about `str`?

src/anomalib/utils/callbacks/metrics_manager.py
```diff
@@ -0,0 +1,173 @@
...
+        if outputs is not None:
+            self._outputs_to_cpu(outputs)
+            self._collect_output(pl_module.image_metrics, pl_module.pixel_metrics, outputs)
+
+    def on_validation_epoch_end(
+        self,
+        trainer: "trainer.AnomalibTrainer",
+        pl_module: AnomalyModule,
+    ) -> None:
+        self._update_metrics_threshold(pl_module)
```
samet-akcay, line 111: i remember you mentioned that in the previous pr, but don't exactly remember, so dropping again.

```suggestion
        self._update_threshold_metrics(pl_module)
```

src/anomalib/utils/callbacks/metrics_manager.py
```diff
@@ -0,0 +1,173 @@
...
+        trainer: "trainer.AnomalibTrainer",
+        pl_module: AnomalyModule,
+        outputs: STEP_OUTPUT | None,
+        batch: Any,
+        batch_idx: int,
+        dataloader_idx: int = 0,
+    ) -> None:
+        if outputs is not None:
+            self._outputs_to_cpu(outputs)
+            self._collect_output(pl_module.image_metrics, pl_module.pixel_metrics, outputs)
```
samet-akcay, line 104: if `outputs` is not `None`, why would we call `collect_output`? I understand, but asking a fundamental question to make it logically more sound.

src/anomalib/utils/callbacks/metrics_manager.py
```diff
@@ -0,0 +1,173 @@
...
+        pl_module: AnomalyModule,
+    ) -> None:
+        self._log_metrics(pl_module)
+
+    def _update_metrics_threshold(self, pl_module: AnomalyModule) -> None:
+        pl_module.image_metrics.set_threshold(pl_module.image_threshold.value.item())
+        pl_module.pixel_metrics.set_threshold(pl_module.pixel_threshold.value.item())
+
+    @staticmethod
+    def _collect_output(
```
samet-akcay, line 147:
```suggestion
    def update_metrics(
```

samet-akcay: Is `_outputs_to_cpu` to compute metrics on cpu? if so can we have a `device` flag in `update_metrics` and set the device when computing the metrics? We might want to compute the metrics on gpu in the future, so this would make it configurable.

samet-akcay (changes requested):
I like it! Pipeline is getting quite clean!

Some minor comments only

src/anomalib/utils/callbacks/metrics_manager.py
```diff
@@ -0,0 +1,173 @@
...
+
+from anomalib import trainer
+from anomalib.data import TaskType
+from anomalib.models import AnomalyModule
+from anomalib.utils.metrics import AnomalibMetricCollection, create_metric_collection
+
+logger = logging.getLogger(__name__)
+
+
+class _MetricsManagerCallback(Callback):
```
samet-akcay, line 23: Also, I've checked the other callbacks that are used by Trainer. They don't seem to be using `_Manager` suffix. To be more consistent, we could maybe rename this to `MetricsCallback`, or vice-versa?

src/anomalib/utils/callbacks/metrics_manager.py
ashwinvaidya17: Good point. I'll add that condition

src/anomalib/utils/callbacks/metrics_manager.py
ashwinvaidya17: The signature of this method has outputs as None. So this is just for mypy. But yeah I'll rename it to `_update_metrics`. Might make it more sound

src/anomalib/utils/callbacks/metrics_manager.py
ashwinvaidya17: How about `_update_threshold`?

samet-akcay (approved):
thanks, looking good!

PR #1313, opened 2023-08-30
[Custom Trainer] Refactor visualization callback

ashwinvaidya17:
## Description

- Move Visualization callback into trainer


samet-akcay (approved):
Thanks @ashwinvaidya17, lgtm

PR #1314, opened 2023-08-31
[Chore] Modify README custom dataset 

Kiminjo:
## Description

An error occurs because the custom database on README.md does not have four required keys: normalization, val_split_mode, and test_split_mode. 

So I added that key.


samet-akcay (approved):
@Kiminjo, thanks for your contribution!

PR #1319, opened 2023-09-01
Fix tiling for Reverse Distillation and STFPM

### LINKED ISSUE #1318: [Bug]: Tiling is not working correctly for some models
### Describe the bug

Tiling is currently not working correctly for STFPM and Reverse Distillation. Both implementations assume that tiler is initialized ([STFPM](https://github.com/blaz-r/anomalib/blob/ensemble_eval/src/anomalib/models/stfpm/torch_model.py#L44-L49), [reverse distillation](https://github.com/blaz-r/anomalib/blob/ensemble_eval/src/anomalib/models/reverse_distillation/torch_model.py#L48-L53)) when model class constructor is called, but this is not the case as tiler gets initialized in callback.
This is not an issue in padim and patchcore, as those two models only tile embeddings.

This would require refactor of either how tiler is initialized or change the way constructors setup dimensions.

### Dataset

MVTec

### Model

STFPM

1. Install from github
2. Add following code inside [get_callbacks](https://github.com/blaz-r/anomalib/blob/ensemble_eval/src/anomalib/utils/callbacks/__init__.py#L47) to insert tiling:
```python
    tiling = TilerConfigurationCallback(
        enable=config.dataset.tiling.apply,
        tile_size=config.dataset.tiling.tile_size,
        stride=config.dataset.tiling.stride,
    )
    callbacks.append(tiling)
```
4. Train STFPM with tiler
5. Everything is successful but results are not scaled correctly (see screenshot bellow)

### OS information

OS information:
- OS: Windows 10
- Python version: 3.10
- Anomalib version: 0.7.0
- PyTorch version: 2.0.1
- CUDA/cuDNN version: 11.7
- GPU models and configuration: 1x gtx 1060
- Any other relevant information: /

### Expected behavior

Training is done and visual results are correct.

### Screenshots

As you can see, due to wrong dimensions, the mask is incorrectly up-scaled and untiled.
image
This should be:
image

### Pip/GitHub

GitHub

latest

### Configuration YAML

```yaml
dataset:
  name: mvtec
  format: mvtec
  path: ./datasets/MVTec
  category: bottle
  task: segmentation
  train_batch_size: 32
  eval_batch_size: 32
  inference_batch_size: 32
  num_workers: 0
  image_size: 256 # dimensions to which images are resized (mandatory)
  center_crop: null # dimensions to which images are center-cropped after resizing (optional)
  normalization: imagenet # data distribution to which the images will be normalized: [none, imagenet]
  transform_config:
    train: null
    eval: null
  test_split_mode: from_dir # options: [from_dir, synthetic]
  test_split_ratio: 0.2 # fraction of train images held out testing (usage depends on test_split_mode)
  val_split_mode: same_as_test # options: [same_as_test, from_test, synthetic]
  val_split_ratio: 0.5 # fraction of train/test images held out for validation (usage depends on val_split_mode)
  tiling:
    apply: true
    tile_size: 128
    stride: null
    remove_border_count: 0
    use_random_tiling: False
    random_tile_count: 16

model:
  name: stfpm
  backbone: resnet18
  layers:
    - layer1
    - layer2
    - layer3
  lr: 0.4
  momentum: 0.9
  weight_decay: 0.0001
  early_stopping:
    patience: 5
    metric: pixel_AUROC
    mode: max
  normalization_method: min_max # options: [null, min_max, cdf]

metrics:
  image:
    - F1Score
    - AUROC
  pixel:
    - F1Score
    - AUROC
  threshold:
    method: adaptive #options: [adaptive, manual]
    manual_image: null
    manual_pixel: null

visualization:
  show_images: False # show images on the screen
  save_images: True # save images to the file system
  log_images: True # log images to the available loggers (if any)
  image_save_path: null # path to which images will be saved
  mode: full # options: ["full", "simple"]

project:
  seed: 0
  path: ./results

logging:
  logger: [] # options: [comet, tensorboard, wandb, csv] or combinations.

optimization:
  export_mode: null # options: torch, onnx, openvino
# PL Trainer Args. Don't add extra parameter here.
trainer:
  enable_checkpointing: true
  default_root_dir: null
  gradient_clip_val: 0
  gradient_clip_algorithm: norm
  num_nodes: 1
  devices: 1
  enable_progress_bar: true
  overfit_batches: 0.0
  track_grad_norm: -1
  check_val_every_n_epoch: 1 # Don't validate before extracting features.
  fast_dev_run: false
  accumulate_grad_batches: 1
  max_epochs: 100
  min_epochs: null
  max_steps: -1
  min_steps: null
  max_time: null
  limit_train_batches: 1.0
  limit_val_batches: 1.0
  limit_test_batches: 1.0
  limit_predict_batches: 1.0
  val_check_interval: 1.0 # Don't validate before extracting features.
  log_every_n_steps: 50
  accelerator: auto # <"cpu", "gpu", "tpu", "ipu", "hpu", "auto">
  strategy: null
  sync_batchnorm: false
  precision: 32
  enable_model_summary: true
  num_sanity_val_steps: 0
  profiler: null
  benchmark: false
  deterministic: false
  reload_dataloaders_every_n_epochs: 0
  auto_lr_find: false
  replace_sampler_ddp: true
  detect_anomaly: false
  auto_scale_batch_size: false
  plugins: null
  move_metrics_to_cpu: false
  multiple_trainloader_mode: max_size_cycle
```

### Logs

```shell
...
```

### Code of Conduct

blaz-r:
## Description

This PR fixes the tiling for STFPM and Reverse Distillation models. It also adds tiling support to Fastflow.

For STFPM and Reverse Distillation, instead of relying on the tiler initialization for size, we avoid this issue by rather untiling the features right before they are returned or use in anomaly map generator.

The implementation of Fastflow tiling is up to debate, but I tried to keep the predictions in tiled shape for the longest duration of procedure to offer the best benefits of reduced memory usage.

This PR does NOT add the callback for tiling setup because it's currently not present so I avoided adding it. To utilize (and verify) tiling, the following should be added to the callbacks:
```python
tiling = TilerConfigurationCallback(
    enable=config.dataset.tiling.apply,
    tile_size=config.dataset.tiling.tile_size,
    stride=config.dataset.tiling.stride,
)
callbacks.append(tiling)
```

- Fixes #1318 

## Changes

<details>

- [ ] Documentation update

</details>

## Checklist

<details>

</details>

### DIFF
--- a/src/anomalib/models/fastflow/lightning_model.py
+++ b/src/anomalib/models/fastflow/lightning_model.py
@@ -21,12 +21,12 @@ class Fastflow(AnomalyModule):
 
     Args:
         input_size (tuple[int, int]): Model input size.
-        image_size (tuple[int, int]): Original image size (needed in case of tiling).
         backbone (str): Backbone CNN network
         pre_trained (bool, optional): Boolean to check whether to use a pre_trained backbone.
         flow_steps (int, optional): Flow steps.
         conv3x3_only (bool, optinoal): Use only conv3x3 in fast_flow model. Defaults to False.
         hidden_ratio (float, optional): Ratio to calculate hidden var channels. Defaults to 1.0.
+        anomaly_map_size (tuple[int, int] | None): Size of the output anomaly map, if None, it matches input_size.
     """
 
     def __init__(
@@ -37,7 +37,7 @@ class Fastflow(AnomalyModule):
         flow_steps: int = 8,
         conv3x3_only: bool = False,
         hidden_ratio: float = 1.0,
-        image_size: tuple[int, int] | None = None,
+        anomaly_map_size: tuple[int, int] | None = None,
     ) -> None:
         super().__init__()
 
@@ -48,7 +48,7 @@ class Fastflow(AnomalyModule):
             flow_steps=flow_steps,
             conv3x3_only=conv3x3_only,
             hidden_ratio=hidden_ratio,
-            image_size=image_size,
+            anomaly_map_size=anomaly_map_size,
         )
         self.loss = FastflowLoss()
 
@@ -101,12 +101,12 @@ class FastflowLightning(Fastflow):
 
         super().__init__(
             input_size=input_size,
-            image_size=hparams.dataset.image_size,
             backbone=hparams.model.backbone,
             pre_trained=hparams.model.pre_trained,
             flow_steps=hparams.model.flow_steps,
             conv3x3_only=hparams.model.conv3x3_only,
             hidden_ratio=hparams.model.hidden_ratio,
+            anomaly_map_size=hparams.dataset.image_size,
         )
         self.hparams: DictConfig | ListConfig  # type: ignore
         self.save_hyperparameters(hparams)
--- a/src/anomalib/models/fastflow/lightning_model.py
+++ b/src/anomalib/models/fastflow/lightning_model.py
@@ -21,6 +21,7 @@ class Fastflow(AnomalyModule):
 
     Args:
         input_size (tuple[int, int]): Model input size.
+        image_size (tuple[int, int]): Original image size (needed in case of tiling).
         backbone (str): Backbone CNN network
         pre_trained (bool, optional): Boolean to check whether to use a pre_trained backbone.
         flow_steps (int, optional): Flow steps.
@@ -31,6 +32,7 @@ class Fastflow(AnomalyModule):
     def __init__(
         self,
         input_size: tuple[int, int],
+        image_size: tuple[int, int],
         backbone: str,
         pre_trained: bool = True,
         flow_steps: int = 8,
@@ -41,6 +43,7 @@ class Fastflow(AnomalyModule):
 
         self.model = FastflowModel(
             input_size=input_size,
+            image_size=image_size,
             backbone=backbone,
             pre_trained=pre_trained,
             flow_steps=flow_steps,
@@ -90,8 +93,15 @@ class FastflowLightning(Fastflow):
     """
 
     def __init__(self, hparams: DictConfig | ListConfig) -> None:
+        # if we use tiling, input size of model needs to be same as tile size
+        if hparams.dataset.tiling.apply:
+            input_size = hparams.dataset.tiling.tile_size
+        else:
+            input_size = hparams.model.input_size
+
         super().__init__(
-            input_size=hparams.model.input_size,
+            input_size=input_size,
+            image_size=hparams.dataset.image_size,
             backbone=hparams.model.backbone,
             pre_trained=hparams.model.pre_trained,
             flow_steps=hparams.model.flow_steps,
--- a/src/anomalib/models/fastflow/lightning_model.py
+++ b/src/anomalib/models/fastflow/lightning_model.py
@@ -26,7 +26,6 @@ class Fastflow(AnomalyModule):
         flow_steps (int, optional): Flow steps.
         conv3x3_only (bool, optinoal): Use only conv3x3 in fast_flow model. Defaults to False.
         hidden_ratio (float, optional): Ratio to calculate hidden var channels. Defaults to 1.0.
-        anomaly_map_size (tuple[int, int] | None): Size of the output anomaly map, if None, it matches input_size.
     """
 
     def __init__(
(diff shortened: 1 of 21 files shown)


blaz-r:
Okay sure. I think I broke backwards compatibility for fastflow. I'll fix this and rerun tests locally.

blaz-r:
Tests are now passing, and fastflow is now backwards compatible.
I see that coverage dropped, because tiling in models is not covered with tests. As far as I am aware, this is not covered for any model, so I didn't add tests, maybe I should?

src/anomalib/models/fastflow/lightning_model.py
```diff
@@ -21,6 +21,7 @@ class Fastflow(AnomalyModule):
 
     Args:
         input_size (tuple[int, int]): Model input size.
+        image_size (tuple[int, int]): Original image size (needed in case of tiling).
```
samet-akcay, line 24: Would having both `image_size` and `input_size` in the model definition be confusing to the user? @ashwinvaidya17, any thoughts?

src/anomalib/models/reverse_distillation/torch_model.py
```diff
@@ -45,12 +45,7 @@ def __init__(
         self.bottleneck = get_bottleneck_layer(backbone)
         self.decoder = get_decoder(backbone)
 
-        if self.tiler:
-            image_size = (self.tiler.tile_size_h, self.tiler.tile_size_w)
-        else:
-            image_size = input_size
-
-        self.anomaly_map_generator = AnomalyMapGenerator(image_size=image_size, mode=anomaly_map_mode)
+        self.anomaly_map_generator = AnomalyMapGenerator(image_size=input_size, mode=anomaly_map_mode)
```
samet-akcay, line 48: A reference for future, but not it might make more sense to rename `image_size` argument to `input_size` in `AnomalyMapGenerator`. As I said, it is not related to this PR, though

samet-akcay (commented):
@blaz-r, thanks for creating this PR. I have a single comment, which we could perhaps discuss with @ashwinvaidya17

blaz-r:
> @blaz-r, thanks for creating this PR. I have a single comment, which we could perhaps discuss with @ashwinvaidya17

Okay. Yeah the naming is a bit weird and I'm open to suggestions for renaming. I'm just not sure what to call image_size since it's called that in AnomalyMapGenerator. In case of normal FastFlow, input size is also image size, but when tiling is performed, input size is actually tile size and image size is image size, so I'm not sure how to name these so this is clearly indicated.

src/anomalib/models/fastflow/lightning_model.py
ashwinvaidya17: Maybe we can just call it `original_image_size`?

ashwinvaidya17: Also, not sure if I quite understood it. The only place we use the original_image_size is in anomaly_map_generator so that it can interpolate the flow maps. But why can't we pass the model input_size? Aren't the tiles generated from the input tensor that is in shape of model input_size? Also, won't the visualizer take care of re-scaling it further? Am I missing something?

ashwinvaidya17 (commented):
Except for the same `input_size` comment raised by Samet I have no other feedback. Just one question though

src/anomalib/models/fastflow/lightning_model.py
blaz-r: Input size is in case of tiling equal to tile size, because the fastflow model requires input size at the time of initialization. If we pass just tile size, we can't really know the size of anomaly maps, and if we pass just image size, we can't init the model as we don't know the tile size in the constructor. 

This is not a problem for other models as they don't require tile size in their constructors. And even if the visualizer handles this, the metrics don't as they expect the same size of GT and anomaly map.

On the other hand, I think this could be renamed to anomal_map_size, which would be better than image_size or original_image_size.

ashwinvaidya17: Alright thanks for the explination. Yeah we can rename it to `anomaly_map_size`. Apart from this I have no other suggestions.

blaz-r:
`image_size` is now renamed to `anomaly_map_size`.

ashwinvaidya17 (approved):
Thanks again for the fix

blaz-r:
Fastflow tiling will be refactored to match other architectures.

blaz-r:
The addition to fastflow is reverted now. This PR only fixes tiling for Reverse Distillation and STFPM.

samet-akcay (approved):
lgtm, thanks

PR #1386, opened 2023-10-05
Patch for the WinError183 on the OpenVino export mode

ggiret-thinkdeep:
## Description

I added checking whether the file exists or not. If the file exists, it is overwritten
The error only occurs on Windows

- Fixes # 1385

## Changes

Add a check for the existence of a file

## Why ?

From the Pathlib doc : 
`Path.rename(target)[¶](https://docs.python.org/3/library/pathlib.html#pathlib.Path.rename)
Rename this file or directory to the given target, and return a new Path instance pointing to target. On Unix, if target exists and is a file, it will be replaced silently if the user has permission. On Windows, if target exists, [FileExistsError](https://docs.python.org/3/library/exceptions.html#FileExistsError) will be raised.`

In order to prevent Windows error, the added line checks if the file exists. If it already exists, then it is deleted. This will allow Pathlib's rename function to run without errors on Windows.


ashwinvaidya17 (commented):
Can you also add a comment why this line is added?

ggiret-thinkdeep:
> Can you also add a comment why this line is added?

Done 👍

ashwinvaidya17:
> > Can you also add a comment why this line is added?
> 
> Done 👍

Sorry for the confusion. I meant adding a comment in the code so that the reader does not delete it by mistake.

ggiret-thinkdeep:
> Sorry for the confusion. I meant adding a comment in the code so that the reader does not delete it by mistake.

Oops, it's ok now normaly

ashwinvaidya17 (approved):
Thanks!

PR #1389, opened 2023-10-08
Refactor Reverse Distillation to match official code

abc-125:
## Description

I was looking into why this version of Reverse Distillation [shows a bit lower results](https://github.com/openvinotoolkit/anomalib/pull/343#discussion_r893471188) compared to the results stated in the paper. I have compared this implementation with the [official code](https://github.com/hq-deng/RD4AD). Seems like there are several issues:

- Early stop causes training to stop too quickly and results are low, the paper states they have used 200 epochs without early stop, the same as the official code.
- The batch size is 16, and beta2 is 0.999 both in the official code and the paper, so I have used these values as well.
- In the OCBE class I have added the 4th conv layer, the same as the official code.
- I have changed the default mode from "multiply" to "add", the same as in the original repository. It also makes more sense if we want to detect more defects. RD creates 3 anomaly maps for different scales by default, and by adding them we can detect defects that are on only one of these maps.
- After all these changes results are still lower than needed, but the official code (not the paper), uses a slightly different loss. I have changed the loss to the one from the official repo and got the final results.
- If somebody wants results that are even closer to the paper, they can use torchvision feature extractor instead of timm. I have added a comment in the code on this, but not implemented it because this seems that in general timm works better, except for Toothbrush in MVTec.

## Changes

<details>

- [ ] Documentation update

</details>

## Checklist

<details>

</details>


src/anomalib/models/reverse_distillation/README.md
```diff
@@ -20,79 +20,32 @@ During testing, a similar step is followed but this time the cosine distance bet
...
-| Zipper     |     0.944 |          0.926 |
-| Average    |     0.922 |          0.891 |
+|                |  Avg  | Carpet | Grid  | Leather | Tile  | Wood  | Bottle | Cable | Capsule | Hazelnut | Metal Nut | Pill  | Screw | Toothbrush | Transistor | Zipper |
+| -------------- | :---: | :----: | :---: | :-----: | :---: | :---: | :----: | :---: | :-----: | :------: | :-------: | :---: | :---: | :--------: | :--------: | :----: |
+| Wide ResNet-50 | 0.976 | 0.977  | 1.000 |  1.000  | 0.994 | 0.992 | 0.984  | 0.930 |  0.982  |  1.000   |   1.000   | 0.967 | 0.963 |   0.952    |   0.927    | 0.975  |
 
 ### Sample Results
 
-![Sample Result 1](../../../docs/source/images/reverse_distillation/results/0.png "Sample Result 1")
+![Sample Result 1](../../../../docs/source/images/reverse_distillation/results/0.png "Sample Result 1")
```
samet-akcay, line 47:
```suggestion
![Sample Result 1](https://raw.githubusercontent.com/openvinotoolkit/anomalib/main/docs/source/images/reverse_distillation/results/0.png "Sample Result 1")
```

src/anomalib/models/reverse_distillation/README.md
```diff
@@ -20,79 +20,32 @@ During testing, a similar step is followed but this time the cosine distance bet
...
 
-![Sample Result 2](../../../docs/source/images/reverse_distillation/results/1.png "Sample Result 2")
+![Sample Result 2](../../../../docs/source/images/reverse_distillation/results/1.png "Sample Result 2")
```
samet-akcay, line 49:
```suggestion
![Sample Result 2](https://raw.githubusercontent.com/openvinotoolkit/anomalib/main/docs/source/images/reverse_distillation/results/1.png "Sample Result 2")
```

src/anomalib/models/reverse_distillation/README.md
```diff
@@ -20,79 +20,32 @@ During testing, a similar step is followed but this time the cosine distance bet
...
 
-![Sample Result 3](../../../docs/source/images/reverse_distillation/results/2.png "Sample Result 3")
+![Sample Result 3](../../../../docs/source/images/reverse_distillation/results/2.png "Sample Result 3")
```
samet-akcay, line 51:
```suggestion
![Sample Result 3](https://raw.githubusercontent.com/openvinotoolkit/anomalib/main/docs/source/images/reverse_distillation/results/2.png "Sample Result 3")
```

src/anomalib/models/reverse_distillation/config.yaml
```diff
@@ -35,21 +35,20 @@ model:
...
     - layer2
     - layer3
-  early_stopping:
-    patience: 3
-    metric: pixel_AUROC
-    mode: max
+  # early_stopping: # optional
+  #   patience: 3
+  #   metric: pixel_AUROC
+  #   mode: max
```
samet-akcay, line 41: We could either increase the patience to act like there is no early stopping. Or we could remove this all together. Commented out code/config might be confusing to some

src/anomalib/models/reverse_distillation/lightning_model.py
```diff
@@ -134,17 +134,20 @@ def __init__(self, hparams: DictConfig | ListConfig) -> None:
...
-        return [early_stopping]
+        if "early_stopping" in self.hparams.model:
+            early_stopping = EarlyStopping(
+                monitor=self.hparams.model.early_stopping.metric,
+                patience=self.hparams.model.early_stopping.patience,
+                mode=self.hparams.model.early_stopping.mode,
+            )
+            return [early_stopping]
+        else:
+            return []
```
samet-akcay, line 153: We could use a single `return` statement and reduce if-else blocks via the following approach?
```suggestion
        callbacks = []
        if "early_stopping" in self.hparams.model:
            early_stopping = EarlyStopping(
                monitor=self.hparams.model.early_stopping.metric,
                patience=self.hparams.model.early_stopping.patience,
                mode=self.hparams.model.early_stopping.mode,
            )
            callbacks.append(early_stopping)
        return callbacks
```

src/anomalib/models/reverse_distillation/loss.py
```diff
@@ -23,8 +27,13 @@ def forward(self, encoder_features: list[Tensor], decoder_features: list[Tensor]
...
-        for loss in losses:
-            loss_sum += torch.mean(1 - loss)  # mean of cosine distance
+        for item in range(len(encoder_features)):
+            loss_sum += torch.mean(
+                1
+                - cos_loss(
+                    encoder_features[item].view(encoder_features[item].shape[0], -1),
+                    decoder_features[item].view(decoder_features[item].shape[0], -1),
+                )
+            )
```
samet-akcay, line 38: Based on the codacy comment above, maybe we could have something like this?
```suggestion
        for _idx, (encoder_feature, decoder_feature) in enumerate(zip(encoder_features, decoder_features)):
            loss_sum += torch.mean(
                1
                - cos_loss(
                    encoder_feature.view(encoder_feature.shape[0], -1),
                    decoder_feature.view(decoder_feature.shape[0], -1),
                )
            )
```

samet-akcay (changes requested):
@abc-125, thanks a lot for your contribution! 

I only have minor comments. I added some suggestions, but you don't have to agree :)

src/anomalib/models/reverse_distillation/config.yaml
abc-125: I think it's better to remove it from the config, it looks like 200 epochs give good results for the variety of tasks. If I remember it correctly, training RD for 200 epochs achieved 96 AUROC for images at VisA.

src/anomalib/models/reverse_distillation/README.md
```diff
@@ -20,79 +20,32 @@ During testing, a similar step is followed but this time the cosine distance bet
...
-| Metal_nut  |     0.999 |          0.929 |
-| Pill       |     0.944 |          0.553 |
-| Screw      |     0.778 |           0.86 |
-| Tile       |     0.833 |          0.513 |
-| Toothbrush |     0.967 |            0.7 |
-| Transistor |     0.928 |          0.829 |
-| Wood       |     0.989 |          0.993 |
-| Zipper     |     0.968 |          0.787 |
-| Average    |     0.936 |           0.79 |
+|                |  Avg  | Carpet | Grid  | Leather | Tile  | Wood  | Bottle | Cable | Capsule | Hazelnut | Metal Nut | Pill  | Screw | Toothbrush | Transistor | Zipper |
```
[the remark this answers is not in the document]
abc-125: Now it's only one backbone in the results and having horizontal table looks better to me, but I am not sure how it should be done in anomalib :) I saw both vertical and horizontal tables for different models, so I went with the one which looks easier to read.

abc-125:
> @abc-125, thanks a lot for your contribution! 

Thank you and all contributors for this amazing library :)

abc-125:
@samet-akcay Thank you for your comments and suggestions. I don't have time now unfortunately, but I will take a look at all of them during the weekend.

samet-akcay (approved):
Thanks @abc-125, this is looking great!

PR #1394, opened 2023-10-10
Code quality/enable rules

samet-akcay:
## Description

TODO
- [x] Add rules to `pyproject.toml`.
- [x] Pyflakes (`F`)
- [x] pycodestyle error (`E`)
- [x] pycodestyle warning (`W`)
- [x] isort (`I`)
- [x] pep8-naming (`N`)

### DIFF
--- a/src/anomalib/data/base/depth.py
+++ b/src/anomalib/data/base/depth.py
@@ -3,7 +3,7 @@
 
 from abc import ABC
 
-import albumentations as A
+import albumentations as A  # noqa: N812
 import cv2
 import numpy as np
 from torch import Tensor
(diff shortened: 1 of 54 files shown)


src/anomalib/data/base/depth.py
```diff
@@ -3,7 +3,7 @@
 
 from abc import ABC
 
-import albumentations as A
+import albumentations as A  # noqa: N812
```
ashwinvaidya17, line 6: Maybe we can keep it as `import albumentations`. And same for `torch.functional`

samet-akcay: I would also prefer that to be honest. Wanted to get your opinion

ashwinvaidya17 (approved):

src/anomalib/data/base/depth.py
samet-akcay: settled then? :)

PR #1407, opened 2023-10-12
Enable Ruff Rules - Part 6

samet-akcay:
Part 6 of the ruff rule enablement. These are the rules that have been enabled in this PR:

- flake8-simplify (`SIM`)
- flake8-tidy-imports (`TID`)
- flake8-type-checking (`TCH`)
- flake8-gettext (`INT`)
- flake8-unsused-arguments (`ARG`)
- flake8-use-pathlib (`PTH`)
- eradicate (`ERA`)
- pylint (`PL`)
- tryceratos (`TRY`)
- flynt (`FLY`)
- NumPy-specific rules (`NPY`)
- Perflint (`PERF`)
- Ruff-specific rules (`RUF`)


pyproject.toml
```diff
@@ -79,12 +79,12 @@ ignore = [
...
-    # # pylint
-    # "PLR0913",  # Too many arguments to function call
-    # "PLR2004",  # consider replacing with a constant variable
-    # "PLW2901",  # `for` loop variable `row` overwritten by assignment target
-    # "PLR0912",  # Too many branches
-    # "PLR0915",  # Too many statements
+    # pylint
+    "PLR0913",  # Too many arguments to function call
+    "PLR2004",  # consider replacing with a constant variable
+    "PLW2901",  # `for` loop variable `row` overwritten by assignment target
```
ashwinvaidya17, line 85: Is this safe to ignore?

ashwinvaidya17 (commented):
Only one comment

pyproject.toml
samet-akcay: We have some implementations that look like this:
```python
        filtered_regions = []
        for im_regions in regions:
            if self.persons_only:
                im_regions = self._keep_only_persons(im_regions)
            im_regions = self._filter_by_area(im_regions, self.min_bbox_area)
            im_regions = self._delete_overlapping_boxes(im_regions, self.max_bbox_overlap)
            filtered_regions.append(im_regions)
```

Maybe I could add a TODO to such implementations to fix them later?

ashwinvaidya17: How many of such lines do we have? Otherwise can we ignore these files and add a todo so that we can keep the rule?

samet-akcay: There are quite a few, but yeah, we could add a todo to address them. I'll do that in the follow-up PRs though. Changing it here might mess up part 7 and 8

PR #1408, opened 2023-10-12
Enable Ruff Rules - Part 7

samet-akcay:
Part 7 of the ruff rule enablement. These are the rules that have been enabled in this PR:
- [mccabe (C90)](https://github.com/openvinotoolkit/anomalib/commit/c7fe4e9cdb293d3ef96841f52b0e78206dc7af70)
- [pygrep-hooks (PGH)](https://github.com/openvinotoolkit/anomalib/commit/250057ad33e1a410e0bff6c43e9fb0269b5b6032)
- [flake8-todos (TD)](https://github.com/openvinotoolkit/anomalib/commit/f111d869df5de88a75070a7491df59f34298dd60)
- [flake8-fixme (FIX)](https://github.com/openvinotoolkit/anomalib/commit/3a63d48b63666107cd9f9db75abf05fc587aee3f)
- [pandas-vet (PD)](https://github.com/openvinotoolkit/anomalib/commit/fe329f15cbedab3bd818954fe6f72052f7632a8d)


src/anomalib/models/cflow/lightning_model.py
```diff
@@ -57,8 +57,8 @@ def __init__(
             permute_soft=permute_soft,
         )
         self.automatic_optimization = False
-        # TODO: LR should be part of optimizer in config.yaml! Since cflow has custom
-        #   optimizer this is to be addressed later.
+        # TODO(ashwinvaidya17): Use padding="same" in nn.Conv2d once PyTorch v2.1 is released
+        # CVS-122671
```
ashwinvaidya17, line 61: I think this comment should be about the lr

tests/pre_merge/datasets/test_dataset.py
```diff
@@ -55,10 +55,8 @@ def test_random_split(self, folder_dataset):
         # check if reconstructed dataset is equal to original dataset
         assert folder_dataset.samples.equals(reconstructed_dataset.samples)
 
-        # check if warning raised when one of the subsets is empty
-        split_ratios = [1 - (1 / (len(folder_dataset) + 1)), 1 / (len(folder_dataset) + 1)]
-        with pytest.warns():
-            subsets = random_split(folder_dataset, split_ratios)
+        # TODO(djdameln): Check if warning raised when one of the subsets is empty
+        # CVS-122709
```
ashwinvaidya17, line 59: Should we leave the removed code in comments? Otherwise we might forget what this todo was referring to

ashwinvaidya17 (commented):
Thanks minor comments

tests/pre_merge/datasets/test_dataset.py
samet-akcay: hmm, one of the rules complain when there is a commented code, but I guess we could temporarily ignore it. Alternatively include that in the ticket

djdameln: Let's change this to
```suggestion
        # TODO(djdameln): Add check for handling of empty subsets in case of small split ratio CVS-122709
```

samet-akcay: I think this is the format suggested by the rule. Your suggestion would not work unfortunately. 
```
# TODO(djdameln): Check if warning raised when one of the subsets is empty
# CVS-122709
```

PR #1412, opened 2023-10-13
Enable Ruff Rules - Part 8

samet-akcay:
Part 8 of the ruff rule enablement. These are the rules that have been enabled in this PR:

- flake8-annotations (`ANN`)


ashwinvaidya17 (approved):
Wow I didn't realize how much of our code was still untyped.

PR #1418, opened 2023-10-16
Add new configs

ashwinvaidya17:
## Description

- Adds new config

## Changes

<details>

- [ ] Documentation update

</details>

## Checklist

<details>

</details>


src/anomalib/models/cflow/anomaly_map.py
```diff
@@ -18,12 +19,12 @@ class AnomalyMapGenerator(nn.Module):
...
         self,
         image_size: ListConfig | tuple,
-        pool_layers: list[str],
+        pool_layers: Sequence[str],
     ) -> None:
         super().__init__()
         self.distance = torch.nn.PairwiseDistance(p=2, keepdim=True)
         self.image_size = image_size if isinstance(image_size, tuple) else tuple(image_size)
-        self.pool_layers: list[str] = pool_layers
+        self.pool_layers: Sequence[str] = pool_layers
```
samet-akcay, line 27: Just curiosity, why is this changed to `Sequence`? CLI-related?

src/anomalib/models/cflow/lightning_model.py
```diff
@@ -64,16 +64,9 @@ def __init__(
...
-            This method is used for the existing CLI.
-            When PL CLI is introduced, configure optimizers method will be
-                deprecated, and optimizers will be configured from either
-                config.yaml file or from CLI.
-
-        Returns:
+        Returns
         -------
-            Optimizer: Adam optimizer for each decoder
+                    Optimizer: Adam optimizer for each decoder
```
samet-akcay, line 69:
```suggestion
    Optimizer: Adam optimizer for each decoder
```

src/anomalib/models/components/sampling/k_center_greedy.py
```diff
@@ -6,9 +6,9 @@
 
 
 import torch
+from rich.progress import track
```
samet-akcay, line 9: Would it be an idea to replace all of the tqdm with track? Maybe in another PR?

src/anomalib/models/csflow/lightning_model.py
```diff
@@ -102,6 +101,21 @@ def trainer_arguments(self) -> dict[str, Any]:
         """CS-Flow-specific trainer arguments."""
         return {"gradient_clip_val": 1, "num_sanity_val_steps": 0}
 
+    def configure_optimizers(self) -> torch.optim.Optimizer:
+        """Configure optimizers.
+
+        Returns
+        -------
```
samet-akcay, line 108:
```suggestion
        Returns:
```

src/anomalib/models/dfm/lightning_model.py
```diff
@@ -24,9 +24,9 @@ class Dfm(AnomalyModule):
 
     Args:
     ----
+        input_size (tuple[int, int]): Input size for the model.
```
samet-akcay, line 27: Should we also add a default value for the `input_size`? Is this specified in the paper?

src/anomalib/models/efficient_ad/lightning_model.py
```diff
@@ -90,8 +89,8 @@ class EfficientAd(AnomalyModule):
 
     def __init__(
         self,
-        teacher_out_channels: int,
-        image_size: tuple[int, int],
+        input_size: tuple[int, int],
```
samet-akcay, line 92: Same as above, would it be an idea to provide a default value to `input_size`? Maybe we could discuss this?

src/configs/data/mvtec.yaml
```diff
@@ -0,0 +1,18 @@
+class_path: anomalib.data.MVTec
```
samet-akcay, line 1: Not in this PR, but I think it is time to rename this to MVTecAD. This is the original name of this dataset, and we have multiple MVTec datasets now

src/configs/model/ai_vad.yaml
samet-akcay, line 1: This config file does not have `metric` or `trainer` arguments. Is it because those are not needed?

samet-akcay: I've noticed that some other models also do not have these sections

samet-akcay (changes requested):
Thanks! The new config structure looks really clean!

I've just minor questions and comments

src/anomalib/models/cflow/anomaly_map.py
ashwinvaidya17: I cannot assign default parameters if I keep the type as list. With Sequence, I can use both tuple and list

src/anomalib/models/dfm/lightning_model.py
ashwinvaidya17: We override the input size with that of the dataloader. And, if someone uses the API and sets the dataloader input size to a different value then it will throw an error.

samet-akcay: yeah, they need to explicitly specify the size to avoid issues. But this is the current way to create the datamodule and the model via API, right?

```python
from anomalib.data import Visa
from anomalib.models import EfficientAd

datamodule = Visa(image_size=(256, 256), ...)
model = EfficientAd(input_size(256, 256))
...
```

src/anomalib/data/base/dataset.py
```diff
@@ -172,11 +172,10 @@ def _setup(self) -> DataFrame:
...
-        |---|-------------------|-----------|-------------|------------------|-------|
-        |   | image_path        | label     | label_index | mask_path        | split |
-        |---|-------------------|-----------|-------------|------------------|-------|
-        | 0 | path/to/image.png | anomalous | 1           | path/to/mask.png | train |
-        |---|-------------------|-----------|-------------|------------------|-------|
+            |---|-------------------|-----------|-------------|------------------|-------|
+            |   | image_path        | label     | label_index | mask_path        | split |
+            |---|-------------------|-----------|-------------|------------------|-------|
+            | 0 | path/to/image.png | anomalous | 1           | path/to/mask.png | train |
+            |---|-------------------|-----------|-------------|------------------|-------|
```
samet-akcay, line 179: I've noticed that this table is not properly shown on the documentation. We need to make this a bit html friendly. Of course, in another PR when we are working on the documentation :)

samet-akcay (approved):
lgtm, thanks!

PR #1431, opened 2023-10-19
Fix DRAEM

blaz-r:
## Description

While I've been experimenting with DRAEM model I noticed that the results are way below the ones presented in paper. After some experimenting I noticed that there are probably some reasons:
1. [**major**], the images are scaled to a different range (imagenet) than the generated anomalies
2. [minor], scheduler wasn't set as described in paper
3. [minor], beta isn't set correctly in case of single float ([original implementation still uses rand * float](https://github.com/VitjanZ/DRAEM/blob/main/data_loader.py#L126))
4. [very minor], [[Describable Textures dataset](https://www.robots.ox.ac.uk/~vgg/data/dtd/)] is not used for generated anomalies

What I changed:
1. Set normalization to none
2. Added scheduler to make a step at 400 and 600 epoch
3. Fixed beta calculation to match the paper
4. DTD can already be utilized, but I don't think it needs to be by default, so I left it as is

---

## Experiments

Default normalization should be set to none, so images are scaled to [0, 1], which is the same range as the anomalies generated and same as in the [original implementation](https://github.com/VitjanZ/DRAEM/blob/main/data_loader.py#L152).
This greatly improved the results:

---

### Cable
- Normalization = none
image
- Normalization = imagenet
image

---

### Transistor
- Normalization = none
image
- Normalization = imagenet
image

---

The above results were obtained with default config, without early stopping and scheduler enabled. The difference is quite significant, while I'm not sure why the image AUROC is so poor in case of cable.

## Changes

<details>

- [ ] Documentation update

</details>

## Checklist

<details>

</details>

### DIFF
--- a/src/anomalib/data/utils/augmenter.py
+++ b/src/anomalib/data/utils/augmenter.py
@@ -158,7 +158,8 @@ class Augmenter:
 
         # Apply perturbations batch wise
         if isinstance(self.beta, float):
-            beta = self.beta
+            beta = torch.rand(batch_size) * self.beta
+            beta = beta.view(batch_size, 1, 1, 1).expand_as(batch).to(batch.device)  # type: ignore
         elif isinstance(self.beta, tuple):
             beta = torch.rand(batch_size) * (self.beta[1] - self.beta[0]) + self.beta[0]
             beta = beta.view(batch_size, 1, 1, 1).expand_as(batch).to(batch.device)  # type: ignore
--- a/src/anomalib/models/draem/lightning_model.py
+++ b/src/anomalib/models/draem/lightning_model.py
@@ -145,9 +145,8 @@ class DraemLightning(Draem):
         )
         return [early_stopping]
 
-    def configure_optimizers(self) -> torch.optim.Optimizer:
+    def configure_optimizers(self) -> tuple[list[torch.optim.Optimizer], list[torch.optim.lr_scheduler.LRScheduler]]:
         """Configure the Adam optimizer."""
         optimizer = torch.optim.Adam(params=self.model.parameters(), lr=self.hparams.model.lr)
-        scheduler1 = torch.optim.lr_scheduler.StepLR(optimizer, step_size=400, gamma=0.1)
-        scheduler2 = torch.optim.lr_scheduler.StepLR(optimizer, step_size=600, gamma=0.1)
-        return optimizer, [scheduler1, scheduler2]
+        scheduler = torch.optim.lr_scheduler.MultiStepLR(optimizer, milestones=[400, 600], gamma=0.1)
+        return [optimizer], [scheduler]
--- a/src/anomalib/models/draem/lightning_model.py
+++ b/src/anomalib/models/draem/lightning_model.py
@@ -147,4 +147,7 @@ class DraemLightning(Draem):
 
     def configure_optimizers(self) -> torch.optim.Optimizer:
         """Configure the Adam optimizer."""
-        return torch.optim.Adam(params=self.model.parameters(), lr=self.hparams.model.lr)
+        optimizer = torch.optim.Adam(params=self.model.parameters(), lr=self.hparams.model.lr)
+        scheduler1 = torch.optim.lr_scheduler.StepLR(optimizer, step_size=400, gamma=0.1)
+        scheduler2 = torch.optim.lr_scheduler.StepLR(optimizer, step_size=600, gamma=0.1)
+        return optimizer, [scheduler1, scheduler2]
--- a/src/anomalib/models/draem/config.yaml
+++ b/src/anomalib/models/draem/config.yaml
@@ -9,7 +9,7 @@ dataset:
   num_workers: 8
   image_size: 256 # dimensions to which images are resized (mandatory)
   center_crop: null # dimensions to which images are center-cropped after resizing (optional)
-  normalization: imagenet # data distribution to which the images will be normalized: [none, imagenet]
+  normalization: none # data distribution to which the images will be normalized: [none, imagenet]
   transform_config:
     train: null
     eval: null


samet-akcay (approved):
Thanks for addressing this @blaz-r, this was on the todo list for some time

src/anomalib/data/utils/augmenter.py
```diff
@@ -158,7 +158,8 @@ def augment_batch(self, batch: Tensor) -> tuple[Tensor, Tensor]:
 
         # Apply perturbations batch wise
         if isinstance(self.beta, float):
-            beta = self.beta
+            beta = torch.rand(batch_size) * self.beta
+            beta = beta.view(batch_size, 1, 1, 1).expand_as(batch).to(batch.device)  # type: ignore
```
djdameln, line 162: I'm not sure about this one. The intended use of passing a single value is to have a fixed beta value across images, which might be preferred in some cases. For random beta values, the user could pass a tuple to set the range of possible beta values. This is already the default behaviour of DRAEM, so I don't think we need to change the augmenter for this.

The only difference that I can see is that the paper mentions a range of [0.1, 1.0], while the default range of the augmenter is [0.2, 1.0]. We could easily address this by either hard-coding the correct range in the lightning model and passing it to the constructor of the augmenter, or by making it configurable from the `model` section of the config file.

src/anomalib/models/draem/lightning_model.py
```diff
@@ -145,6 +145,8 @@ def configure_callbacks(self) -> list[EarlyStopping]:
         )
         return [early_stopping]
 
-    def configure_optimizers(self) -> torch.optim.Optimizer:
+    def configure_optimizers(self) -> tuple[list[torch.optim.Optimizer], list[torch.optim.lr_scheduler.LRScheduler]]:
         """Configure the Adam optimizer."""
-        return torch.optim.Adam(params=self.model.parameters(), lr=self.hparams.model.lr)
+        optimizer = torch.optim.Adam(params=self.model.parameters(), lr=self.hparams.model.lr)
+        scheduler = torch.optim.lr_scheduler.MultiStepLR(optimizer, milestones=[400, 600], gamma=0.1)
+        return [optimizer], [scheduler]
```
djdameln, line 152: We previously did not add this because early stopping is enabled, and the training rarely reaches 400 epochs. But since we recently decided to always use the configuration recommended in the paper as the default, I agree it would be good to add it now.

In this case we should probably also disable early stopping by default.

djdameln (changes requested):
> DTD can already be utilized, but I don't think it needs to be by default, so I left it as is. But I think we might want to note that it's necessary to match the performance from the paper.

If we set the DTD dataset as the default image source, we also need to make sure that Anomalib automatically downloads the dataset if it's not available locally, otherwise DRAEM will not run out of the box. The problem here is that we are unsure of the legal implications of downloading the dataset through Anomalib, because the dataset does not seem to be released under some specific license. This is something that we need to investigate first. I could maybe ask some guidance from the creators of the dataset for this.

We already recommend to use the DTD dataset in the `README` file of the DRAEM model ([here](https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/models/draem/README.md?plain=1#L11)). We could make it a bit more explicit and mention that this is needed to match the performance of the original paper.

blaz-r:
> > DTD can already be utilized, but I don't think it needs to be by default, so I left it as is. But I think we might want to note that it's necessary to match the performance from the paper.
> 
> If we set the DTD dataset as the default image source, we also need to make sure that Anomalib automatically downloads the dataset if it's not available locally, otherwise DRAEM will not run out of the box. The problem here is that we are unsure of the legal implications of downloading the dataset through Anomalib, because the dataset does not seem to be released under some specific license. This is something that we need to investigate first. I could maybe ask some guidance from the creators of the dataset for this.
> 
> We already recommend to use the DTD dataset in the `README` file of the DRAEM model ([here](https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/models/draem/README.md?plain=1#L11)). We could make it a bit more explicit and mention that this is needed to match the performance of the original paper.

Oh, I didn't check the readme, thanks for pointing that out. I think the current description is sufficient.

src/anomalib/data/utils/augmenter.py
blaz-r: This makes sense yes. I changed this based on [original implementation](https://github.com/VitjanZ/DRAEM/blob/main/data_loader.py#L126), but as you said this can be covered by simply setting the range to [0, 0.8].
I will revert this, while I think that it would be good idea to make it part of config.

src/anomalib/models/draem/lightning_model.py
blaz-r: From a few training runs I did, the last few epochs make a little difference, but it is noticeable. With current design, removing early stopping would also require changing of this function:
https://github.com/openvinotoolkit/anomalib/blob/66532fc1625020ce7ff6f3fc1565663aa540e834/src/anomalib/models/draem/lightning_model.py#L132-L146
Should I then remove it from config and also add an if statement here to check if it's present or should it be removed altogether.

src/anomalib/data/utils/augmenter.py
blaz-r: This is now reverted.

src/anomalib/models/draem/lightning_model.py
djdameln: I think the user should have the possibility to enable early stopping if they want to, so let's add a check

blaz-r: Okay, this is now implemented.

src/anomalib/data/utils/augmenter.py
djdameln: > I think that it would be good idea to make it part of config.

I agree. Would be good if you could add it to this PR. Just be aware that there is a small discrepancy between the paper and the original implementation. The paper defines beta as the weight of the anomalous perturbations and proposes a beta range of [0.1, 1.0]. The implementation on the other hand, samples beta from the [0, 0.8] range and then uses (1-beta) as the weight of the perturbations, which effectively results in a [0.2, 1.0] range (which is the same range that we were using in Anomalib).

I would propose to use the numbers mentioned in the paper for the default configuration

blaz-r: I now added this to config.

Issue #1144, opened 2023-06-19
📚 Improve documentation on using different backbones


ninafiona:
### What is the motivation for this task?

I want to try different backbones (EfficientNet, ViT, MobileNet, ...), but there is not much documentation on it. 
I tried this configuration, but `blocks.4` was not recognized. 
```
model:
  name: patchcore
  backbone: efficientnet_b4  # wide_resnet50_2, resnet18
  pre_trained: true
  layers:
    # - layer2
    # - layer3
    - blocks.4
    - blocks.5
```

### Describe the solution you'd like

I would like information about which backbones are compatible and how to reference the mid-level feature layers.

callzwl:
Have you solved the problem? Please answer me if it's resolved. Thank you

samet-akcay:
Hi everyone, we have now added some tutorials to v1 documentation. 

v1 itself is not officially on PyPI yet, but you could try it out by following the link here
https://github.com/openvinotoolkit/anomalib?tab=readme-ov-file#-upcoming-release-anomalib-v1-

I'm closing this for now. Feel free to re-open if you think that the documentation is still not clear or needs some more work. Cheers!

PR #1443, opened 2023-10-26
Extend supported models in TimmFeatureExtractor

### LINKED ISSUE #1442: [Bug]: Not able to use Mobilenet V3 as a backbone for CFlow despite being used in paper
### Describe the bug

The [CFlow paper](https://arxiv.org/pdf/2107.12571.pdf) uses resnet variants as well as Mobilenet V3 as backbones. However, anomalib does not appear to work when trying to change the backbone to Mobilenet v3. I tried changing the backbone to the timm feature extractor "mobilenetv3_large_100" but no configuration of the layers (blocks) seems to work. It appears to be tied to how the layers are exposed for resnet and how mobilenet v3 uses blocks in the timm feature extraction implementation.

There is an issue [logged here](https://github.com/openvinotoolkit/anomalib/issues/1144) that has displayed similar problems

### Dataset

N/A

### Model

CFlow

create a yaml config containing "mobilenetv3_large_100" as the backbone. Try various forms of blocks (e.g. "blocks", "blocks.x.y") in the layers.

### OS information

OS information:
- OS: Ubuntu 22.04
- Python version: 3.8
- Anomalib version: [e.g. 0.3.6]
- PyTorch version: [e.g. 1.9.0]
- CUDA/cuDNN version: [e.g. 11.1]
- GPU models and configuration: [e.g. 2x GeForce RTX 3090]
- Any other relevant information: [e.g. I'm using a custom dataset]

### Expected behavior

The paper of CFlow has shown good results with MobilenetV3. Ideally this would work and allow a much smaller model that can run faster on limited hardware

### Screenshots

_No response_

### Pip/GitHub

pip

_No response_

### Configuration YAML

```yaml
N/A
```

### Logs

```shell
N/A
```

### Code of Conduct

blaz-r:
## Description

This PR extends supported backbones inside TimmFeatureExtractor. I refactored the function that maps layer names to feature IDs, so it now also supports models that don't have layer names inside `named_children` dict.

- Fixes #1442 

As requested in #1144 I think that documentation would need to be extended with some more info on backbones.

## Changes

<details>

- [ ] Documentation update

</details>

## Checklist

<details>

</details>


samet-akcay (approved):

saifvazir:
Hi @blaz-r thanks for this PR! It helped me test with a few backbones that I couldn't do previously. 
I have a small question. I'm trying to use [ViT base](https://huggingface.co/timm/vit_base_patch14_dinov2.lvd142m) from timm as a feature extractor for patchcore/other models. Since these models don't support `features_only` flag while using the `create_model` function, this PR would fail in that case. I know it might be out of scope for the current PR, but just wanted to bring this up in case we can make it more extendable. I'm working on it and will update if I find something that works for transformer models too.

blaz-r:
Hi @saifvazir, nice to see you find this useful. Thanks for pointing out the problem with transformer based models. 
There are some used in the Fastflow model if you want to see how to use them. As you said `features_only` is omitted as well as `return_indices`. Maybe we could add an option inside the `layers` of config for something like `final_layer` which would indicate to avoid using the two parameters above.

saifvazir:
Yes, I took a look at the the Fastflow model and I found that we've sort of hardcoded the allowed transformer model feature extractors in `torch_model.py`. I feel that this is kind of restricting given the amount of ViT models available in timm. Maybe requires a separate discussion thread on how to handle transformer models as feature extractors. However, thanks for putting up this PR and replying :)

blaz-r:
Glad to see that you find this PR useful. While we maybe could use a separate discussion thread, I think it's still a possibility to extend this PR as it technically covers this exact thing. 

Maybe we could specify a special parameter for `layers` in config, that would indicate that the final layer should be used.

@ashwinvaidya17 is this worth implementing, since timm is supposed to be deprecated?

ashwinvaidya17:
@blaz-r I think there is a bit of confusion due to my last PR. We are deprecating the class `FeatureExtractor` in favour of explicit `TimmFeatureExtractor`. This is because we have both TorchFX and Timm. Actually, if it isn't too much work can you target this PR to the v1 branch? The CI in main is failing due to which a lot of PRs are now outstanding.

blaz-r:
Okay. I'll change this to v1.

blaz-r:
Well that was a bit messy, but it should be ready for v1.0 now.

blaz-r:
The question we discussed with @saifvazir still remains. What should we do with ViT like models?

saifvazir:
Hi @blaz-r , I might be wrong here, but can we use `torchfx` to load the backbone weights directly and then pass in the layer names to extract the relevant features? I see that TorchFx is just being used in the `CSFlowModel` as of now with the Multi Scale extractor. I think we could make it more generalizable and use it directly by specifying something in the config file? @ashwinvaidya17 since you were the author for TorchFx, what are your comments on this idea?

blaz-r:
@saifvazir sorry for the late reply. There is TorchFX feature extractor implemented, so new models can already use that. But many of the current models use Timm. Having this as part of the config would probably introduce some problems, and I think it just unnecessary complexion to code and usage of library.

samet-akcay:
@blaz-r is this pr still up to date and relevant? if yes, I'll merge it

blaz-r:
@samet-akcay yes, it is still relevant. I just merged the new additions of v1, if all the checks pass it should be ready to merge.

PR #1483, opened 2023-11-17
Reorg Part I: Data

samet-akcay:
## Description

- Part I of this reorg is to re-structure the data subpackage.


src/anomalib/utils/types/task_type.py
```diff
@@ -1,6 +1,6 @@
 """Task type enum."""
 
-# Copyright (C) 2022 Intel Corporation
+# Copyright (C) 2022-2023 Intel Corporation
```
ashwinvaidya17, line 3:
```suggestion
# Copyright (C) 2023 Intel Corporation
```

ashwinvaidya17 (commented):
Thanks. Just one minor comment

src/anomalib/utils/types/task_type.py
samet-akcay: as far as I know this is the suggested way when we create the file earlier and modify it later on.

ashwinvaidya17: Ah didn't know. Let's follow this convention from now on then.

ashwinvaidya17 (approved):

PR #1485, opened 2023-11-18
Reorg Part II: Remove `pre_processor` and `post_processor` subpackages

samet-akcay:
## Dependency  ⚠️
- Please review and merge #1483 first

## Description

- Remove  `pre_processor` and `post_processor` subpackages
- Move the subpackages and modules to the corresponding subpackages

## Discussion ❓
- Where exactly `transforms should go? 
- `anomalib.data.utils.transforms` (current) or `anomalib.data.transforms` ?


src/anomalib/data/utils/transforms.py
```diff
@@ -162,3 +164,86 @@ def get_transforms(
         transforms = A.Compose(transforms_list, additional_targets={"image": "image", "depth_image": "image"})
 
     return transforms
+
+
+class Denormalize:
```
djdameln, line 169: Is this still used? I think it's been deprecated for a while so we could probably remove it (after confirming that it's not used anywhere of course)

samet-akcay: It is used in some of the jupyter notebooks

ashwinvaidya17 (approved):
Thanks

djdameln (approved):
Thanks!

PR #1487, opened 2023-11-19
Reorg Part III: Move the main anomalib components from `anomalib.utils` to `anomalib`

samet-akcay:
## Description
- Move the main anomalib components from `anomalib.utils` to `anomalib`. This includes components such as metrics, loggers, callbacks, hpo, sweep and benchmarking


src/anomalib/utils/config/config.py
djdameln, line 1: As discussed, should this be like `src/anomalib/utils/config.py`?

samet-akcay: yes, you are right. Fixed in #bdce87e01f9917c2ad4ecc5d93a39cc206c27ba6

ashwinvaidya17 (approved):
Thanks

PR #1503, opened 2023-11-27
v1: Update the readme file

samet-akcay:
## Description
This pr updates the readme file with the following changes.
- Re-organize the sections
- Add CLI and API options. 
- Replace the readme figure with a visa image.
- Add emojis to the sections.


README.md
```diff
@@ -14,395 +14,245 @@
...
+<summary>Inference via CLI</summary>
 
-- Papers With Code has an interface to easily browse models available in timm: [https://paperswithcode.com/lib/timm](https://paperswithcode.com/lib/timm)
+```bash
+# To get help about the arguments, run:
+anomalib predict -h
 
-- You can also find them with the function `timm.list_models("resnet*", pretrained=True)`
+# Predict via providing the model name
+anomalib predict --model anomalib.models.Patchcore
```
djdameln, line 149: This does not work. It complains about `--data` arg missing. I also think we should provide a `--ckpt_path` arg here, because `anomalib predict` is only useful when we specify a previously trained model.

README.md
```diff
@@ -14,395 +14,245 @@
...
+# Predict by overriding the various values with commands.
+anomalib predict --model anomalib.models.Patchcore --data <CONFIG | CLASS_PATH_OR_NAME>
 
-```yaml
-model:
-  name: cflow
-  backbone: wide_resnet50_2
-  pre_trained: true
+# Predict by using a config file.
+anomalib predict --config <config_file_path> --return_predictions
```
djdameln, line 155: If we give multiple examples here, maybe we should do the same for `anomalib train` above

README.md
```diff
@@ -14,395 +14,245 @@
...
 [![Downloads](https://static.pepy.tech/personalized-badge/anomalib?period=total&units=international_system&left_color=grey&right_color=green&left_text=PyPI%20Downloads)](https://pepy.tech/project/anomalib)
 
 </div>
 
 ---
 
-# Introduction
+# 👋 Introduction
 
 Anomalib is a deep learning library that aims to collect state-of-the-art anomaly detection algorithms for benchmarking on both public and private datasets. Anomalib provides several ready-to-use implementations of anomaly detection algorithms described in the recent literature, as well as a set of tools that facilitate the development and implementation of custom models. The library has a strong focus on image-based anomaly detection, where the goal of the algorithm is to identify anomalous images, or anomalous pixel regions within images in a dataset. Anomalib is constantly updated with new algorithms and training/inference extensions, so keep checking!
```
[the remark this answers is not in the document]
samet-akcay: thanks for spotting this! It was on my to-do list, but slipped my mind

README.md
samet-akcay: @ashwinvaidya17 ?

README.md
samet-akcay: what would be your preference here? Remove the second example here, or add a second example to the other sections?

PR #1510, opened 2023-11-29
Fix result image URLs

f0k:
## Description

The sample result images in the README files for several of the models did not display.

## Changes

The wrong image URLs were corrected. I opted to stick with relative paths as this makes more sense in case of a fork, for example. I did not touch the existing absolute paths.

- [x] Documentation update


ashwinvaidya17 (approved):
Thanks for the fix!

PR #1514, opened 2023-11-30
Modify `Engine.predict`

ashwinvaidya17:
## Description

- Predict now accepts path to image/folder `--data <path-to-image>`
- Extend model tests
- Depends on https://github.com/openvinotoolkit/anomalib/pull/1509

## Changes

<details>

- [ ] Documentation update

</details>

## Checklist

<details>

</details>


djdameln (commented):
- I think it would be good to also allow the following:
`anomalib --model Padim --data MVTec --ckpt_path <path/to/model.ckpt>`.
In which case anomalib would generate the predictions on the MVTec test set. I think this would be a simple change as we only have to implement `predict_dataloader()` in the base datamodule.

- If I omit the `--data` argument, I get a not very descriptive error: `NSKeyError: Key "data" not found in namespace.`. I would expect to see some message here indicating that passing the `--data` arg is mandatory. (I also noticed this earlier for the other entrypoints like `anomalib fit`, `anomalib test` etc.)

- We could consider making the `--ckpt_path` argument mandatory, as I do not see a use case where a user would want to predict using a non-trained model.

samet-akcay (commented):
thanks for the change, overall looks great. Apart from that I've got some questions for certain parts.

tools/inference/lightning_inference.py
```diff
@@ -57,19 +57,17 @@ def infer(args: Namespace) -> None:
...
-    model = get_model(args.model)
     engine.predict(model=model, dataloaders=[dataloader], ckpt_path=args.ckpt_path)
 
 
 if __name__ == "__main__":
     parser = get_parser()
     config = parser.parse_args()
-    args = parser.instantiate_classes(config)
-    infer(args)
+    infer(config)
```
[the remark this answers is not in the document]
ashwinvaidya17: I introduced it earlier because I was using the objects directly that were returned. But I changed the design so I reverted it. Plus, since we use the API in this method, it now instantiates the objects in the infer method. One of those lock step changes with API and CLI

src/anomalib/cli/cli.py
```diff
@@ -185,20 +195,21 @@ def add_export_arguments(self, parser: LightningArgumentParser) -> None:
...
-        self.add_default_arguments_to_parser(parser)
 
     def add_hpo_arguments(self, parser: LightningArgumentParser) -> None:
         """Add hyperparameter optimization arguments."""
         parser = get_hpo_parser(parser)
 
     def add_benchmark_arguments(self, parser: LightningArgumentParser) -> None:
         """Add benchmark arguments to the parser."""
-        parser.add_argument("--config", type=Path, help="Path to the benchmark config.", required=True)
+        parser.add_argument("--benchmark_config", type=Path, help="Path to the benchmark config.", required=True)
```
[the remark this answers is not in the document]
samet-akcay: why we change `config` to `benchmark_config`? from the first glance it looks a bit verbose to me. I might prefer `config`

tools/inference/lightning_inference.py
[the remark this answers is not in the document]
samet-akcay: maybe we could still use `args`? It might look a bit more consistent with the name of the function `parse_args`? Thoughts?

src/anomalib/cli/cli.py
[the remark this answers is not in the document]
ashwinvaidya17: It conflicts with the LightningArgparse's config command. I am not sure what's the best approach here. I can revert this to config but this might mean some additional checks that might pollute the implementation. Once we have pipelines this command might change. In that case should we even support hpo and benchmarking from the cli now? I am expecting the new command to be `anomalib pipelines benchmark ...` and `anomalib pipelines hpo ...`

djdameln:
- In the latest version, `anomalib predict --model Padim --data <path/to/image/folder> --ckpt_path <path/to/ckpt>` fails with the following error:
`AttributeError: 'AnomalibCLI' object has no attribute '_set_predict_dataloader'`. I guess it's because `set_predict_dataloader` was changed to `set_predict_dataloader_namespace` but not updated in [line 221 here](https://github.com/ashwinvaidya17/anomalib/blob/ashwin/fix_predict/src/anomalib/cli/cli.py#L221). 
- I think it would be good to extend the cli tests with the different ways of passing data (data class, folder of images, single image) so that our CI catches bugs like the above.

> We could consider making the --ckpt_path argument mandatory, as I do not see a use case where a user would want to predict using a non-trained model.

I'm afraid we need to reverse this, because there's a very obvious use-case which I didn't consider earlier: Zero-shot models! My mistake, sorry

ashwinvaidya17:
> * In the latest version, `anomalib predict --model Padim --data <path/to/image/folder> --ckpt_path <path/to/ckpt>` fails with the following error:
>   `AttributeError: 'AnomalibCLI' object has no attribute '_set_predict_dataloader'`. I guess it's because `set_predict_dataloader` was changed to `set_predict_dataloader_namespace` but not updated in [line 221 here](https://github.com/ashwinvaidya17/anomalib/blob/ashwin/fix_predict/src/anomalib/cli/cli.py#L221).
> * I think it would be good to extend the cli tests with the different ways of passing data (data class, folder of images, single image) so that our CI catches bugs like the above.
> 
> > We could consider making the --ckpt_path argument mandatory, as I do not see a use case where a user would want to predict using a non-trained model.
> 
> I'm afraid we need to reverse this, because there's a very obvious use-case which I didn't consider earlier: Zero-shot models! My mistake, sorry

Done!

djdameln:
In addition to running predict on a folder of images, I think we should allow running it on a single image. but the following command results in an error atm:
`anomalib predict --model Padim --data datasets/mvtec/bottle/test/broken_large/000.png`

tests/integration/cli/test_cli.py
```diff
@@ -13,210 +13,189 @@
...
                     project_path,
                 ),
                 "--ckpt_path",
-                f"{project_path}/{model_name}/dummy/weights/last.ckpt",
+                f"{project_path}/padim/dummy/weights/last.ckpt",
+            ],
+        )
+        torch.cuda.empty_cache()
+
+    def test_predict_with_image_paths(self, project_path: Path) -> None:
```
djdameln, line 112: Let's add a similar test for predicting on a single image

tests/integration/cli/test_cli.py
```diff
@@ -13,210 +13,189 @@
...
+            args=[
+                "predict",
+                "--data",
+                f"{project_path}/datasets/visa_pytorch/dummy/test/bad",
+                *self._get_common_cli_args(
+                    None,
+                    project_path,
+                ),
+                "--ckpt_path",
+                f"{project_path}/padim/dummy/weights/last.ckpt",
```
djdameln, line 131: in line with the recent change we can omit the --ckpt path here

src/anomalib/engine/engine.py
```diff
@@ -14,13 +14,14 @@
...
 from omegaconf import DictConfig, ListConfig
+from torch.utils.data import DataLoader, Dataset
 
 from anomalib.callbacks import get_visualization_callbacks
 from anomalib.callbacks.metrics import _MetricsCallback
 from anomalib.callbacks.normalization import get_normalization_callback
 from anomalib.callbacks.post_processor import _PostProcessorCallback
 from anomalib.callbacks.thresholding import _ThresholdCallback
-from anomalib.data import AnomalibDataModule, AnomalibDataset
+from anomalib.data import AnomalibDataModule, AnomalibDataset, InferenceDataset
```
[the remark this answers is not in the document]
ashwinvaidya17: Yeap. I already have a PR for this https://github.com/openvinotoolkit/anomalib/pull/1544

src/anomalib/engine/engine.py
```diff
@@ -349,14 +350,14 @@ def test(
...
-            self._setup_dataset_task(dataloaders)
+        self._setup_dataset_task(dataloaders)
         return self.trainer.test(model, dataloaders, ckpt_path, verbose, datamodule)
 
     def predict(
         self,
         model: AnomalyModule | None = None,
         dataloaders: EVAL_DATALOADERS | AnomalibDataModule | None = None,
-        datamodule: AnomalibDataModule | None = None,
+        datamodule: AnomalibDataModule | Dataset | InferenceDataset | None = None,
```
samet-akcay, line 360: for a future reference, it would be nice if we use type variables for our main components and store them in like `anomalib.utils.types`

src/anomalib/engine/engine.py
```diff
@@ -403,7 +405,22 @@ def predict(
...
+
+        if datamodule is not None and isinstance(datamodule, Dataset):
+            dataloader = DataLoader(datamodule)
+            datamodule = None
+            if dataloaders is None:
+                dataloaders = dataloader
+            elif isinstance(dataloaders, DataLoader):
+                dataloaders = [dataloaders, dataloader]
+            else:  # dataloader is a list
+                dataloaders = [*dataloaders, dataloader]
```
samet-akcay, line 420: I've recently created this #1576. Would this part resolve that issue?

samet-akcay (commented):
Thanks @ashwinvaidya17. Looks good to me overall. Just had some questions

src/anomalib/engine/engine.py
```diff
@@ -390,7 +392,7 @@ def predict(
         CLI Usage:
             1. you can pick a model.
                 ```python
-                anomalib predict --model anomalib.models.Padim
+                anomalib predict --model Padim --data datasets/MVTec/bottle/test/broken_large
```
samet-akcay, line 395: this is how the example appears in the documentation:
image

Would it be en idea to use linebreaks for each arg for a better look?

```python
"""
        CLI Usage:
            1. you can pick a model.
                ```python
                anomalib predict --model anomalib.models.Padim
                anomalib predict --model Padim \
                                 --data datasets/MVTec/bottle/test/broken_large
                ```
            2. Of course, you can override the various values with commands.
                ```python
                anomalib predict --model anomalib.models.Padim \
                                 --data <CONFIG | CLASS_PATH_OR_NAME>
                ```
            4. If you have a ready configuration file, run it like this.
                ```python
                anomalib predict --config <config_file_path> --return_predictions
                ```
"""

```

src/anomalib/engine/engine.py
ashwinvaidya17: Yeah. I think we should refactor all these long types. I've taken a note

src/anomalib/engine/engine.py
ashwinvaidya17: No, this only handles the instance when a dataset is passed to the predict method. I've added a comment now to inform this.

samet-akcay (approved):
thanks

djdameln (commented):
There seems to be an issue with the checkpoint loading.

When I call this:
`anomalib predict --model Padim --data MVTec --ckpt_path model.ckpt`
I get the following warning:
`WARNING  ckpt_path is not provided. Model weights will not be loaded.`

PR #1534, opened 2023-12-08
OpenVINO NNCF updates

adrianboguszewski:
## Description

This PR is a part of bigger effort to fully support OpenVINO in Anomalib. This PR is focused on updating OpenVINO and NNCF.

- OpenVINO version >= 2023.0, NNCF version >= 2.5.0
- Default device changed to AUTO
- Updated usage of OV
- Export function tests

## Checklist

<details>

- [x] I have made corresponding changes to the documentation (If applicable)
- [x] I have added tests that prove my fix is effective or that my feature works (If applicable)

</details>


adrianboguszewski:
@samet-akcay @djdameln @ashwinvaidya17 it's ready for your review :)

tests/integration/model/test_models.py
```diff
@@ -18,11 +18,90 @@
 from anomalib.utils.types import TaskType
 
 
-def models() -> list[str]:
+def models() -> set[str]:
     """Return all available models except ai_vad."""
     return get_available_models()
 
 
+def _get_objects(
```
ashwinvaidya17, line 26: Any particular reason this is moved out of the class?

tests/unit/deploy/test_export.py
```diff
@@ -0,0 +1,48 @@
+"""Tests for exporting."""
```
ashwinvaidya17, line 1: I don't think we need this test. We already add `engine.export` in one of the open PRs

tests/integration/model/test_models.py
adrianboguszewski: To use as a utility in export tests

tests/unit/deploy/test_export.py
adrianboguszewski: Which one? In this case, I'm not testing the export function but the ability to export all models to OpenVINO. All of them must work in OV. Should I move this test to another directory, then?

ashwinvaidya17: The problem is that currently the video models cannot be exported, and some other models have incorrect graphs due to frozen ops, unrolled for-loops based on example tensor dimensions, and flattened conditionals. While ideally we would like to support OpenVINO conversion for all the models, it might require some extra effort that might be out of scope for this PR. That's why I feel we should drop this test. Also, personally I feel unit tests should not test for export coverage. It should just test the particular method. We can include a test for export coverage later in integration tests or nightly. Maybe other have a different opinion.

tests/integration/model/test_models.py
ashwinvaidya17: In this particular case, if `_get_objects` needs to be used by other tests it would be better to move it to `conftest.py`
But I would suggest waiting till we merge https://github.com/openvinotoolkit/anomalib/pull/1514 as there is some overlapping work here.

tests/unit/deploy/test_export.py
adrianboguszewski: I'm gonna move all OpenVINO export tests to a separate PR.

adrianboguszewski: #1561

src/anomalib/deploy/inferencers/openvino_inferencer.py
```diff
@@ -37,11 +37,12 @@ class OpenVINOInferencer(Inferencer):
...
             metadata.
             Defaults to ``None``.
-        device (str | None, optional): Device to run the inference on.
-            Defaults to ``CPU``.
+        device (str | None, optional): Device to run the inference on (AUTO, CPU, GPU, NPU).
+            Defaults to ``AUTO``.
         task (TaskType | None, optional): Task type.
             Defaults to ``None``.
-
+        config (dict | None, optional): Configuration parameters for the inference
```
samet-akcay, line 44: Would we still need the `config` in the new approach? We ideally would like to get rid of `config` since the dictionary hides the arguments in the class signature.

tests/unit/deploy/test_inferencer.py
```diff
@@ -110,7 +110,7 @@ def test_openvino_inference(task: TaskType, ckpt_path: Callable[[str], Path], da
     engine = Engine(task=task)
     export_path = ckpt_path("Padim").parent.parent
     datamodule = MVTec(root=dataset_path / "mvtec", category="dummy")
-    engine.export(
+    real_exported_path = engine.export(
```
samet-akcay, line 113: Looks like we don't use `export_path` from Line 111 anymore. We could remove it and rename `real_exported_path` to `export_path`?

samet-akcay (changes requested):
@adrianboguszewski, thanks for the PR! I only have two minor comments.

src/anomalib/deploy/inferencers/openvino_inferencer.py
adrianboguszewski: I don't want to break the current behavior in this PR. Reworking inference config is on my list for future work.

tests/unit/deploy/test_inferencer.py
adrianboguszewski: Not really. `export_path` points to the directory and is used in the export function. `real_exported_path` is returned by a function and indicates the .xml file

adrianboguszewski: I hope I understood you correctly

samet-akcay: yes, that's what I want. Perhaps, we could also remove Line 111 now? This one ?

```python 
export_path = ckpt_path("Padim").parent.parent
```

samet-akcay (approved):

PR #1575, opened 2024-01-03
WinCLIP Model implementation

djdameln:
## Description

This PR adds the [WinCLIP](https://arxiv.org/pdf/2303.14814.pdf) zero-/few-shot anomaly detection model.

WinCLIP is a zero-shot model, so no training or fine-tuning on normal images is required. This means that the interaction with the model is slightly different compared to other models:

- We can immediately evaluate the model on a test set without running a training sequence first, e.g:
`anomalib test --model WinClip --data MVTec --data.category <category> --data.image_size 240 --data.normalization clip`
This will evaluate WinCLIP on the test set of the specified MVTec category. Before testing, Anomalib will run the validation sequence to collect the normalization and thresholding statistics.

## Changes

- Adds lightning model, torch model and utils for WinCLIP model.
- In `engine.predict`, adds a check if a checkpoint patch is specified by the user. If not, we first run the validation sequence to collect normalization and thresholding stats.
- Adds CLIP normalization statistics for proper input normalization when using pre-trained CLIP models.

## Known Issues

- Results are slightly lower than the paper. This is mainly because we use Albumentations transforms while the CLIP backbone was pretrained using torchvision transforms. This will be addressed in a future PR replacing Albumentations with TorchVision throughout the library.
- Benchmarking results are not reported because of the previous point.
- WIP: unit and integration tests

## Checklist

<details>

</details>


tests/integration/model/test_models.py
```diff
@@ -19,8 +19,8 @@
 
 
 def models() -> list[str]:
-    """Return all available models except ai_vad."""
-    return get_available_models()
+    """Return all available models except winclip."""
```
samet-akcay, line 22: for a future reference, can you also explain why?

samet-akcay (changes requested):
@djdameln, great work! The code is really clean and flows really well! I've got some comments.

Also, can you add the model to the documentation? You need to do the following:
- Create a `win_clip.md` file in [here](https://github.com/openvinotoolkit/anomalib/tree/v1/docs/source/markdown/guides/reference/models/image).
- Refer to this md file [here](https://raw.githubusercontent.com/openvinotoolkit/anomalib/v1/docs/source/markdown/guides/reference/models/image/index.md).

src/anomalib/models/image/winclip/torch_model.py
```diff
@@ -0,0 +1,282 @@
...
+
+        # initialize CLIP model
+        self.clip = open_clip.create_model(self.backbone, pretrained=self.pretrained)
+        self.clip.visual.output_tokens = True
+        self.grid_size = self.clip.visual.grid_size
+
+        self.masks: list[torch.Tensor] | None = None
+        self.text_embeddings: torch.Tensor | None = None
+        self.visual_embeddings: list[torch.Tensor] | None = None
+        self.patch_embeddings: torch.Tensor | None = None
```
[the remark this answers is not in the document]
djdameln: Previously pylint would complain about this, but it seems that it ruff does allow assigning the attributes outside of the init.

src/anomalib/engine/engine.py
```diff
@@ -350,6 +350,10 @@ def test(
         if model:
             self._setup_trainer(model)
             self._setup_dataset_task(dataloaders)
+        if ckpt_path is None:
```
ashwinvaidya17, line 353: What if someone calls the API like this
```python
model = Padim()
engine.fit(model, train_dataloader)
engine.test(model, test_dataloader)
```
Maybe we need to already consider the "model type" in the design. Or maybe extend `is_fitted` to all the models

src/anomalib/models/image/winclip/lightning_model.py
```diff
@@ -0,0 +1,133 @@
...
+        """Set model-specific trainer arguments.
+
+        Sets the number of epochs to 1 and the number of training batches to 1. This is needed to ensure that lightning
+        runs the validation loop when calling ``engine.fit()`` or ``engine.train()`` so that the normalization and
+        thresholding statistics are computed.
+        """
+        return {
+            "max_epochs": 1,
+            "limit_train_batches": 1,
+        }
```
ashwinvaidya17, line 133: Maybe if we introduce a property like 
```python
@property
def model_type(self):
    return ModelType.ZEROSHOT
```
then we can just check this in engine.fit and call engine.validate instead. But these changes might be the scope of another PR

ashwinvaidya17 (commented):
Nice! I am excited for this PR. I have two minor comments. But probably addressing them would require a separate PR

src/anomalib/models/image/winclip/lightning_model.py
samet-akcay: would it be the model type or learning type? The same model can also be few-show, I think

src/anomalib/models/image/winclip/README.md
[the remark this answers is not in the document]
djdameln: Would be nice if we could re-use the same markdown file to keep the maintenance effort low. How do we do this for other models?

src/anomalib/models/image/winclip/torch_model.py
```diff
@@ -0,0 +1,282 @@
...
+        text_embeddings (torch.Tensor | None): The text embeddings for the compositional prompt ensemble.
+        visual_embeddings (list[torch.Tensor] | None): The multiscale embeddings for the reference images.
+        patch_embeddings (torch.Tensor | None): The patch embeddings for the reference images.
+    """
+
+    def __init__(self, k_shot: int = 0, scales: tuple = (2, 3)) -> None:
+        super().__init__()
+        self.backbone = BACKBONE
+        self.pretrained = PRETRAINED
+        self.temperature = TEMPERATURE
```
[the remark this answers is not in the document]
djdameln: These will be made configurable once we sort out the torchvision transforms. Let's keep it like this until then.

src/anomalib/models/image/winclip/lightning_model.py
ashwinvaidya17: Good point. Maybe we should discuss a design for this.

djdameln: I was thinking of a design similar to this. We could name the property as `learning_strategy` and set it dynamically based on the class and configuration. e.g.
```python
  @property
  def learning_strategy(self):
      return LearningStrategy.FEW_SHOT if self.k_shot > 0 else LearningStrategy.ZERO_SHOT
```
Note that we technically do not need the distinction between `FEW_SHOT` and `ZERO_SHOT` for the WinCLIP model, but this might change in the future once we start adding more zero-/few-shot models that may have different requirements.

For all other models currently in the library we would use
```python
  @property
  def learning_strategy(self):
      return LearningStrategy.ONE_CLASS
```

src/anomalib/models/components/base/buffer_list.py
```diff
@@ -0,0 +1,108 @@
...
+            [
+                tensor([10., 10., 10.], device='cuda:0'),
+                tensor([11., 11., 11.], device='cuda:0'),
+                tensor([12., 12., 12.], device='cuda:0')
+            ]
+        """
+        for i, value in enumerate(values):
+            self.register_buffer(f"_{name}_{i}", value, persistent=persistent, **kwargs)
+
+        setattr(BufferListMixin, name, BufferListDescriptor(name, len(values)))
```
ashwinvaidya17, line 65: What's the advantage of setting this to `BufferListMixin` rather than `self`?

src/anomalib/models/image/winclip/torch_model.py
```diff
@@ -0,0 +1,401 @@
...
+BACKBONE = "ViT-B-16-plus-240"
+PRETRAINED = "laion400m_e31"
+TEMPERATURE = 0.07  # temperature hyperparameter from the clip paper
+
+
+class WinClipModel(DynamicBufferModule, BufferListMixin, nn.Module):
+    """PyTorch module that implements the WinClip model for image anomaly detection.
+
+    Args:
+        k_shot (int, optional): The number of reference images used for few-shot anomaly detection.
```
ashwinvaidya17, line 30: These might be outdated.

ashwinvaidya17 (commented):
Thanks. I don't have any major comments

src/anomalib/models/components/base/buffer_list.py
djdameln: It's because descriptors need to be assigned to the class, not the instance, in order to work properly.
see https://stackoverflow.com/questions/12599972/descriptors-as-instance-attributes-in-python
and https://docs.python.org/3/howto/descriptor.html#closing-thoughts

src/anomalib/engine/engine.py
```diff
@@ -161,12 +164,63 @@ def trainer(self) -> Trainer:
...
+        Returns:
+            Optional[_ThresholdCallback]: Threshold callback, if available.
+
+        Raises:
+            ValueError: If there are multiple threshold callbacks.
+        """
+        callbacks = [callback for callback in self.trainer.callbacks if isinstance(callback, _ThresholdCallback)]
+        if len(callbacks) > 1:
+            msg = f"Trainer can only have one thresholding callback but multiple found: {callbacks}. Please check your \
+                configuration. Exiting to avoid unexpected behavior."
```
samet-akcay, line 212:
```suggestion
            msg = (
                f"Trainer can only have one thresholding callback but multiple found: {callbacks}. "
                "Please check your configuration. Exiting to avoid unexpected behavior."
            )
```

src/anomalib/engine/engine.py
```diff
@@ -161,12 +164,63 @@ def trainer(self) -> Trainer:
...
+        Returns:
+            Optional[_MinMaxNormalizationCallback]: Normalization callback, if available.
+
+        Raises:
+            ValueError: If there are multiple normalization callbacks.
+        """
+        callbacks = [callback for callback in self.trainer.callbacks if isinstance(callback, NormalizationCallback)]
+        if len(callbacks) > 1:
+            msg = "Trainer can only have one normalization callback but multiple found: {}. Please check your \
+                configuration. Exiting to avoid unexpected behavior.".format(callbacks)
```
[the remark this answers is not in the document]
samet-akcay: This is to follow the coherency for the rest of the codebase.

src/anomalib/engine/engine.py
samet-akcay: same as above

src/anomalib/engine/engine.py
```diff
@@ -213,6 +267,46 @@ def _setup_anomalib_callbacks(self) -> None:
...
+        We need to run validation if the model is configured with normalization enabled, but no normalization metrics
+        have been collected yet. Similarly, we need to run validation if the model is configured with adaptive
+        thresholding enabled, but no thresholds have been computed yet.
+
+        We can only run validation if we have validation data available, so we check if the dataloaders or datamodule
+        are available. If neither is available, we can't run validation.
+
+        Args:
+            model (AnomalyModule): Model passed to the entrypoint.
+            dataloaders (EVAL_DATALOADERS | AnomalibDataModule | None): Dataloaders passed to the entrypoint.
```
samet-akcay, line 291: I guess we'll address this `AnomalibDataModule` annotation later, as we discussed. It looks a bit confusing.

src/anomalib/engine/engine.py
```diff
@@ -334,6 +435,30 @@ def test(
...
+            >>> from anomalib.engine import Engine
+            >>> from anomalib.models import Padim
+            >>> from anomalib.data import MVTec
+            >>> engine = Engine()
+            >>> model = Padim()
+            >>> datamodule = MVTec()
+            >>> model.learning_type
+            <LearningType.ONE_CLASS: 'one_class'>
+            >>> engine.fit(model, datamodule=datamodule)
+            >>> engine.test(model, datamodule=datamodule)
```
samet-akcay, line 449: Just to make it look better on the sphinx build
```suggestion
            # fit and test a one-class model
            >>> from anomalib.data import MVTec            
            >>> from anomalib.models import Padim            
            >>> from anomalib.engine import Engine

            >>> datamodule = MVTec()
            >>> model = Padim()            
            >>> model.learning_type
            <LearningType.ONE_CLASS: 'one_class'>

            >>> engine = Engine()
            >>> engine.fit(model, datamodule=datamodule)
            >>> engine.test(model, datamodule=datamodule)
```

src/anomalib/engine/engine.py
```diff
@@ -334,6 +435,30 @@ def test(
...
+            # Test a zero-shot model
+            >>> from anomalib.engine import Engine
+            >>> from anomalib.models import Padim
+            >>> from anomalib.data import MVTec
+            >>> engine = Engine()
+            >>> model = Padim()
+            >>> datamodule = MVTec(image_size=240, normalization="clip")
+            >>> model.learning_type
+            <LearningType.ZERO_SHOT: 'zero_shot'>
+            >>> engine.test(model, datamodule=datamodule)
```
samet-akcay, line 460:
```suggestion
            >>> from anomalib.data import MVTec
            >>> from anomalib.models import Padim            
            >>> from anomalib.engine import Engine
  
            >>> datamodule = MVTec(image_size=240, normalization="clip")
            >>> model = Padim()            
            >>> model.learning_type
            <LearningType.ZERO_SHOT: 'zero_shot'>
            
            >>> engine = Engine()            
            >>> engine.test(model, datamodule=datamodule)
```

src/anomalib/engine/engine.py
```diff
@@ -348,11 +473,19 @@ def test(
                 anomalib test --config <config_file_path>
                 ```
         """
+        assert (
+            model or self.model
+        ), "`Engine.test()` requires an `AnomalyModule` when it hasn't been passed in a previous run."
```
samet-akcay, line 478: would it be possible to use exception handling here instead of assert. Bandit complains about this stuff for the security checks.

src/anomalib/models/components/__init__.py
samet-akcay, line 1: I guess the license header needs an update too

src/anomalib/models/components/base/buffer_list.py
```diff
@@ -0,0 +1,108 @@
...
+        To update and retrieve the list of tensors, we dynamically assign a descriptor attribute to the class.
+
+        Args:
+            name (str): Name of the bufferlist.
+            values (list[torch.Tensor]): List of tensors to register as buffers.
+            persistent (bool, optional): Whether the buffers should be saved as part of the module state_dict.
+                Defaults to True.
+            **kwargs: Additional keyword arguments to pass to `torch.nn.Module.register_buffer`.
+
+        Example:
```
samet-akcay, line 30: Would it be an idea to have this example in the main class docstring. User would see this example at the top of the documentation. Otherwise, they might need to scroll?

This is an idea for any other module in the library actually?

tests/unit/models/components/base/test_buffer_list_mixin.py
```diff
@@ -0,0 +1,81 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+
+import pytest
+import torch
+
+from anomalib.models.components.base.buffer_list import BufferListMixin
+
+
+def tensorlists_are_equal(tensorlist1: list[torch.Tensor], tensorlist2: list[torch.Tensor]) -> None:
```
samet-akcay, line 13:
```suggestion
def tensor_lists_are_equal(tensorlist1: list[torch.Tensor], tensorlist2: list[torch.Tensor]) -> None:
```

src/anomalib/models/image/winclip/torch_model.py
```diff
@@ -0,0 +1,398 @@
...
+        The k_shot attribute is updated based on the number of reference images.
+
+        The setup method is called internally by the constructor. However, it can also be called manually to update the
+        text and visual embeddings after the model has been initialized.
+
+        Args:
+            class_name (str): The name of the object class used in the prompt ensemble.
+            reference_images (torch.Tensor): Tensor of shape ``(batch_size, C, H, W)`` containing the reference images.
+
+        Examples:
```
samet-akcay, line 97: as mentioned above, would it be an idea to move these examples to the main class docstring, so the reader can directly see some examples when they first refer the documenation, without scrolling?

samet-akcay (changes requested):
Looking really good 🔥 Just some minor cosmetics here and there.

src/anomalib/models/components/base/__init__.py
```diff
@@ -4,7 +4,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from .anomaly_module import AnomalyModule
+from .buffer_list import BufferListMixin
```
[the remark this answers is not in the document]
djdameln: I'm planning to address this separately. There's an issue for this already: https://github.com/openvinotoolkit/anomalib/issues/1649

src/anomalib/models/image/winclip/torch_model.py
djdameln: In this case I feel it's best to leave it here, because these examples specifically illustrate how to set up the model in zero- and few-shot mode using the setup method.

ashwinvaidya17 (approved):
Let's merge 🔥

PR #1628, opened 2024-01-12
v1 - [Bug] Fix issues when validation and test split modes set to 'none'

willyfh:
## 📝 Description

These changes follow the logic in the current main branch, where no issues occur when the split modes are set to 'none':

- To skip the validation stage, set `config.trainer.limit_val_batches` to `0.0` when `val_split_mode` is 'NONE' in the `config.py` file.
- To skip the test stage after training, include a conditional check `datamodule.test_split_mode == TestSplitMode.NONE` in the `engine.py` file.

**Note**: I'm not quite sure why those changes were removed in v1, but without those logics, the skipping process is not working properly.

🛠️ Fixes # (issue number)

## ✨ Changes

- [ ] 📚 Documentation update

## ✅ Checklist

### DIFF
--- a/src/anomalib/engine/engine.py
+++ b/src/anomalib/engine/engine.py
@@ -23,6 +23,7 @@ from anomalib.callbacks.normalization import get_normalization_callback
 from anomalib.callbacks.post_processor import _PostProcessorCallback
 from anomalib.callbacks.thresholding import _ThresholdCallback
 from anomalib.data import AnomalibDataModule, AnomalibDataset, PredictDataset
+from anomalib.data.utils import TestSplitMode
 from anomalib.deploy.export import ExportType, export_to_onnx, export_to_openvino, export_to_torch
 from anomalib.metrics.threshold import BaseThreshold
 from anomalib.models import AnomalyModule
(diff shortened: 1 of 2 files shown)


src/anomalib/engine/engine.py
```diff
@@ -476,7 +480,11 @@ def train(
         self._setup_trainer(model)
         self._setup_dataset_task(train_dataloaders, val_dataloaders, test_dataloaders, datamodule)
         self.trainer.fit(model, train_dataloaders, val_dataloaders, datamodule, ckpt_path)
-        self.trainer.test(model, test_dataloaders, ckpt_path=ckpt_path, datamodule=datamodule)
+
+        if datamodule is not None and datamodule.test_split_mode == TestSplitMode.NONE:
```
samet-akcay, line 484: Would there be an edge case where test split is set to None, but test set is not empty? @djdameln ?
Maybe we could also check the length of the test set to be on the safe side?

samet-akcay (commented):
Thanks @willyfh! I only have a single comment.

src/anomalib/engine/engine.py
willyfh: Do you mean something like this? If someone intentionally want to skip the test stage by setting the test split mode to None, but the test set is not empty, we give them a warning.

```python
if datamodule is not None and datamodule.test_split_mode == TestSplitMode.NONE:
    logger.info(f"The test_split_mode is set to '{TestSplitMode.NONE}'. Skipping test stage.")
    if len(datamodule.test_data) > 0:
        logger.warning(f"Found {len(datamodule.test_data)} images in the test set.")
    return None
```

Issue #1341, opened 2023-09-12
[Task]: logical anomaly detection


lemonbuilder:
### What is the motivation for this task?

I'm going to train 'Mvtec AD Loco' dataset using EfficientAd.
Could I have some info for that?
I wonder what part of tools/train.py I should modify for **logical anomaly detection**.
Or should I use another appropriate command?

### Describe the solution you'd like

I downloaded 'Mvtec AD Loco' dataset.

### Additional context

I read the following post.
- Add support for MVTEC LOCO AD #471
- Jpcbertoldo/mvtec ad loco #538
- Jpcbertoldo/mvtec ad loco 2 #553

jpcbertoldo:
@lemonbuilder I lost track of tha script, but I believe there were refactors on the parent classes

https://github.com/openvinotoolkit/anomalib/blob/ed1d9fa5c7050a6edf79febd256d27b261a441ae/src/anomalib/data/base/dataset.py#L36

https://github.com/openvinotoolkit/anomalib/blob/ed1d9fa5c7050a6edf79febd256d27b261a441ae/src/anomalib/data/base/datamodule.py#L52

so it probably won't work right away

samet-akcay:
thanks for the update @jpcbertoldo. 

@lemonbuilder, you could then follow #471 for any update since this will be a duplicate

samet-akcay:
If you want to implement it yourself, you could use [AnomalibDataset](https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/data/base/dataset.py) and [AnomalibDatamodules](https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/data/base/datamodule.py) to implement

Issue #471, opened 2022-08-02
Add support for MVTEC LOCO AD


ORippler:
**Is your feature request related to a problem? Please describe.**

MVTec AD dataset only consists of low-level, textural/structural anomalies. To test the capability at detecting high-level, semantic/logical anomalies, additional datasets are needed.

**Describe the solution you'd like**

Integrate a default dataloader and evaluation scheme for [MVTEC LOCO AD](https://link.springer.com/article/10.1007/s11263-022-01578-9#Fn1) dataset

PR #1635, opened 2024-01-13
🚀 v1 - Add support for MVTec LOCO dataset and sPRO metric

willyfh:
## 📝 Description

Add the implementation of MVTec LOCO AD dataset [[Paper](https://link.springer.com/article/10.1007/s11263-022-01578-9#Fn1)]

The dataset structure could contains multiple ground-truth masks for the corresponding anomalous images. The dataset `_setup` method will merge the multiple masks and store the merged ground-truth masks into a separate ground-truth directory via the `_merge_gt_mask` function.

Note: I used the downloadable dataset link found from this closed PR: https://github.com/openvinotoolkit/anomalib/pull/553

This PR does not include the implementation of sPRO metrics, which seems the metric is designed to additionally evaluate MVTec LOCO dataset. I plan to implement the metric in a separate PR later.

🛠️ Fixes #574 #471 #1341 

## ✨ Changes

- [ ] 📚 Documentation update

## ✅ Checklist

- [x] 📋 I have summarized my changes in the [CHANGELOG](https://github.com/openvinotoolkit/anomalib/blob/main/CHANGELOG.md) and followed the guidelines for my type of change (skip for minor changes, documentation updates, and test enhancements).

### DIFF
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -98,13 +98,10 @@ def _merge_gt_mask(
                 mask = cv2.imread(str(mask_file), cv2.IMREAD_UNCHANGED)
                 if merged_mask is None:
                     merged_mask = np.zeros_like(mask)
-                merged_mask = np.maximum(merged_mask, mask)
+                merged_mask += mask
 
         if merged_mask is None:
             continue
-
-        # Binarize masks
-        merged_mask = np.where(merged_mask > 0, 255, 0)
 
         # Define the path for the new merged mask
         _, anomaly_dir, image_filename = mask_path.parts[-3:]
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -338,9 +338,9 @@ class MVTecLocoDataset(AnomalibDataset):
     def __getitem__(self, index: int) -> dict[str, str | torch.Tensor]:
         """Get dataset item for the index ``index``.
 
-        The implementation of this method is mostly based on the parent's class implementation, with some different as follows:
+        This method is mostly based on the super class implementation, with some different as follows:
             - Using 'torch.where' to make sure the 'mask' in the return item is binarized
-            - An additional 'masks' is added to pass the non-binary masks with original size for the sPRO metric calculation
+            - An additional 'masks' is added, the non-binary masks with original size for the sPRO metric calculation
         Args:
             index (int): Index to get the item.
 
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -100,8 +100,11 @@ def _merge_gt_mask(
                     merged_mask = np.zeros_like(mask)
                 merged_mask = np.maximum(merged_mask, mask)
 
+        if merged_mask is None:
+            continue
+
         # Binarize masks
-        merged_mask = np.minimum(merged_mask, 255)
+        merged_mask = np.where(merged_mask > 0, 255, 0)
 
         # Define the path for the new merged mask
         _, anomaly_dir, image_filename = mask_path.parts[-3:]
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -105,7 +105,7 @@ def _merge_gt_mask(
 
         # Define the path for the new merged mask
         _, anomaly_dir, image_filename = mask_path.parts[-3:]
-        new_mask_path = root / Path(gt_merged_dir) / anomaly_dir / (image_filename + ".png")
+        new_mask_path = root / gt_merged_dir / anomaly_dir / (image_filename + ".png")
 
         # Create the necessary directories if they do not exist
         new_mask_path.parent.mkdir(parents=True, exist_ok=True)
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -134,7 +134,7 @@ def make_mvtec_loco_dataset(
     +---+---------------+-------+---------+---------------+---------------------------------------+-------------+
     |   | path          | split | label   | image_path    | mask_path                             | label_index |
     +===+===============+=======+=========+===============+=======================================+=============+
-    | 0 | datasets/name | test  | defect  | filename.png  | path/to/merged_masks/filename.png      | 1           |
+    | 0 | datasets/name | test  | defect  | filename.png  | path/to/merged_masks/filename.png     | 1           |
     +---+---------------+-------+---------+---------------+---------------------------------------+-------------+
 
     Note: the final image_path is converted to full path by combining it with the path, split, and label columns
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -16,9 +16,11 @@ References:
       in: International Journal of Computer Vision (IJCV) 130, 947-969, 2022, DOI: 10.1007/s11263-022-01578-9
 """
 
+import json
 import logging
 from collections.abc import Sequence
 from pathlib import Path
+from typing import Any
 
 import albumentations as A  # noqa: N812
 import cv2
@@ -62,6 +64,41 @@ CATEGORIES = (
 )
 
 GT_MERGED_DIR = "ground_truth_merged"
+
+SATURATION_CONFIG_FILENAME = "defects_config.json"
+
+
+def load_saturation_config(config_path: str | Path) -> dict[int, Any]:
+    """Load saturation configurations from a JSON file.
+
+    Args:
+        config_path (str | Path): Path to the saturation configuration file.
+
+    Returns:
+        Dict: A dictionary with pixel values as keys and the corresponding configurations as values.
+
+    Example JSON format in the file:
+    [
+        {
+            "defect_name": "1_additional_pushpin",
+            "pixel_value": 255,
+            "saturation_threshold": 6300,
+            "relative_saturation": false
+        },
+        {
+            "defect_name": "2_additional_pushpins",
+            "pixel_value": 254,
+            "saturation_threshold": 12600,
+            "relative_saturation": false
+        },
+        ...
+    ]
+    """
+    with Path.open(Path(config_path)) as file:
+        configs = json.load(file)
+
+    # Create a dictionary with pixel values as keys
+    return {conf["pixel_value"]: conf for conf in configs}
 
 
 def _merge_gt_mask(
@@ -448,9 +485,10 @@ class MVTecLoco(AnomalibDataModule):
             val_split_ratio=val_split_ratio,
             seed=seed,
         )
-
+        self.saturation_config: dict[int, Any]
         self.root = Path(root)
         self.category = Path(category)
+        self.saturation_config = {}
 
         transform_train = get_transforms(
             config=transform_config_train,
@@ -533,7 +571,7 @@ class MVTecLoco(AnomalibDataModule):
             download_and_extract(self.root, DOWNLOAD_INFO)
 
     def _setup(self, _stage: str | None = None) -> None:
-        """Set up the datasets and perform dynamic subset splitting.
+        """Set up the datasets, configs, and perform dynamic subset splitting.
 
         This method overrides the parent class's method to also setup the val dataset.
         The MVTec LOCO dataset provides an independent validation subset.
@@ -548,3 +586,6 @@ class MVTecLoco(AnomalibDataModule):
 
         self._create_test_split()
         self._create_val_split()
+
+        saturation_path = self.root / self.category / SATURATION_CONFIG_FILENAME
+        self.saturation_config = load_saturation_config(saturation_path)
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -23,6 +23,7 @@ from pathlib import Path
 import albumentations as A  # noqa: N812
 import cv2
 import numpy as np
+import torch
 from pandas import DataFrame
 
 from anomalib import TaskType
@@ -36,6 +37,8 @@ from anomalib.data.utils import (
     ValSplitMode,
     download_and_extract,
     get_transforms,
+    masks_to_boxes,
+    read_image,
 )
 
 logger = logging.getLogger(__name__)
@@ -294,6 +297,54 @@ class MVTecLocoDataset(AnomalibDataset):
             extensions=IMG_EXTENSIONS,
             gt_merged_dir=GT_MERGED_DIR,
         )
+
+    def __getitem__(self, index: int) -> dict[str, str | torch.Tensor]:
+        """Get dataset item for the index ``index``.
+
+        The implementation of this method is mostly based on the parent's class implementation, with some different as follows:
+            - Using 'torch.where' to make sure the 'mask' in the return item is binarized
+            - An additional 'masks' is added to pass the non-binary masks with original size for the sPRO metric calculation
+        Args:
+            index (int): Index to get the item.
+
+        Returns:
+            dict[str, str | torch.Tensor]: Dict of image tensor during training. Otherwise, Dict containing image path,
+                target path, image tensor, label and transformed bounding box.
+        """
+        image_path = self._samples.iloc[index].image_path
+        mask_path = self._samples.iloc[index].mask_path
+        label_index = self._samples.iloc[index].label_index
+
+        image = read_image(image_path)
+        item = {"image_path": image_path, "label": label_index}
+
+        if self.task == TaskType.CLASSIFICATION:
+            transformed = self.transform(image=image)
+            item["image"] = transformed["image"]
+        elif self.task in (TaskType.DETECTION, TaskType.SEGMENTATION):
+            # Only Anomalous (1) images have masks in anomaly datasets
+            # Therefore, create empty mask for Normal (0) images.
+            mask = np.zeros(shape=image.shape[:2]) if label_index == 0 else cv2.imread(mask_path, flags=0)
+            mask = mask.astype(np.single)
+
+            transformed = self.transform(image=image, mask=mask)
+
+            item["image"] = transformed["image"]
+            item["mask_path"] = mask_path
+            # transform and binarize the mask
+            item["mask"] = torch.where(transformed["mask"] > 0, torch.tensor(1.0), torch.tensor(0.0))
+            # The non-binary masks with the original size for saturation based metrics calculation
+            item["masks"] = torch.tensor(mask)
+
+            if self.task == TaskType.DETECTION:
+                # create boxes from masks for detection task
+                boxes, _ = masks_to_boxes(item["mask"])
+                item["boxes"] = boxes[0]
+        else:
+            msg = f"Unknown task type: {self.task}"
+            raise ValueError(msg)
+
+        return item
 
 
 class MVTecLoco(AnomalibDataModule):
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -1,0 +1,499 @@
+"""MVTec LOCO AD Dataset (CC BY-NC-SA 4.0).
+
+Description:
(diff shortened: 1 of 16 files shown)


src/anomalib/data/image/mvtec_loco.py
```diff
@@ -0,0 +1,499 @@
...
+        merged_mask = None
+        for mask_file in mask_path.glob("*"):
+            if mask_file.suffix in extensions:
+                mask = cv2.imread(str(mask_file), cv2.IMREAD_UNCHANGED)
+                if merged_mask is None:
+                    merged_mask = np.zeros_like(mask)
+                merged_mask = np.maximum(merged_mask, mask)
+
+        # Binarize masks
+        merged_mask = np.minimum(merged_mask, 255)
```
willyfh, line 104: Not sure if i need to binarize the masks here or not. The ground-truth masks of the dataset is can vary, e.g., 241, 253, etc, depending on the sub-categories of the defects

blaz-r: Interesting observation.  I think that is a good idea to only have 255 values in any case, since the masks are divided by 255 here:
https://github.com/openvinotoolkit/anomalib/blob/69184e378ea025d1f23ad86ebac1ada53debbe25/src/anomalib/data/base/dataset.py#L130

But I think that this line doesn't exactly do that? Isn't this element wise minimum, for example in case of 245, going to set output to 245, not 255. 
I think this should probably be `merged_mask = np.where(merged_mask > 0, 255, 0)`

blaz-r:
Nice work 😃, thanks for this addition. I didn't thoroughly check all the code, but I did comment on the question you had.

src/anomalib/data/image/mvtec_loco.py
willyfh: Thanks for your comment! ah yes, I think it should be `merged_mask = np.where(merged_mask > 0, 255, 0)` as you said. I have made the changes.

blaz-r: Great.

fengchuibeixiang:
Hi, a good work! Are you using the "V1" branch of this project? I cannot find the path "src/anomalib/data/image" in the main branch. Should I switch to the "V1" branch to use your change? Looking forward to your reply. Thank you!

samet-akcay:
> Hi, a good work! Are you using the "V1" branch of this project? I cannot find the path "src/anomalib/data/image" in the main branch. Should I switch to the "V1" branch to use your change? Looking forward to your reply. Thank you!

@fengchuibeixiang , v1 will be released on the 31 of January

phcarval:
Hi, thanks for the work. If I understand correctly, you merge all existing ground truth maps into one and binarize the result. Will it not make it difficult to work with the sPRO metric, which requires the different ground truth maps and their corresponding threshold?

willyfh:
> Hi, thanks for the work. If I understand correctly, you merge all existing ground truth maps into one and binarize the result. Will it not make it difficult to work with the sPRO metric, which requires the different ground truth maps and their corresponding threshold?

@phcarval thanks for your comment! Yes, I binarized the multi ground truth maps into one and stored it in a separate folder, without removing the original unmerged ground truth maps. This is necessary for the existing metrics. With and without `merged_mask = np.minimum(merged_mask, 255)`, the evaluation existing metrics give different results.

But as you said, for sPRO metric, we need to consider to have a case where the validation/test set also can support the unmerged ones. I was thinking to add additional columns, something like `multi_mask_path` or `unmerged_mask_path`, in the returned `samples`, see here: https://github.com/openvinotoolkit/anomalib/pull/1635/files#diff-7e31cb96286793cb547505586ea1e87b6825cf6111db0b686bf276a732769300R138

The evaluation of the sPRO metrics can utilize the additional column to access the unmerged maps. It seems i may also need to modify the `__getitem__` or override the method. I was planning to add the additional column when I work on the sPRO metric later. But maybe it will be better to address it in this PR.

phcarval:
> > Hi, thanks for the work. If I understand correctly, you merge all existing ground truth maps into one and binarize the result. Will it not make it difficult to work with the sPRO metric, which requires the different ground truth maps and their corresponding threshold?
> 
> @phcarval thanks for your comment! Yes, I binarized the multi ground truth maps into one and stored it in a separate folder, without removing the original unmerged ground truth maps. This is necessary for the existing metrics. With and without `merged_mask = np.minimum(merged_mask, 255)`, the evaluation existing metrics give different results.
> 
> But as you said, for sPRO metric, we need to consider to have a case where the validation/test set also can support the unmerged ones. I was thinking to an add additional column, something like `multi_mask_path` or `unmerged_mask_path`, in the returned `samples`, see here: https://github.com/openvinotoolkit/anomalib/pull/1635/files#diff-7e31cb96286793cb547505586ea1e87b6825cf6111db0b686bf276a732769300R138

I think it's a good idea. I think it's pretty close to what @jpcbertoldo did in his PR https://github.com/openvinotoolkit/anomalib/pull/553.



> The evaluation of the sPRO metrics can utilize the additional column to access the unmerged maps. It seems i may also need to modify the `__getitem__` or override the method. I was planning to add the additional column when I work on the sPRO metric later. But maybe it will be better to address it in this PR.
> 
> **P.S.:** I haven't fully checked how the metrics calculation flow in anomalib, so my idea on using the additional column can be wrong :D

I have already thought about adding LOCO myself, but I haven't had time to do it for the moment. I think to properly implement the dataset the following points must be addressed:

- Add a column with the "true values" of the anomaly maps (perhaps a simple sum?)
- Add a dict in the MVTecLOCO class that is loaded using the .yaml files of LOCO to get the thresholds corresponding to each defect category (i.e. the pixel values in the anomaly maps)
- Add an sPRO metric which will load the "true values" with their corresponding thresholds.

I think metrics are loaded somewhere in `metrics_configuration.py` or `metrics/__init__.py`. Right now I don't know exactly how a metric can retrieve the data that is loaded in MVTecLOCO though :/

Thanks anyway for the PR!

blaz-r:
Yeah as @phcarval said, there are quite some things to consider. I think that we'd need to think through how to best support logical anomaly datasets as well as supervised datasets in the long run.

willyfh:
@blaz-r @phcarval Thanks for your comments! I will try to address those asap :D

samet-akcay:
@jpcbertoldo, any comments from pimo perspective?

jpcbertoldo:
Disclaimer: I read the paper in-depth long ago and just re-skimmed it, so i might miss some details.

If I understood correctly, the GT merging is done for the sake of simplifying the use of other metrics. However, (pixel-level) AUROC doesnt make sense for logical anomalies.

An example, to make sure we're on the same page: in the image below, each square is supposed to have one pushpin but they all have 2 pushpins. 

`pushpins/test/logical_anomalies/018.png`



masks `001.png` and `003.png`

image

So this image has 15 sPRO curves. Some of them (e.g. like above) have disconnected regions, but they're kind of considered as a single region for the metric (not in the morphological sense) because chosing one pushpin or another is the same ("Additional objects" in section "3.4 Selection of Saturation Thresholds" in the paper talks about this case).

So if you merge all 15 annotations you still need to encode the information of where each connected region comes from. Right? If that is correct, is the merging actually useful?

> @jpcbertoldo, any comments from pimo perspective?

As for PIMO, good news : )

Since the annotations are already split in different files, an analogous sPIMO curve could be defined as (for instance) the average (on the y-axis) of the sPRO curves coming from the same image.

phcarval:
> So if you merge all 15 annotations you still need to encode the information of where each connected region comes from. Right? If that is correct, is the merging actually useful?

The difficulty comes from the fact that you can only associate one mask to one image. Here we have 15 masks, so in order for it to work properly, we're forced to merge them together. But I thought that's what you also did on your original PR, since I see that you use both "mask" and "masks" key for the dataset. I assume you merged the maps for one and collected them in a list for the other.

jpcbertoldo:
> > So if you merge all 15 annotations you still need to encode the information of where each connected region comes from. Right? If that is correct, is the merging actually useful?
> 
> The difficulty comes from the fact that you can only associate one mask to one image. Here we have 15 masks, so in order for it to work properly, we're forced to merge them together. But I thought that's what you also did on your original PR, since I see that you use both "mask" and "masks" key for the dataset. I assume you merged the maps for one and collected them in a list for the other.

I vaguely recall doing that indeed, but looking back it doesn't feel like the best idea. Again, i may be missing another detail that made me do that at the time.

Maybe it could be managed the other way around? Instead of forcing the masks to become a single one, make the single source image be repeated. 

That internal dataframe you could make each row point to a specific mask (as it is in the file structure) and you would see several rows with the same input image. Kind of suboptimal computationally but i think there a not sooo many cases with multiple masks.

phcarval:
> > > So if you merge all 15 annotations you still need to encode the information of where each connected region comes from. Right? If that is correct, is the merging actually useful?
> > 
> > 
> > The difficulty comes from the fact that you can only associate one mask to one image. Here we have 15 masks, so in order for it to work properly, we're forced to merge them together. But I thought that's what you also did on your original PR, since I see that you use both "mask" and "masks" key for the dataset. I assume you merged the maps for one and collected them in a list for the other.
> 
> I vaguely recall doing that indeed, but looking back it doesn't feel like the best idea. Again, i may be missing another detail that made me do that at the time.
> 
> Maybe it could be managed the other way around? Instead of forcing the masks to become a single one, make the single source image be repeated.
> 
> That internal dataframe you could make each row point to a specific mask (as it is in the file structure) and you would see several rows with the same input image. Kind of suboptimal computationally but i think there a not sooo many cases with multiple masks.

I think duplicating the source images means that you risk skewing the image-level metrics though.

willyfh:
@jpcbertoldo Thank you so much for your comments! I've just started working on anomaly detection and anomalib recently, so please understand if my response or concern below maybe wrong and sounds too naive 😅 I think some things need to be clarified.

> If I understood correctly, the GT merging is done for the sake of simplifying the use of other metrics. However, (pixel-level) AUROC doesn't make sense for logical anomalies.

 Initially, I thought merging the masks would be helpful if we wanted to use this dataset with existing metrics. I believed this approach is simple and doesn't require invasive changes that might break many parts of the existing code, which expects a single mask path/map (e.g., get item method, metric calculation, visualization, etc.).


I wonder, is there any case where an Anomalib user wants to evaluate the MVTec LOCO dataset using existing metric behaviors (including using connected regions morphologically), even if it isn't entirely suitable? Perhaps just for the sake of experimentation, research, or simple curiosity?

> So if you merge all 15 annotations you still need to encode the information of where each connected region comes from. Right? If that is correct, is the merging actually useful?

Yes, I think, as @phcarval suggested, which seems similar to your PR on #553, using the sum operation would help encode the information into a single image, avoiding duplicating rows in the samples. This "merging" may also help "connect" the multiple masks within a single row of samples, which I believe is necessary for batch/dataset operations, as I will explain further below.

> Maybe it could be managed the other way around? Instead of forcing the masks to become a single one, make the single source image be repeated. That internal dataframe you could make each row point to a specific mask (as it is in the file structure) and you would see several rows with the same input image. Kind of suboptimal computationally but i think there a not sooo many cases with multiple masks.


I think the problem with using a repeated row for the same image is that it will "disconnect" the masks in the batching operation. Since each row is treated as a single instance, the same images with different masks can be separated into different batches. However, don't we need all masks to correspond to each image for one step of metric calculation or visualization? Perhaps image-level metrics, as mentioned by @phcarval, could be relevant. I could be wrong since I am not fully familiar with the metrics calculation yet. The issue of "disconnected" masks could occur even if the samples are not shuffled for each step.

jpcbertoldo:
> I think duplicating the source images means that you risk skewing the image-level metrics though.

Hmmm... Yeah, you're right. LOCO is complicated indeed haha.

> I wonder, is there any case where an Anomalib user wants to evaluate the MVTec LOCO dataset using existing metric behaviors (including using connected regions morphologically), even if it isn't entirely suitable? Perhaps just for the sake of experimentation, research, or simple curiosity?

Maybe for debug. I don't see it otherwise because LOCO's annotations are quite case-specific.

> Yes, I think, as @phcarval suggested, which seems similar to your PR on https://github.com/openvinotoolkit/anomalib/pull/553, using the sum operation would help encode the information into a single image, avoiding duplicating rows in the samples.

I don't recall exactly what I did, but if you do the sum indeed then I believe you should need to encode different sub-masks (?) with different integers.

For instance, the example in https://github.com/openvinotoolkit/anomalib/pull/1635#issuecomment-1896114054 (say there are only those two masks) would be something like 

```python
merged = 1 * mask_001 + 2 * mask_003
```

(0=black, 1=white, 2=green)
image

so that you are able to recover the individual masks from `merged`. Otherwise you'd get

(0=black, 1=white)
image

where you cannot distinguish anymore which regions are from the same mask, so sPRO would be wrong.

Not sure we're on the same page but I think it's important to address this.

phcarval:
> I don't recall exactly what I did, but if you do the sum indeed then I believe you should need to encode different sub-masks (?) with different integers.

That's a very good point I hadn't thought about! Not only do you need to encode for the different defect categories, but you also have to differentiate different masks that have the same value but are loaded from different files. And you somehow need to make it work through Torchmetrics' trainer. LOCO is complicated indeed.

jpcbertoldo:
> ... but you also have to differentiate **different masks that have the same value but are loaded from different files.**

I'm not sure I understand when that would happen. The one-file-one-integer encoding I proposed was specifically to address this issue.

phcarval:
No I agree with you here. Just making the point that you must differentiate not only the different defect types but also the different map files.

willyfh:
> Maybe for debug. I don't see it otherwise because LOCO's annotations are quite case-specific.
I see, i will consider this in the implementation, maybe by giving a warning when someone trying to use the metrics on the existing metrics implementation (in morphological way), or just forcing it via assert by not allowing them at all.

> I don't recall exactly what I did, but **if you do the sum indeed then I believe you should need to encode different sub-masks (?) with different integers.**

Ah, now I understand what you meant here. Based on your code, since the masks were binarized after being read, simply performing a sum would result in everything becoming 0 or 1. That's why the `_sum_masks` is necessary to encode the values with different integers. I got it, I will also address this in the implementation.

image


> ... but you also have to differentiate **different masks that have the same value but are loaded from different files.**

Oh, because after the encoding the masks to the integers, we lost the information of the original pixel value if there is no mapping to the original pixel value (?). If let say in one image has two different masks with pixel value of 241 and 253, and another image has also two different masks, 231 and 251, the masks for the different images will be encoded into 1 and 2 (?). So connecting to the `defect config` later would be difficult. A sample of config:
```
{
        "defect_name": "2_additional_pushpins",
        "pixel_value": 254,
        "saturation_threshold": 12600,
        "relative_saturation": false
    },
    {
        "defect_name": "missing_pushpin",
        "pixel_value": 253,
        "saturation_threshold": 6300,
        "relative_saturation": false
    },

**Comment by phcarval** [comment]
If I can give you a hand at the interface level: to communicate with the sPRO metric, I believe the best way to do that is using dict metrics initialization. In the file `/anomalib/utils/metrics/__init__.py` you have:
```python
def metric_collection_from_dicts(metrics: dict[str, dict[str, Any]], prefix: str | None) -> AnomalibMetricCollection:
``` 
that allows you to get any kind of kwargs to your sPRO metric class though the `init_args` field:
```yaml
metrics:
        pixel:
                PixelWiseF1Score:
                    class_path: torchmetrics.F1Score
                    init_args: {}
                PixelWiseAUROC:
                    class_path: anomalib.utils.metrics.AUROC
                    init_args:
                        compute_on_cpu: true
```
Now this function is called from the `MetricsConfigurationCallback` in file `src/anomalib/utils/callbacks/metrics_configuration.py`, called in the `get_callbacks()` function in file `src/anomalib/utils/callbacks/__init__.py` which is itself directly called in the `train()` and `test()` functions.

Now, I'm not sure how to proceed to actually pass something that has been computed in the MVTec LOCO initialization to the sPRO metric using these functions. Perhaps implementing some kind of callback in `get_callbacks()` that is going to gather this map information from MVTec LOCO (or any dataset that would work similarly) and directly feed it to the sPRO metric?

jpcbertoldo:
So the question is basically how to make the saturation value of each image reach the metric, right?

Couldn't you have the dataset return it in the batch?

phcarval:
You can, but how does it reach the metric? From the PRO metric:
```python
def update(self, predictions: Tensor, targets: Tensor) -> None:
```
AFAIK, you can only use the preds and targets in the update. And the targets would be the merged anomaly maps (I presume?)

jpcbertoldo:
Hm.. change the signature of update() to get a new arg? Optionally?

I'm guessing it'd require some annoying changes in the code that assumes this interface. 

An alternative could be to just sneak in another tensor and make target a tuple (masks, saturations) or (masks, configs) both in the metric and in the dataset.

phcarval:
It's possible that something can be done in `src/anomalib/models/components/base/anomaly_module.py` in the method `_collect_outputs()`, which calls the metrics. You could squeeze an extra argument in `pixel_metric.update()` if the key "masks" is defined in the output dictionary, for instance. I don't know if the masks will be in this dictionary, though, or if something else needs to be done for it to be the case.

jpcbertoldo:
> I don't know if the masks will be in this dictionary, though, or if something else needs to be done for it to be the case.

We're talking about the `dict` returned by `AnomalibDataset.__getitem__()`, right?

The masks are returned, yes: 

https://github.com/openvinotoolkit/anomalib/blob/ff808e342ba6a11ce2a0e0d0cb3ea02e3a9c9e23/src/anomalib/data/base/dataset.py#L137

phcarval:
So if we add the masks to the output of `getitem()` the LOCO dataset, then they're going to be available in `_collect_outputs()` too! That means that we can add a piece of code in the latter that adds the masks to the targets if you have the sPRO metric. I guess that's the cleanest way to do it.

willyfh:
Thank you for your comments! @jpcbertoldo @phcarval , 

Based on your comments, I have made the following changes:
- Used the sum operation to merge the masks based on the original pixel values.
- Overridden `__getitem()` to return the additional `masks` and to handle the binarization of the `mask`.
- Loaded the saturation config in the datamodule and passed the config to the metric via the `setup()` of `_MetricsCallback`.
- Passed the `masks` to the metric via `pixel_metric.update` on the validation batch end by adding an additional (keyword) argument.
- Added a warning if the dataset is evaluated using the existing pixel-level metric implementation, where the 'region' is interpreted in a morphological sense.
- Added the sPRO metric implementation.

Remaining things to do:
- Add unit tests.
- Update documentation.
- Consider adding AU-sPRO (?).

phcarval:
Great job, thanks! Now I haven't tested it all but AUsPRO should definitely be added (as the AU- metrics are widely used in the library, AUPRO, AUROC...). I don't think it's very difficult if you've added sPRO, I believe there is an auc class somewhere in torchmetrics that will do the computation by itself.

jpcbertoldo:
Great @willyfh !

> * Loaded the saturation config in the datamodule and passed the config to the metric via the `setup()` of `_MetricsCallback`.

I just didn't quite get this one. 

The saturation parameter of sPRO is passed to the metric object on the setup? Can you point out to that part in the code plz?

willyfh:
> I don't think it's very difficult if you've added sPRO, I believe there is an auc class somewhere in torchmetrics that will do the computation by itself.

I checked the implementation of aupro.py, seems it is quite complicated. the approach doesn't seem similar with pro.py. But ya, i need to understand the implementation first.

> > * Loaded the saturation config in the datamodule and passed the config to the metric via the `setup()` of `_MetricsCallback`.
> 
> I just didn't quite get this one.
> 
> The saturation parameter of sPRO is passed to the metric object on the setup? Can you point out to that part in the code plz?

The saturation_config which is loaded in the dataset is passed to the metric in this part:
https://github.com/openvinotoolkit/anomalib/blob/e0678c00dd8713c4cdae450696d352475a5d4a52/src/anomalib/callbacks/metrics.py#L93-L95

where finally this method will be executed, to update the metric which having the `saturation_config` attribute:

https://github.com/openvinotoolkit/anomalib/blob/e0678c00dd8713c4cdae450696d352475a5d4a52/src/anomalib/metrics/collection.py#L30-L40

Please let me know if you have any concerns.

samet-akcay:
Hi @willyfh, sorry merging v1 to main closed all the PRs that were targeting v1 :(

If you could create the PR targeting main, we could still merge it I think

PR #1650, opened 2024-01-17
🔨 Check value of max_epochs in config for PatchCore

abc-125:
## 📝 Description

PatchCore model should be trained for 1 epoch due to how the algorithm works, but now it is possible to change it in the config, causing unexpected behavior, as in this [discussion](https://github.com/openvinotoolkit/anomalib/discussions/1645).

## ✨ Changes

Added `ValueError` if the number of epochs differs from one and a comment in the `config.yaml`.

- [ ] 📚 Documentation update

## ✅ Checklist


samet-akcay:
Thanks @abc-125, I think we should have done this way earlier as it caused quite confusion.

In v1, we overwrite these parameters with the following trainer arguments property.
https://github.com/openvinotoolkit/anomalib/blob/b204c508e631e765fb5817a18dce2052b72ffa6c/src/anomalib/models/image/patchcore/lightning_model.py#L120-L123

Do you think if we should still have this check?

abc-125:
Thank you for the explanation @samet-akcay ,

> In v1, we overwrite these parameters with the following trainer arguments property.

The current version has too many things in the config files that can be accidentally broken. This way of handling the problem looks perfect to me! Also, it is much easier to understand which parameters can be changed now.

> Do you think if we should still have this check?

I don't think so, I will close the pull request :)

PR #1656, opened 2024-01-22
🔒 Validate path before processing

samet-akcay:
## 📝 Description

This pr adds additional checks to the input path to avoid path traversal, and read/write permission issues

🛠️ Fixes # CVS-129511, CVS-129516

## ✨ Changes

- [ ] 📚 Documentation update
- [x] 🔒 Security update

## ✅ Checklist

### DIFF
--- a/src/anomalib/data/utils/path.py
+++ b/src/anomalib/data/utils/path.py
@@ -5,6 +5,7 @@
 
 
 import os
+import re
 from enum import Enum
 from pathlib import Path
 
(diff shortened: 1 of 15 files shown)


src/anomalib/data/utils/path.py
```diff
@@ -98,3 +99,119 @@ def _resolve_path(folder: str | Path, root: str | Path | None = None) -> Path:
...
+        >>> contains_null_bytes("./datasets/MVTec/bottle/train/good/000.png")
+        False
+
+        >>> contains_null_bytes("./datasets/MVTec/bottle/train/good/000.png\0")
+        True
+    """
+    return "\0" in str(path)
+
+
+def validate_path(path: str | Path, base_dir: str | Path | None = None) -> Path:
```
ashwinvaidya17, line 123: I can't think of any other name but I would expect `validate_path` to return a bool. How about `valid_path`?

ashwinvaidya17 (approved):
Thanks for the efforts! I have one super minor comment but feel free to ignore.

src/anomalib/data/utils/path.py
samet-akcay: hm, I would expect valid_path to return bool :) maybe `validate_and_return_path`, but a bit verbose? 

@djdameln, any thoughts on this ?

src/anomalib/data/utils/path.py
```diff
@@ -98,3 +100,147 @@ def _resolve_path(folder: str | Path, root: str | Path | None = None) -> Path:
         # root provided; prepend root and return absolute path
         path = (Path(root) / folder).resolve()
     return path
+
+
+def is_path_too_long(path: str | Path, max_length: int = 512) -> bool:
+    r"""Check if the path contains too long input.
```
ashwinvaidya17, line 106: Is the `r` intentional?

src/anomalib/data/utils/path.py
djdameln: It's a bit unclear to me what is the difference between `validate_path` and `validate_and_resolve_path`, because `validate_path` also seems to resolve the path from relative to absolute. Are both really needed, or could we reduce these two one single function called `validate_and_resolve_path`?

samet-akcay: validate_path resolves it to avoid accessing any other directory beyond `base_dir`? I created `validate_and_resolve_path` to not mess with the previous behaviour of `_resolve_path`

Issue #1654, opened 2024-01-19
🔨 Refactor assert statements with explicit error handling


samet-akcay:
### What is the motivation for this task?

bandit scans treat `assert` as a potential security issue with the following warning:

```bash
>> Issue: Use of assert detected. The enclosed code will be removed when
   compiling to optimised byte code.
   Severity: Low   Confidence: High
   CWE: CWE-703 (https://cwe.mitre.org/data/definitions/703.html)
   Location: ./examples/assert.py:1
1 assert logged_in
2 display_assets()
```

### Describe the solution you'd like

remove asserts from the main codebase, only use it in the test suite.

sahusiddharth:
Hello @samet-akcay, I have started working on this.

I have a small query, `assert` is the code. I'm planning to change it using conditionals, for raising errors what would you prefer 
- Raising a `ValueError` or any single type of error everywhere [quick fix]
for example 

from this
```python
assert isinstance(order_of_parts, list), 'The field "order_of_parts" in optimisation config should be a list'
```
to
```python
if not sinstance(order_of_parts, list):
    raise ValueError('The field "order_of_parts" in optimisation config should be a list')
```
- Raising specific errors  
for example 

from this
```python
assert isinstance(order_of_parts, list), 'The field "order_of_parts" in optimisation config should be a list'
```
to
```python
if not sinstance(order_of_parts, list):
    raise TypeError('The field "order_of_parts" in optimisation config should be a list')
```

Which of the two would you prefer? I'm open to suggestions :)

sahusiddharth:
.[take]

samet-akcay:
@sahusiddharth , thank you for your interest! My answer is, it depends on the use-case. If our focus in the condition is type such as list, in this case it could be `TypeError`. If we want to check the value instead, it could be `ValueError`. There is a better explanation down below
https://teamtreehouse.com/community/in-python-whats-the-difference-between-a-type-error-and-a-value-error

sahusiddharth:
@samet-akcay, Can you help me with checking it is taking way too long to check or it is giving me an error `OSError: [Errno 24] Too many open files:` when running the command

```zsh
pytest tests/integration
```

<img width="1440" alt="Screenshot 2024-02-05 at 3 17 40 PM" src="https://github.com/openvinotoolkit/anomalib/assets/112792547/a26b3b6c-2221-4ba7-b334-65be05b4f1a1">

[...]

Issue #1629, opened 2024-01-12
🐞 v1 - Issues when validation and test split modes are set to 'none'


willyfh:
### Describe the bug

- When val_split_mode is set to none >  Error: datamodule object has no attribute 'val_data'
- When test_split_mode is set to none > the test stage is not skipped

### Dataset

Other (please specify in the text field below)

### Model

Other (please specify in the field below)

### Steps to reproduce the behavior

Set val_split_mode or test_split_mode to NONE in a config data file, then start the training process using the config file.

### OS information

-

### Expected behavior

when NONE is set to either val_split_mode or test_split_mode, the corresponding stage should be skipped.

### Pip/GitHub

pip

### What version/branch did you use?

v1

### Configuration YAML

```yaml
class_path: anomalib.data.Kolektor
init_args:
  root: "./datasets/kolektor"
  image_size: [256, 256]
  center_crop: null
  normalization: imagenet
  train_batch_size: 32
  eval_batch_size: 32
  num_workers: 8
  transform_config_train: null
  transform_config_eval: null
  test_split_mode: NONE
  test_split_ratio: 0.2
  val_split_mode: NONE
  val_split_ratio: 0.5
  seed: null
```

### Logs

```shell
│ /media/willy/Windows/data_ubuntu/anomalib/anomalib/src/anomalib/data/base/datamodule.py:198 in   │
│ val_dataloader                                                                                   │
│                                                                                                  │
│   195 │   def val_dataloader(self) -> EVAL_DATALOADERS:                                          │
│   196 │   │   """Get validation dataloader."""                                                   │
│   197 │   │   return DataLoader(                                                                 │
│ ❱ 198 │   │   │   dataset=self.val_data,                                                         │
│   199 │   │   │   shuffle=False,                                                                 │
│   200 │   │   │   batch_size=self.eval_batch_size,                                               │
│   201 │   │   │   num_workers=self.num_workers,                                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
AttributeError: 'Kolektor' object has no attribute 'val_data'
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1557, opened 2023-12-19
🚀 Pimo

jpcbertoldo:
## Description

Replaces the PRs from https://gist.github.com/jpcbertoldo/12553b7eaa97cfbf3e55bfd7d1cafe88 to catch up with v1.

Implements refactors from https://github.com/jpcbertoldo/anomalib/blob/metrics/refactors/src/anomalib/utils/metrics/perimg/.refactors .

- arxiv: https://arxiv.org/abs/2401.01984
- medium post: https://medium.com/p/c653ac30e802
- GSoC deliverable: https://gist.github.com/jpcbertoldo/12553b7eaa97cfbf3e55bfd7d1cafe88

## Changes

<details>
<summary>Describe the changes you made</summary>

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor (non-breaking change which refactors the code base)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

</details>

## Checklist

<details>
<summary>Ensure that you followed the following</summary>

- [ ] I have added a summary of my changes to the [CHANGELOG](https://github.com/openvinotoolkit/anomalib/blob/main/CHANGELOG.md) (not for minor changes, docs and tests).
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas (If applicable)
- [ ] I have made corresponding changes to the documentation (If applicable)
- [ ] I have added tests that prove my fix is effective or that my feature works (If applicable)

</details>

PR #1686, opened 2024-01-27
🚀 Add support for MVTec LOCO dataset and sPRO metric

willyfh:
## 📝 Description

This PR is based on the closed PR #1635 , which adds the implementation of MVTec LOCO AD dataset [[Paper](https://link.springer.com/article/10.1007/s11263-022-01578-9#Fn1)] along with the sPRO metric.

🛠️ Fixes #574 #471 #1341 

## ✨ Changes

- [x] 📚 Documentation update
- [ ] 🔒 Security update

## ✅ Checklist

- [x] 📋 I have summarized my changes in the [CHANGELOG](https://github.com/openvinotoolkit/anomalib/blob/main/CHANGELOG.md) and followed the guidelines for my type of change (skip for minor changes, documentation updates, and test enhancements).

### DIFF
--- a/src/anomalib/callbacks/metrics.py
+++ b/src/anomalib/callbacks/metrics.py
@@ -67,7 +67,7 @@ class _MetricsCallback(Callback):
             pl_module (AnomalyModule): Anomalib Model that inherits pl LightningModule.
             stage (str | None, optional): fit, validate, test or predict. Defaults to None.
         """
-        del stage  # this variable is not used.
+        del stage, trainer  # this variable is not used.
         image_metric_names = [] if self.image_metric_names is None else self.image_metric_names
         if isinstance(image_metric_names, str):
             image_metric_names = [image_metric_names]
@@ -97,8 +97,6 @@ class _MetricsCallback(Callback):
             else:
                 pl_module.pixel_metrics = create_metric_collection(pixel_metric_names, "pixel_")
             self._set_threshold(pl_module)
-            if hasattr(trainer.datamodule, "saturation_config"):
-                self._set_saturation_config(pl_module, trainer.datamodule.saturation_config)
 
     def on_validation_epoch_start(
         self,
@@ -173,9 +171,6 @@ class _MetricsCallback(Callback):
         pl_module.image_metrics.set_threshold(pl_module.image_threshold.value.item())
         pl_module.pixel_metrics.set_threshold(pl_module.pixel_threshold.value.item())
 
-    def _set_saturation_config(self, pl_module: AnomalyModule, saturation_config: dict[int, Any]) -> None:
-        pl_module.pixel_metrics.set_saturation_config(saturation_config)
-
     def _update_metrics(
         self,
         image_metric: AnomalibMetricCollection,
@@ -205,10 +200,14 @@ class _MetricsCallback(Callback):
     def _update_pixel_metrics(self, pixel_metric: AnomalibMetricCollection, output: STEP_OUTPUT) -> None:
         """Handle metric updates when the SPRO metric is used alongside other pixel-level metrics."""
         update = False
-        for metric in pixel_metric.values(copy_state=False):
+        for name, metric in pixel_metric.items(copy_state=False):
             if isinstance(metric, SPRO):
                 metric.update(torch.squeeze(output["anomaly_maps"]), output["masks"])
             else:
+                logger.warning(
+                    f"Metric {name} may not be suitable for a dataset with the region separated "
+                    "in multiple ground-truth masks.",
+                )
                 metric.update(torch.squeeze(output["anomaly_maps"]), torch.squeeze(output["mask"].int()))
             update = True
         pixel_metric.set_update_called(update)
--- a/src/anomalib/callbacks/metrics.py
+++ b/src/anomalib/callbacks/metrics.py
@@ -186,7 +186,11 @@ class _MetricsCallback(Callback):
         image_metric.update(output["pred_scores"], output["label"].int())
         if "mask" in output and "anomaly_maps" in output:
             pixel_metric.to(self.device)
-            pixel_metric.update(torch.squeeze(output["anomaly_maps"]), torch.squeeze(output["mask"].int()))
+            pixel_metric.update(
+                torch.squeeze(output["anomaly_maps"]),
+                torch.squeeze(output["mask"].int()),
+                masks=torch.squeeze(output["masks"]) if "masks" in output else None,
+            )
 
     def _outputs_to_device(self, output: STEP_OUTPUT) -> STEP_OUTPUT | dict[str, Any]:
         if isinstance(output, dict):
--- a/src/anomalib/callbacks/metrics.py
+++ b/src/anomalib/callbacks/metrics.py
@@ -13,7 +13,7 @@ from lightning.pytorch import Callback, Trainer
 from lightning.pytorch.utilities.types import STEP_OUTPUT
 
 from anomalib import TaskType
-from anomalib.metrics import AnomalibMetricCollection, create_metric_collection
+from anomalib.metrics import SPRO, AnomalibMetricCollection, create_metric_collection
 from anomalib.models import AnomalyModule
 
 logger = logging.getLogger(__name__)
@@ -186,11 +186,10 @@ class _MetricsCallback(Callback):
         image_metric.update(output["pred_scores"], output["label"].int())
         if "mask" in output and "anomaly_maps" in output:
             pixel_metric.to(self.device)
-            pixel_metric.update(
-                torch.squeeze(output["anomaly_maps"]),
-                torch.squeeze(output["mask"].int()),
-                masks=torch.squeeze(output["masks"]) if "masks" in output else None,
-            )
+            if "masks" in output:
+                self._update_pixel_metrics(pixel_metric, output)
+            else:
+                pixel_metric.update(torch.squeeze(output["anomaly_maps"]), torch.squeeze(output["mask"].int()))
 
     def _outputs_to_device(self, output: STEP_OUTPUT) -> STEP_OUTPUT | dict[str, Any]:
         if isinstance(output, dict):
@@ -198,7 +197,21 @@ class _MetricsCallback(Callback):
                 output[key] = self._outputs_to_device(value)
         elif isinstance(output, torch.Tensor):
             output = output.to(self.device)
+        elif isinstance(output, list):
+            for i, value in enumerate(output):
+                output[i] = self._outputs_to_device(value)
         return output
+
+    def _update_pixel_metrics(self, pixel_metric: AnomalibMetricCollection, output: STEP_OUTPUT) -> None:
+        """Handle metric updates when the SPRO metric is used alongside other pixel-level metrics."""
+        update = False
+        for metric in pixel_metric.values(copy_state=False):
+            if isinstance(metric, SPRO):
+                metric.update(torch.squeeze(output["anomaly_maps"]), output["masks"])
+            else:
+                metric.update(torch.squeeze(output["anomaly_maps"]), torch.squeeze(output["mask"].int()))
+            update = True
+        pixel_metric.set_update_called(update)
 
     @staticmethod
     def _log_metrics(pl_module: AnomalyModule) -> None:
--- a/src/anomalib/callbacks/metrics.py
+++ b/src/anomalib/callbacks/metrics.py
@@ -67,8 +67,7 @@ class _MetricsCallback(Callback):
             pl_module (AnomalyModule): Anomalib Model that inherits pl LightningModule.
             stage (str | None, optional): fit, validate, test or predict. Defaults to None.
         """
-        del trainer, stage  # These variables are not used.
-
+        del stage  # this variable is not used.
         image_metric_names = [] if self.image_metric_names is None else self.image_metric_names
         if isinstance(image_metric_names, str):
             image_metric_names = [image_metric_names]
@@ -98,6 +97,8 @@ class _MetricsCallback(Callback):
             else:
                 pl_module.pixel_metrics = create_metric_collection(pixel_metric_names, "pixel_")
             self._set_threshold(pl_module)
+            if hasattr(trainer.datamodule, "saturation_config"):
+                self._set_saturation_config(pl_module, trainer.datamodule.saturation_config)
 
     def on_validation_epoch_start(
         self,
@@ -171,6 +172,9 @@ class _MetricsCallback(Callback):
     def _set_threshold(self, pl_module: AnomalyModule) -> None:
         pl_module.image_metrics.set_threshold(pl_module.image_threshold.value.item())
         pl_module.pixel_metrics.set_threshold(pl_module.pixel_threshold.value.item())
+
+    def _set_saturation_config(self, pl_module: AnomalyModule, saturation_config: dict[int, Any]) -> None:
+        pl_module.pixel_metrics.set_saturation_config(saturation_config)
 
     def _update_metrics(
         self,
--- a/src/anomalib/data/base/datamodule.py
+++ b/src/anomalib/data/base/datamodule.py
@@ -164,6 +164,9 @@ class AnomalibDataModule(LightningDataModule, ABC):
             # converted from random training sample
             self.train_data, normal_val_data = random_split(self.train_data, self.val_split_ratio, seed=self.seed)
             self.val_data = SyntheticAnomalyDataset.from_dataset(normal_val_data)
+        elif self.val_split_mode == ValSplitMode.FROM_DIR:
+            # the val_data is prepared in subclass
+            pass
         elif self.val_split_mode != ValSplitMode.NONE:
             msg = f"Unknown validation split mode: {self.val_split_mode}"
             raise ValueError(msg)
--- a/src/anomalib/data/base/datamodule.py
+++ b/src/anomalib/data/base/datamodule.py
@@ -26,7 +26,9 @@ logger = logging.getLogger(__name__)
 def collate_fn(batch: list) -> dict[str, Any]:
     """Collate bounding boxes as lists.
 
-    Bounding boxes are collated as a list of tensors, while the default collate function is used for all other entries.
+    Bounding boxes and `masks` (not `mask`) are collated as a list of tensors. If `masks` is exist,
+    the `mask_path` is also collated as a list since each element in the batch could be unequal.
+    For all other entries, the default collate function is used.
 
     Args:
         batch (List): list of items in the batch where len(batch) is equal to the batch size.
@@ -40,6 +42,10 @@ def collate_fn(batch: list) -> dict[str, Any]:
         if "boxes" in elem:
             # collate boxes as list
             out_dict["boxes"] = [item.pop("boxes") for item in batch]
+        if "masks" in elem:
+            # collate masks and mask_path as list
+            out_dict["masks"] = [item.pop("masks") for item in batch]
+            out_dict["mask_path"] = [item.pop("mask_path") for item in batch]
         # collate other data normally
         out_dict.update({key: default_collate([item[key] for item in batch]) for key in elem})
         return out_dict
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -98,13 +98,10 @@ def _merge_gt_mask(
                 mask = cv2.imread(str(mask_file), cv2.IMREAD_UNCHANGED)
                 if merged_mask is None:
                     merged_mask = np.zeros_like(mask)
-                merged_mask = np.maximum(merged_mask, mask)
+                merged_mask += mask
 
         if merged_mask is None:
             continue
-
-        # Binarize masks
-        merged_mask = np.where(merged_mask > 0, 255, 0)
 
         # Define the path for the new merged mask
         _, anomaly_dir, image_filename = mask_path.parts[-3:]
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -338,9 +338,9 @@ class MVTecLocoDataset(AnomalibDataset):
     def __getitem__(self, index: int) -> dict[str, str | torch.Tensor]:
         """Get dataset item for the index ``index``.
 
-        The implementation of this method is mostly based on the parent's class implementation, with some different as follows:
+        This method is mostly based on the super class implementation, with some different as follows:
             - Using 'torch.where' to make sure the 'mask' in the return item is binarized
-            - An additional 'masks' is added to pass the non-binary masks with original size for the sPRO metric calculation
+            - An additional 'masks' is added, the non-binary masks with original size for the sPRO metric calculation
         Args:
             index (int): Index to get the item.
 
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -100,8 +100,11 @@ def _merge_gt_mask(
                     merged_mask = np.zeros_like(mask)
                 merged_mask = np.maximum(merged_mask, mask)
 
+        if merged_mask is None:
+            continue
+
         # Binarize masks
-        merged_mask = np.minimum(merged_mask, 255)
+        merged_mask = np.where(merged_mask > 0, 255, 0)
 
         # Define the path for the new merged mask
         _, anomaly_dir, image_filename = mask_path.parts[-3:]
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -105,7 +105,7 @@ def _merge_gt_mask(
 
         # Define the path for the new merged mask
         _, anomaly_dir, image_filename = mask_path.parts[-3:]
-        new_mask_path = root / Path(gt_merged_dir) / anomaly_dir / (image_filename + ".png")
+        new_mask_path = root / gt_merged_dir / anomaly_dir / (image_filename + ".png")
 
         # Create the necessary directories if they do not exist
         new_mask_path.parent.mkdir(parents=True, exist_ok=True)
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -134,7 +134,7 @@ def make_mvtec_loco_dataset(
     +---+---------------+-------+---------+---------------+---------------------------------------+-------------+
     |   | path          | split | label   | image_path    | mask_path                             | label_index |
     +===+===============+=======+=========+===============+=======================================+=============+
-    | 0 | datasets/name | test  | defect  | filename.png  | path/to/merged_masks/filename.png      | 1           |
+    | 0 | datasets/name | test  | defect  | filename.png  | path/to/merged_masks/filename.png     | 1           |
     +---+---------------+-------+---------+---------------+---------------------------------------+-------------+
 
     Note: the final image_path is converted to full path by combining it with the path, split, and label columns
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -16,9 +16,11 @@ References:
       in: International Journal of Computer Vision (IJCV) 130, 947-969, 2022, DOI: 10.1007/s11263-022-01578-9
 """
 
+import json
 import logging
 from collections.abc import Sequence
 from pathlib import Path
+from typing import Any
 
 import albumentations as A  # noqa: N812
 import cv2
@@ -62,6 +64,41 @@ CATEGORIES = (
 )
 
 GT_MERGED_DIR = "ground_truth_merged"
+
+SATURATION_CONFIG_FILENAME = "defects_config.json"
+
+
+def load_saturation_config(config_path: str | Path) -> dict[int, Any]:
+    """Load saturation configurations from a JSON file.
+
+    Args:
+        config_path (str | Path): Path to the saturation configuration file.
+
+    Returns:
+        Dict: A dictionary with pixel values as keys and the corresponding configurations as values.
+
+    Example JSON format in the file:
+    [
+        {
+            "defect_name": "1_additional_pushpin",
+            "pixel_value": 255,
+            "saturation_threshold": 6300,
+            "relative_saturation": false
+        },
+        {
+            "defect_name": "2_additional_pushpins",
+            "pixel_value": 254,
+            "saturation_threshold": 12600,
+            "relative_saturation": false
+        },
+        ...
+    ]
+    """
+    with Path.open(Path(config_path)) as file:
+        configs = json.load(file)
+
+    # Create a dictionary with pixel values as keys
+    return {conf["pixel_value"]: conf for conf in configs}
 
 
 def _merge_gt_mask(
@@ -448,9 +485,10 @@ class MVTecLoco(AnomalibDataModule):
             val_split_ratio=val_split_ratio,
             seed=seed,
         )
-
+        self.saturation_config: dict[int, Any]
         self.root = Path(root)
         self.category = Path(category)
+        self.saturation_config = {}
 
         transform_train = get_transforms(
             config=transform_config_train,
@@ -533,7 +571,7 @@ class MVTecLoco(AnomalibDataModule):
             download_and_extract(self.root, DOWNLOAD_INFO)
 
     def _setup(self, _stage: str | None = None) -> None:
-        """Set up the datasets and perform dynamic subset splitting.
+        """Set up the datasets, configs, and perform dynamic subset splitting.
 
         This method overrides the parent class's method to also setup the val dataset.
         The MVTec LOCO dataset provides an independent validation subset.
@@ -548,3 +586,6 @@ class MVTecLoco(AnomalibDataModule):
 
         self._create_test_split()
         self._create_val_split()
+
+        saturation_path = self.root / self.category / SATURATION_CONFIG_FILENAME
+        self.saturation_config = load_saturation_config(saturation_path)
--- a/src/anomalib/data/image/mvtec_loco.py
+++ b/src/anomalib/data/image/mvtec_loco.py
@@ -23,6 +23,7 @@ from pathlib import Path
 import albumentations as A  # noqa: N812
 import cv2
 import numpy as np
+import torch
 from pandas import DataFrame
 
 from anomalib import TaskType
@@ -36,6 +37,8 @@ from anomalib.data.utils import (
     ValSplitMode,
     download_and_extract,
     get_transforms,
+    masks_to_boxes,
+    read_image,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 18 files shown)


samet-akcay:
@willyfh, is this ready for the review?

willyfh:
> @willyfh, is this ready for the review?

@samet-akcay I have implemented the MVTec Loco dataset and SPRO metric, and I have added the unit test for both. However, I haven't implemented the AUSPRO yet. Should AUSPRO be addressed later in a separate PR? If so, I think this PR is ready for review. :)

phcarval:
@willyfh , I would be very interested in getting the AUsPRO metric working. I would be happy to collaborate with you on this one if you're willing to :)

samet-akcay:
> > @willyfh, is this ready for the review?
> 
> @samet-akcay I have implemented the MVTec Loco dataset and SPRO metric, and I have added the unit test for both. However, I haven't implemented the AUSPRO yet. Should AUSPRO be addressed later in a separate PR? If so, I think this PR is ready for review. :)

This PR is already big, I think it would also be easier for the reviewers if you auspro is a separate one

willyfh:
> This PR is already big, I think it would also be easier for the reviewers if you auspro is a separate one

@samet-akcay ah, I see. Okay, I will mark this PR as 'Ready for review' after updating the changelog soon. 👍

willyfh:
> @willyfh , I would be very interested in getting the AUsPRO metric working. I would be happy to collaborate with you on this one if you're willing to :)

@phcarval Ah, sorry I missed your comment. If you want, you can implement the auspro metric and create the PR. I'm getting very busy at work these days, so I don't have much time. :crying_cat_face:

willyfh:
Now this PR is ready for review :) @samet-akcay

samet-akcay:
@jpcbertoldo, @phcarval, @blaz-r , do you guys have any concerns about this pr

jpcbertoldo:
> @jpcbertoldo, @phcarval, @blaz-r , do you guys have any concerns about this pr

I'll have sometime to check on wed only 😬

src/anomalib/data/image/mvtec_loco.py
```diff
@@ -0,0 +1,597 @@
...
+        >>> root = Path('./MVTec_LOCO')
+        >>> category = 'pushpins'
+        >>> path = root / category
+        >>> samples = make_mvtec_loco_dataset(path, split='test')
+    """
+    root = validate_path(root)
+    gt_merged_dir = Path(gt_merged_dir)
+
+    # assert the directory to store the merged ground-truth masks is different than the original gt directory
+    assert gt_merged_dir != "ground_truth"
```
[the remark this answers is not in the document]
samet-akcay: For details, see this 
https://github.com/openvinotoolkit/anomalib/issues/1654

blaz-r:
> @jpcbertoldo, @phcarval, @blaz-r , do you guys have any concerns about this pr

I can check it tomorrow.

phcarval:
Will check tomorrow or wednesday.

src/anomalib/data/image/mvtec_loco.py
```diff
@@ -0,0 +1,597 @@
...
+        self,
+        task: TaskType,
+        transform: A.Compose,
+        root: Path | str = "./datasets/MVTec_LOCO",
+        category: str = "breakfast_box",
+        split: str | Split | None = None,
+    ) -> None:
+        super().__init__(task=task, transform=transform)
+
+        self.root_category = Path(root) / Path(category)
```
samet-akcay, line 333:
```suggestion
        self.root_category = Path(root) / category
```

src/anomalib/metrics/spro.py
```diff
@@ -0,0 +1,158 @@
...
+    preds: list[torch.Tensor]
+
+    def __init__(self, threshold: float = 0.5, saturation_config: dict | None = None, **kwargs) -> None:
+        super().__init__(**kwargs)
+        self.threshold = threshold
+        self.saturation_config = saturation_config
+        self.add_state("preds", default=[], dist_reduce_fx="cat")
+        self.add_state("targets", default=[], dist_reduce_fx="cat")
+
+    def update(self, predictions: torch.Tensor, _: torch.Tensor, masks: torch.Tensor) -> None:
```
samet-akcay, line 63: both metric-collection and torch metric has ` def update(self, *args, **kwargs) -> None` signature. Would it be possible to remove `_: torch.Tensor`?

src/anomalib/metrics/spro.py
```diff
@@ -0,0 +1,158 @@
...
+            predictions (torch.Tensor): Predicted anomaly masks
+            _ (torch.Tensor): Unused argument, but needed for different metrics within the same AnomalibMetricCollection
+            masks (torch.Tensor): Ground truth anomaly masks with non-binary values and, original height and width
+
+        Example:
+            To update the metric state for the current batch, use the ``update`` method:
+
+            >>> spro.update(preds, _, labels)
+        """
+        assert masks is not None
```
samet-akcay, line 76: would it be possible to use explicit error handling?

src/anomalib/metrics/spro.py
```diff
@@ -0,0 +1,158 @@
...
+        predictions = predictions > threshold
+
+    score = torch.tensor(0.0)
+    m = 0
+    # Iterate for each image in the batch
+    for i, target in enumerate(targets):
+        unique_labels = torch.unique(target)
+
+        # Iterate for each ground-truth mask per image
+        for label in unique_labels[1:]:
```
samet-akcay, line 126: can you expand the comment, explaining why starting from unique_labels[1:]

tests/helpers/data.py
```diff
@@ -389,6 +393,40 @@ def _generate_dummy_mvtec_3d_dataset(self) -> None:
...
+        # Create normal images.
+        for split in ("train", "validation", "test"):
+            path = self.dataset_root / dataset_category / split / "good"
+            if split == "train":
+                num_images = self.num_train
+            elif split == "val":
+                num_images = self.num_val
+            else:
+                num_images = self.num_test
+            for i in range(num_images):
```
samet-akcay, line 412:
```suggestion
            
            for i in range(num_images):
```

src/anomalib/metrics/collection.py
```diff
@@ -3,16 +3,22 @@
...
 
+logger = logging.getLogger(__name__)
+
 
 class AnomalibMetricCollection(MetricCollection):
     """Extends the MetricCollection class for use in the Anomalib pipeline."""
 
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
+        self._saturation_config: dict
```
samet-akcay, line 18: `self._saturation_config` is defined twice here.

Also, I was wondering if it is possible to somehow exclude self._saturation_config from AnomalibMetricCollection? As you mention in the codebase, it is quite specific to a single dataset

src/anomalib/callbacks/metrics.py
```diff
@@ -98,6 +97,8 @@ def setup(
             else:
                 pl_module.pixel_metrics = create_metric_collection(pixel_metric_names, "pixel_")
             self._set_threshold(pl_module)
+            if hasattr(trainer.datamodule, "saturation_config"):
+                self._set_saturation_config(pl_module, trainer.datamodule.saturation_config)
```
samet-akcay, line 101: would it be possible to somehow handle this within the SPRO itself? This is metric specific, would be great if we could sort it out in the metric

samet-akcay (commented):
thanks a lot for adding this support, @willyfh! I've got some questions and comments. I'll try to go for another round of review later.

src/anomalib/data/image/mvtec_loco.py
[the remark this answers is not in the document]
willyfh: Sure, I will raise the errors instead of using assert, as discussed in the issue 👍

src/anomalib/callbacks/metrics.py
willyfh: @samet-akcay do you mean loading the config directly in the metric? i think it would be possible as long as we introduce a new parameter in the `yaml`, so that we can load the config in the metric. I will do as follows if it is okay.
```
metrics:
  pixel:
    - AUROC
    - SPRO
  saturation_config: ./datasets/MVTec_LOCO/breakfast_box/defects_config.json
```

src/anomalib/metrics/collection.py
willyfh: ah yeah, it seems we don't necessarily need the `self._saturation_config` AnomalibMetricCollection. But maybe we still need the `def set_saturation_config` method in the AnomalibMetricCollection to pass the config/path to the metric, i.e., SPRO and AUSPRO

src/anomalib/metrics/spro.py
willyfh: I think there might be two possible solutions (or more?) if we want to remove `_: torch.Tensor`, but we still want to handle the case of simultaneously evaluating the dataset using existing metrics (PRO, etc) and new metrics (SPRO and AUSPRO).

#### 1. Separate SPRO and AUSPRO into different metric collection
Currently, when we put the SPRO and PRO into the same pixel metric collection, it means we need to handle both case of `mask` and `masks`. so that's why we need to pass both mask and masks to the update function of the metric collection like this:
https://github.com/openvinotoolkit/anomalib/blob/a87cdf533c3a8b1c744d081c6db51edc54baa95c/src/anomalib/callbacks/metrics.py#L185-L193

However, if we can separate the metric collection, we can call update(anomaly_map, mask) and update(anomaly_map, masks) separately, so that we can remove the `_`.

#### 2. Calling update without using the `update` function of the metric collection
Instead of using `pixel_metric.update` (AnomalibMetricCollection), we iterate the metrics in the collection via `pixel_metric.items(copy_state=False)`, and call update method of the torch metric directly. Thus, each metric can call either update(anomaly_map, mask) or update(anomaly_map, masks). And after this update, we need to set  `_updated_called` attribute of the AnomalibMetricCollection to True, since it is needed for this part:

https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/callbacks/metrics.py#L198-L202

I am not sure which one is better, but the first approach seems requires lots of modification in the codes. The second approach is more simple, but i haven't fully investigate it, if it is really okay to call the torch metric update directly without using the update method of the metric collection.

src/anomalib/metrics/spro.py
willyfh: Sure, I will use an explicit error handling here. Or when we decided to remove the `_`, maybe we can safely remove this error check.

src/anomalib/metrics/spro.py
willyfh: Sure, I will add more comment, explaining the index 0 is the background/normal part of the masks

src/anomalib/metrics/collection.py
blaz-r: I too think that maybe we should explore the idea of moving this to be only sPRO specific.

src/anomalib/data/image/mvtec_loco.py
```diff
@@ -0,0 +1,597 @@
...
+    gt_mask_paths = {f.parent for f in root.glob("ground_truth/**/*") if f.suffix in extensions}
+
+    for mask_path in gt_mask_paths:
+        # Merge each mask inside mask_path into a single mask
+        merged_mask = None
+        for mask_file in mask_path.glob("*"):
+            if mask_file.suffix in extensions:
+                mask = cv2.imread(str(mask_file), cv2.IMREAD_UNCHANGED)
+                if merged_mask is None:
+                    merged_mask = np.zeros_like(mask)
```
blaz-r, line 146: Do we assume that masks are here not overlapping. Is this assumption correct for all cases?

src/anomalib/data/image/mvtec_loco.py
```diff
@@ -0,0 +1,597 @@
...
+            mask = mask.astype(np.single)
+
+            transformed = self.transform(image=image, mask=mask)
+
+            item["image"] = transformed["image"]
+            item["mask_path"] = mask_path
+            # transform and binarize the mask
+            item["mask"] = torch.where(transformed["mask"] > 0, torch.tensor(1.0), torch.tensor(0.0))
+            # The non-binary masks with the original size for saturation based metrics calculation
+            item["masks"] = torch.tensor(mask)
```
blaz-r, line 380: Just a quick question here. Did you verify that transforms don't by any chance change the values inside mask (like 244) as mask are usually all 255.

src/anomalib/metrics/spro.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # Apply threshold to binary predictions
+    if predictions.dtype == torch.float:
+        predictions = predictions > threshold
+
+    score = torch.tensor(0.0)
+    m = 0
+    # Iterate for each image in the batch
+    for i, target in enumerate(targets):
+        unique_labels = torch.unique(target)
```
blaz-r, line 123: If I understand sPRO correctly, it works in the same way as PRO in a sense that each defective region counts as a separate instance. The code here assumes that each unique value is a separate instance.

So if we have a case, where let's say an image contains two regions, marked by the same value, this will not be correct, as they will be counted as one region, but should be considered as two separate instances. (I think that's why PRO implementation uses connected components to solve this issue)

Do such cases occur inside MVTec Loco? And even if they don't, what happens if datasets in future break this assumptions?

src/anomalib/metrics/spro.py
```diff
@@ -0,0 +1,158 @@
...
+            # Calculate the areas of the ground-truth
+            defect_areas = torch.sum(target_per_label)
+
+            if saturation_config is not None:
+                # Adjust saturation threshold based on configuration
+                saturation_per_label = saturation_config[label.int().item()]
+                saturation_threshold = torch.minimum(
+                    torch.tensor(saturation_per_label["saturation_threshold"]),
+                    defect_areas,
+                )
```
blaz-r, line 140: I don't think this minimum is mentioned in the paper? I assume that if the saturation threshold is set correctly, it should always be smaller than defect area, or is this wrong?

blaz-r (commented):
Thanks for this impressive work 😄. I left some comments and questions.

src/anomalib/callbacks/metrics.py
samet-akcay: @ashwinvaidya17, can you confirm whether we could use `class_path` and `init_args` for metrics above? The implementation would be quite straightforward

src/anomalib/data/image/mvtec_loco.py
willyfh: Oh, that's a great observation! After checking all categories, it seems there's a possibility of overlapping only in the 'screw_bag' category. Although the overlapping annotations might not be easily visible, there are indeed small overlaps at the borders between the regions. I believe this implies that we can't simply merge the masks into a single mask, right? Even if such overlaps are unintentional, we lack information about which regions the overlapping annotations are supposed to belong to. It seems we need to pass a list of tensors instead of a single merged mask to handle the possibility of such cases.

For example:

src/anomalib/data/image/mvtec_loco.py
willyfh: At least for the current implementation, i have verified the transform did not change the value inside mask. However, it seems there is a plan to change the masking approach using tv_tensors `Mask` this PR https://github.com/openvinotoolkit/anomalib/pull/1706 . For that case, I haven't checked it.

src/anomalib/metrics/spro.py
willyfh: > an image contains two regions, marked by the same value
Ah ya, good point! I missed this important point 😅. the problem occurs when we do the sum operation (merging the masks) for the case of two (or more) regions marked by the same pixel. we lose the "region" information.

Related with the overlapping issue, I think we should entirely remove the merging process, just pass the list of masks from the dataset. And then here, iterate over the list instead of over the unique labels. But this would consumes more memory i guess.

src/anomalib/metrics/spro.py
willyfh: For this case, I was referring to the evaluation code I found here: https://github.com/nelson1425/EfficientAD?tab=readme-ov-file#mvtec-loco-dataset. It seems to be just a precautionary measure, in case the saturation_threshold is incorrectly set to a larger value than the defect area. However, providing a warning might be a better way to handle such a case instead of using this minimum.

image

willyfh:
@blaz-r thanks for your comments and questions! There is an important point that would be changed based on the discussions. The merged tensor for the multi masks, i.e., `masks`, seems can't handle all cases. (see: https://github.com/openvinotoolkit/anomalib/pull/1686#discussion_r1480428113 and https://github.com/openvinotoolkit/anomalib/pull/1686#discussion_r1480445582). So we could just simply pass a list of masks without merging. But for the binary mask, i.e., `mask`, we may still need to merge it for the usage in the other metrics. cc: @samet-akcay @jpcbertoldo @phcarval

src/anomalib/metrics/spro.py
blaz-r: It does seem like a precautionary measure to me as well and I agree that user should be warned in such case. Maybe @nelson1425 can provide more insight here.

src/anomalib/metrics/spro.py
blaz-r: Yes, that can happen. I think that not merging is actually a good option as you then don't have to use the connected components which is quite time-consuming. But again we'd need to verify that even a single mask doesn't have two regions with same annotation.

src/anomalib/data/image/mvtec_loco.py
blaz-r: Okay, we'll need to keep this in mind for future, maybe we can cover this with tests?

src/anomalib/data/image/mvtec_loco.py
blaz-r: > It seems we need to pass a list of tensors instead of a single merged mask to handle the possibility of such cases.
I believe this would indeed be the best option for all these problems, or at least something similar to this approach. It'd also avoid problem if somewhere in the future a new such dataset appears that would break our assumptions.

jpcbertoldo: > > It seems we need to pass a list of tensors instead of a single merged mask to handle the possibility of such cases.
> 
> ...It'd also avoid problem if somewhere in the future a new such dataset appears that would break our assumptions.

This example looks like an imperfection in the annotations. 

But, yeah, overlap of logical masks sounds like something that will eventually happen. So I agree we should assume overlap as possible. 

Does it require changing the interface (a method would now require accepting tensor + list of tensors) in many places, @willyfh ?

---

just for the sake of documenting that example in https://github.com/openvinotoolkit/anomalib/pull/1686#discussion_r1480649912 

MVTec LOCO, Sec 3.1 and Tab 12
A screw bag contains exactly two washers, two nuts, one long screw, and one short screw.

MVTecLOCO/screw_bag/ground_truth/logical_anomalies/129/
000.png --> white = 248 --> 2_additional_nuts
001.png --> white = 240 --> 2_missing_washers

jpcbertoldo (commented):
@willyfh i didn't througly review because it looks like there will an essential change about dealing with the multiple masks of a single image (which will eventually change interfaces as well). 

I left comments on this topic, but overall it seems like there will be a `list[list[tensor]]` being passed around.

I'm imagining a solution like this (just a scratch):
https://gist.github.com/jpcbertoldo/b766da23f45a4117b428940764c50de3

Which has a weird effect: a batch with `n` predictions can spit out `>= n` spro curves (1 per mask, but each image can have 1+ masks). 
However it should be ok for SPRO because it is averaged on the batch dimension before the AUC computation.

One major point is that going through these masks with `for` loops will be slow.
The good news is that https://github.com/openvinotoolkit/anomalib/pull/1557#event-11583633153 (replacement PR to be created) will have a method that is much faster for computing the TPR (ie. overlap, recall) curves. 

If you manage to do something like the approach in my scratch above, it will be compatible with the code coming from there and we'll be able to accelerate the computation here later.

--- 

Disclaimer: Raising another topic that i'm not necessarily in favor of adressing for the sake of getting this PR through. 

MVTecLoco assumes that each anomaly type (which have different labels) define the saturation parameter of each image. This strategy is rather tight because one could define image-specific saturation.

Again, not necessary to cover this here, but having it in mind can't hurt IMO.

---

@willyfh i know this dataset is very painful but thanks for the effort, you're doing great!

src/anomalib/data/image/mvtec_loco.py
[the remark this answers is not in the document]
willyfh: In the latest updates, there is no `gt_merged_dir` anymore, and i have replaced assert with raising error in the other parts of codes.

src/anomalib/metrics/spro.py
willyfh: I have removed `_: torch.Tensor` in the latest implementation, by using the 2nd approach. I have tested it, it is working as expected. 

https://github.com/openvinotoolkit/anomalib/blob/3bbc7507ea29f537118728b579e72e04b998f78f/src/anomalib/callbacks/metrics.py#L205-L214

src/anomalib/metrics/spro.py
willyfh: I have removed the `_`, now this assert is unnecessary and already removed.

src/anomalib/metrics/spro.py
willyfh: The latest implementation is not using `unique_labels[1:]` anymore. But i have provided the related comment in this line:

https://github.com/openvinotoolkit/anomalib/blob/3bbc7507ea29f537118728b579e72e04b998f78f/src/anomalib/metrics/spro.py#L136

src/anomalib/metrics/collection.py
willyfh: I have removed `self._saturation_config`, but moving the saturation_config completely to SPRO is not explored yet.  Maybe it depends on this comment? https://github.com/openvinotoolkit/anomalib/pull/1686#discussion_r1480484061

This code is still in the metric collection
https://github.com/openvinotoolkit/anomalib/blob/2cdeb82eb79348a23325fb3d873862775366faa2/src/anomalib/metrics/collection.py#L28-L37

src/anomalib/data/image/mvtec_loco.py
willyfh: I have finally managed to pass the list of tensors (after a painfully trial and error process 😅). We need to update collate function to handle the unequal length of list in the batch, and also update the `_outputs_to_device`.

https://github.com/openvinotoolkit/anomalib/blob/2cdeb82eb79348a23325fb3d873862775366faa2/src/anomalib/data/base/datamodule.py#L45-L48

https://github.com/openvinotoolkit/anomalib/blob/2cdeb82eb79348a23325fb3d873862775366faa2/src/anomalib/callbacks/metrics.py#L200-L202

src/anomalib/metrics/spro.py
willyfh: i have changed the implementation, now a list of tensor is passed to the metric. seems we can resolve this conversation since the `unique_labels` approach is not used anymore.

src/anomalib/metrics/spro.py
willyfh: @jpcbertoldo I think Eq1 is this:
https://github.com/openvinotoolkit/anomalib/blob/2cdeb82eb79348a23325fb3d873862775366faa2/src/anomalib/metrics/spro.py#L164

But the one mentioned by @blaz-r is to avoid the saturation_threshold set to a larger value than the actual anomalous areas, which could cause wrong calculation. For example, if the actual anomalous area is only 100, but the threshold is set to 200, it means there is no way for the predictions to achieve a perfect score even if all anomaly areas are correctly predicted. `score=min(100/200, 1) = 0.5`

src/configs/data/mvtec_loco.yaml
```diff
@@ -0,0 +1,18 @@
...
+  train_batch_size: 32
+  eval_batch_size: 32
+  num_workers: 8
+  task: SEGMENTATION
+  transform_config_train: null
+  transform_config_eval: null
+  test_split_mode: FROM_DIR
+  test_split_ratio: 0.2
+  val_split_mode: FROM_DIR
+  val_split_ratio: 0.5
```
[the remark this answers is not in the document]
willyfh: I believe using 'null' might lead to ambiguity with the 'none' option, no? As far as I know, 'none' is supposed to be used to skip the validation/testing stage. You can refer to this issue context:https://github.com/openvinotoolkit/anomalib/issues/1629.

I thought 'FROM_DIR' would be suitable since in the case of 'test_split_mode', as also used by other datasets like MVTec, etc., the FROM_DIR option retrieves the test sample from a fixed test set. Even if FROM_DIR is used, the split ratio actually ignored because the test set already provides both normal and abnormal samples.

But yeah, the `*_split_ratio` is quite confusing because it is actually ignored. So it might be better to remove the split ratio from the `yaml` file.

https://github.com/openvinotoolkit/anomalib/blob/2cdeb82eb79348a23325fb3d873862775366faa2/src/configs/data/mvtec.yaml#L14-L17

willyfh:
@jpcbertoldo thanks for your comments! I have provided responses to them. I have managed to pass list of tensor from the dataset to the metrics. I have few questions.

>I'm imagining a solution like this (just a scratch):
https://gist.github.com/jpcbertoldo/b766da23f45a4117b428940764c50de3

Is this solution also needed for the the current sPRO (spro.py) metric? or for the AU-sPRO (auspro.py) metric which will be implemented later in a separate PR? Currently the spro.py only provides the spro computation for a single threshold (no curve), which is similar with `pro.py` for the PRO metric.

> MVTecLoco assumes that each anomaly type (which have different labels) define the saturation parameter of each image. This strategy is rather tight because one could define image-specific saturation.

Does this mean that each mask of each image is annotated with its own saturation parameter? I wonder if there is any existing dataset that defines such image-specific saturation? If so, maybe we can "upgrade" the implementation to accommodate such dataset when the dataset is implemented later 😅

src/configs/data/mvtec_loco.yaml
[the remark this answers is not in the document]
jpcbertoldo: `test_split_mode: FROM_DIR` and `val_split_mode: FROM_DIR` is fine, i was just referring to this: 

>  if FROM_DIR is used, the split ratio actually ignored because the test set already provides both normal and abnormal samples.

>  So it might be better to remove the split ratio from the yaml file.

@ashwinvaidya17 is that the convention?

jpcbertoldo (commented):
> I have managed to pass list of tensor from the dataset to the metrics. 

You're right, `list[Tensor]` should do it (instead of `list[list[Tensor]]` as I had proposed). 

---

> Is this solution also needed for the the current sPRO (spro.py) metric? or for the AU-sPRO (auspro.py) metric which will be implemented later in a separate PR? Currently the spro.py only provides the spro computation for a single threshold (no curve), which is similar with `pro.py` for the PRO metric.

Ok , i didn't realize this was also the approach for PRO. I was a surprised to see it is defined as the score at a single threshold because it is always used a curve (threhold vs. score) in the end, and what actually matters is its  AUC (the AU(s(PRO)). 

As this is already the design of PRO, i think it is ok follow it for the sake of consistency.

---

> Does this mean that each mask of each image is annotated with its own saturation parameter? I wonder if there is any existing dataset that defines such image-specific saturation? 

To be clear: LOCO does **not** annotated per mask; each mask's saturation is deduced from its label.

I was just raising the possibility of that happening, but I don't think _this_ PR should cover that case.

src/configs/data/mvtec_loco.yaml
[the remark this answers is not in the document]
willyfh: Ah i see, I misunderstood your comments. Yeah, i wonder why the test_split_ratio and val_split_ratio are provided in the config file of MVTec and some other datasets even though it seems that those are ignored.

src/anomalib/metrics/spro.py
```diff
@@ -0,0 +1,166 @@
...
+
+    Returns:
+        torch.Tensor: Scalar value representing the average SPRO score for the input batch.
+    """
+    # Add batch dim if not exist
+    if len(predictions.shape) == 2:
+        predictions = predictions.unsqueeze(0)
+
+    # Resize the prediction to have the same size as the target mask
+    predictions = torch.nn.functional.interpolate(predictions.unsqueeze(1), targets[0].shape[-2:])
```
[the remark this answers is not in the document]
willyfh: @jpcbertoldo Ah yes, I agree the resizing process should be outside of the metric. Regarding this resizing process, there is a thing that I'm not really sure about the standard or correct procedure to handle this.

(a) Should we resize the predictions to match the size of the original mask? (b) Or should we resize the original mask to match the size of the prediction?

For the existing `mask`, it seems the mask is the one being resized to match the prediction, i.e., it depends on the `image_size` parameter in the config, right? The resizing process happens in `__getitem__`.
https://github.com/openvinotoolkit/anomalib/blob/2cdeb82eb79348a23325fb3d873862775366faa2/src/anomalib/data/base/dataset.py#L133

However, I noticed the way MVTec LOCO is evaluated [here](https://github.com/nelson1425/EfficientAD) (I'm not sure if this is the common procedure or not): the prediction is resized back to the original size during inference, i.e., following the size of the original mask. Then, resized predictions are evaluated on the original mask size.

If (b), maybe we can simply remove this interpolation and let the transform in `__get_item__` transform the `masks` to match the prediction size. If (a), maybe we need to resize the prediction back to the original size in this `post_process`, along with providing a warning to let the user know the image size will be resized back to the original mask during evaluation.
https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/callbacks/post_processor.py#L97

@samet-akcay @ashwinvaidya17 @blaz-r

src/anomalib/metrics/collection.py
```diff
@@ -21,6 +25,21 @@ def set_threshold(self, threshold_value: float) -> None:
...
+                metric.saturation_config = saturation_config
+            else:
+                logger.warning(
+                    f"Metric {name} may not be suitable for a dataset with the region separated "
+                    "in multiple ground-truth masks.",
+                )
+
+    def set_update_called(self, val: bool) -> None:
+        """Set the flag indicating whether the update method has been called."""
+        self._update_called = val
```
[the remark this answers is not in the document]
willyfh: it it used in here:

https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/callbacks/metrics.py#L198

If we don't set _update_called  of the pixel_metric to true, the pixel-level metrics will not displayed in the terminal.

src/anomalib/data/image/mvtec_loco.py
willyfh: I added a unit test to check the output  the `mask` here:

https://github.com/openvinotoolkit/anomalib/blob/829289b80a1b34645d8f054965f8be4f88a71fc4/tests/unit/data/image/test_mvtec_loco.py#L34-L39

src/anomalib/metrics/spro.py
willyfh: I have added the warning. Along with the warning, I think we still need to change the threshold for the case when it is incorrectly set to a larger value than the defect area. Because having larger value in the config does not make sense and would cause incorrect score calculation.

https://github.com/openvinotoolkit/anomalib/blob/53aa07f9209dd19add239db8b147adf70ff880ed/src/anomalib/metrics/spro.py#L151-L158

src/anomalib/metrics/collection.py
willyfh: Now I have managed to remove the saturation_config related code from the metric collection in this commit https://github.com/openvinotoolkit/anomalib/pull/1686/commits/d9a233340061c70334f726869e4189d33e01c061

I will explain more in this comment https://github.com/openvinotoolkit/anomalib/pull/1686#discussion_r1478746532

src/anomalib/callbacks/metrics.py
willyfh: @samet-akcay @ashwinvaidya17 In the latest implementation, I have handled the loading of the saturation config in the metric itself. You can see what I modified in this commit: https://github.com/openvinotoolkit/anomalib/pull/1686/commits/d9a233340061c70334f726869e4189d33e01c061.

I found there is an existing function, i.e., `metric_collection_from_dicts`, which allows us to use `class_path` and `init_args` by passing a `dict[str, dict[str, Any]]` to the function

https://github.com/openvinotoolkit/anomalib/blob/main/src/anomalib/metrics/__init__.py#L184-L186

However, the current CLI doesn't support the dict type. So I modified this part by adding the dict type to make it work:
https://github.com/openvinotoolkit/anomalib/blob/d9a233340061c70334f726869e4189d33e01c061/src/anomalib/cli/cli.py#L146-L151

I have tested this configuration in the YAML file, and the saturation_config path can be loaded on the metric side:

```yaml
metrics:
  pixel:
    SPRO:
      class_path: anomalib.metrics.SPRO
      init_args:
        saturation_config: ./datasets/MVTec_LOCO/pushpins/defects_config.json
```

src/anomalib/metrics/collection.py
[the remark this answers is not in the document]
jpcbertoldo: ok. i'm not so familiar with this specific feature

phcarval:
@willyfh : Sorry I don't have much time to do an in-depth review of your PR, but I see a lot of great comments already. Very nice job anyway!

src/anomalib/data/base/datamodule.py
```diff
@@ -164,6 +170,9 @@ def _create_val_split(self) -> None:
             # converted from random training sample
             self.train_data, normal_val_data = random_split(self.train_data, self.val_split_ratio, seed=self.seed)
             self.val_data = SyntheticAnomalyDataset.from_dataset(normal_val_data)
+        elif self.val_split_mode == ValSplitMode.FROM_DIR:
+            # the val_data is prepared in subclass
+            pass
```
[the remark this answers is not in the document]
willyfh: > we do want to stick to the predefined splits for reproducibility.
@djdameln Yeah, actually, this is why the FROM_DIR option was implemented in the first place, to use the predefined validation split as it is. However, if a user doesn't want to use this option for some reason, or perhaps for the reason you mentioned, can they freely change it to same_as_test or synthetic?

src/anomalib/data/base/datamodule.py
```diff
@@ -40,6 +42,10 @@ def collate_fn(batch: list) -> dict[str, Any]:
         if "boxes" in elem:
             # collate boxes as list
             out_dict["boxes"] = [item.pop("boxes") for item in batch]
+        if "masks" in elem:
+            # collate masks and mask_path as list
+            out_dict["masks"] = [item.pop("masks") for item in batch]
+            out_dict["mask_path"] = [item.pop("mask_path") for item in batch]
```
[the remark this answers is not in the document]
willyfh: @djdameln In the MVTec LOCO dataset, a single image could be paired with multiple mask paths stored as a list, where the number of mask paths for each image can vary. When using the default collate function, it will raise an error because the Torch stack function cannot stack items with different lengths. The code is needed to avoid such error.

https://github.com/openvinotoolkit/anomalib/blob/d9a233340061c70334f726869e4189d33e01c061/src/anomalib/data/image/mvtec_loco.py#L81-L85

src/anomalib/callbacks/metrics.py
```diff
@@ -182,16 +181,37 @@ def _update_metrics(
...
+            for i, value in enumerate(output):
+                output[i] = self._outputs_to_device(value)
         return output
 
+    def _update_pixel_metrics(self, pixel_metric: AnomalibMetricCollection, output: STEP_OUTPUT) -> None:
+        """Handle metric updates when the SPRO metric is used alongside other pixel-level metrics."""
+        update = False
+        for name, metric in pixel_metric.items(copy_state=False):
+            if isinstance(metric, SPRO):
+                metric.update(torch.squeeze(output["anomaly_maps"]), output["masks"])
```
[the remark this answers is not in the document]
willyfh: @djdameln this approach was actually to address this comment https://github.com/openvinotoolkit/anomalib/pull/1686#discussion_r1478718458 . I provided a more detailed explanation in this comment https://github.com/openvinotoolkit/anomalib/pull/1686#discussion_r1479756383

My initial approach was as follows:

```
pixel_metric.update( 
         torch.squeeze(output["anomaly_maps"]), 
         torch.squeeze(output["mask"].int()), 
         masks=torch.squeeze(output["masks"]) if "masks" in output else None, 
     ) 
```

My initial approach can handle all metrics (including SPRO and others) without metric-specific logic in the callbacks.metrics.py file. However, the problem is that we need an additional argument, `_`:

```
def update(self, predictions: torch.Tensor, _: torch.Tensor, masks: torch.Tensor) -> None:
```
So, I changed the solution to avoid using `_` by incorporating metric-specific logic. Is there a better idea to avoid metric-specific logic while also avoiding the use of `_`?

src/configs/data/mvtec_loco.yaml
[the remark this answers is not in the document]
willyfh: > We could consider to set them default as null and raise a warning when a value is provided but is ignored.

@djdameln Got it. Should we just set them as null in this PR? I mean, this part `raise a warning when a value is provided but is ignored.`, I think out of the scope of this PR since it is actually the problem of all existing datasets, right? Not only for this specific dataset.

src/anomalib/metrics/spro.py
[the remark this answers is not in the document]
willyfh: >  I would propose to use the same convention for the SPRO metric as for the other metrics, which is to resize the masks and evaluate at the user-defined image size. We can make the switch to evaluation at the original size in a separate PR.

@djdameln I agree with this. I will resize the `masks` and evaluate at the user-defined image size 👍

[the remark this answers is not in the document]
jpcbertoldo: As @djdameln said, it's probably a good idea to make it simple for the sake of finishing this PR, but that's not ideal. In our latest paper (AUPIMO http://arxiv.org/abs/2401.01984) we specifically talk about this actually : ) because authors often use different ways to resize the masks and that creates artifacts.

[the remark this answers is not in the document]
ashwinvaidya17: Let's add a todo note + GitHub issue somewhere so that we don't forget it.

ashwinvaidya17 (approved):
Thanks for all the efforts! Apologies for the late review. I am fine with the changes. Just some minor comments.

PR #1706, opened 2024-02-02
🚀 Replace `albumentations` with `torchvision` transforms

djdameln:
## 📝 Description

PR for replacing albumentations transforms with torchvision transforms

## ✨ Changes

- Use torchvision v2 transforms instead of albumentations transforms throughout the code base.
- Load images as torch tensors instead of numpy images in dataset `__getitem__` methods.
- Use tv_tensors `Mask` format for GT masks
- Transforms are included in the exported model. This is done by wrapping the transform and the torch model in the new `InferenceModel` class before exporting.
- The datamodules take a Transform object as input to directly set the transform that is applied to the input images.
- The model implementer can define a default model-specific transform using the `configure_transforms` method of the Lit model. When no transform is passed to the datamodule, this default transform will be used. The default model transform may use the `image_size` parameter from the datamodule to resize the images to the size specified by the user.

## To be addressed:

- ~~Inferencers still use albumentations~~
- ~~Current design does not support model-specific transforms (i.e. defining a default transform from the lightning model).~~
- Still need to implement the model-specific default transforms for most models. This can be done later as they now just use the default one from the base module.
- ~~Docstrings in the data classes need to be updated to reflect the new signatures.~~
- ~~Patchcore fails to export because of the `CenterCrop` transform. We may have to implement our own version of this transform to allow exporting.~~
- ~~Some models still have the `input_size` parameter, which needs to be removed.~~
- ~~CLI tests are failing because the export API has changed.~~
- ~~The `ConfigAdapter` needs to be updated to reflect the new datamodules API.~~

- [x] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📚 Documentation update

## ✅ Checklist


src/anomalib/data/base/dataset.py
```diff
@@ -117,24 +118,21 @@ def __getitem__(self, index: int) -> dict[str, str | torch.Tensor]:
         mask_path = self._samples.iloc[index].mask_path
         label_index = self._samples.iloc[index].label_index
 
-        image = read_image(image_path)
+        image = to_tensor(Image.open(image_path))
```
samet-akcay, line 121: Would it be an idea to modify the `read_image` function to replace the cv2 stuff with Image.open? This would then be something like this

```python
image = read_image(image_path, to_tensor=True)
```

src/anomalib/data/utils/transforms.py
```diff
@@ -46,126 +50,45 @@ def get_transforms(
...
             Defaults to True.
 
-    Raises:
-        ValueError: When both ``config`` and ``image_size`` is ``None``.
-        ValueError: When ``config`` is not a ``str`` or `A.Compose`` object.
-
     Returns:
-        A.Compose: Albumentation ``Compose`` object containing the image transforms.
-
-    Examples:
```
samet-akcay, line 60: Would it be an idea to update this examples section?

src/anomalib/data/base/dataset.py
samet-akcay: Another question, now that we are reading the images with PIL, this would mean that we are reading based on RGB? Would this mean that we need to check the visualizer stuff that converts rgb-2-bgr?

samet-akcay (commented):
Thanks for creating this pr. I've added my initial comments.. Will go for another round

src/anomalib/data/base/dataset.py
willyfh: Maybe we also check some other files (notebooks, inference tools, etc.) where the `read_image` function is used.

Screenshot_20240203_082656_Chrome

willyfh: Same for `cv2.imread`

Screenshot_20240203_084231_Chrome

src/anomalib/data/utils/transforms.py
```diff
@@ -46,126 +50,45 @@ def get_transforms(
...
         if image_size is None:
-            msg = (
-                "Both config and image_size cannot be `None`. "
-                "Provide either config file to de-serialize transforms or image_size to get the default transformations"
-            )
+            msg = "Both config and image_size cannot be `None`."
             raise ValueError(msg)
         resize_height, resize_width = get_image_height_and_width(image_size)
-        transforms_list.append(A.Resize(height=resize_height, width=resize_width, always_apply=True))
+        transforms_list.append(v2.Resize(size=(resize_height, resize_width), interpolation=3, antialias=True))
```
willyfh, line 72: Is there any specific reason why the interpolation mode is set to 3 (Bicubic)? It seems that the default interpolation for the albumentation is `cv2.INTER_LINEAR`?

willyfh: Another comment, to improve code readibilty, maybe it would be better to use the enum `InterpolationMode` instead of using the integer value directly.

samet-akcay: @djdameln, won't be around for a week to answer, but as far as I remember, interpolation is one of the root cause of not getting good results on some models such as EfficientAD and WinCLIP

willyfh: @samet-akcay Ah, I see. If it is not affecting the speed much, then maybe there should be no problem. It may be less desirable in situations where speed is more important, or the usage of less complex interpolation is enough for the other models to get good results.

src/anomalib/data/base/dataset.py
```diff
@@ -117,24 +118,21 @@ def __getitem__(self, index: int) -> dict[str, str | torch.Tensor]:
...
-
-            transformed = self.transform(image=image, mask=mask)
-
-            item["image"] = transformed["image"]
+            mask = (
+                Mask(torch.zeros(image.shape[-2:]))
+                if label_index == 0
+                else Mask(to_tensor(Image.open(mask_path)).squeeze())
+            )
+            item["image"], item["mask"] = self.transform(image, mask)
```
ashwinvaidya17, line 134: How does this work? If we have transforms like normalise in the list then would it apply normalisation as well to the mask?

src/anomalib/data/base/video.py
```diff
@@ -148,22 +149,14 @@ def __getitem__(self, index: int) -> dict[str, str | torch.Tensor]:
...
-                self.transform(image=frame.numpy(), mask=mask)
-                for frame, mask in zip(item["image"], item["mask"], strict=True)
-            ]
-            item["image"] = torch.stack([item["image"] for item in processed_frames]).squeeze(0)
-            mask = torch.as_tensor(item["mask"])
-            item["mask"] = torch.stack([item["mask"] for item in processed_frames]).squeeze(0)
-            item["label"] = torch.Tensor([1 in frame for frame in mask]).int().squeeze(0)
+        if item.get("mask") is not None:
+            item["image"], item["mask"] = self.transform(item["image"], Mask(item["mask"]))
+            item["label"] = torch.Tensor([1 in frame for frame in item["mask"]]).int().squeeze(0)
```
ashwinvaidya17, line 154: Maybe not related to this PR but should we consider using torch.book for labels across the repo? We can save on some gpu memory.

src/anomalib/data/utils/augmenter.py
```diff
@@ -111,7 +111,7 @@ def generate_perturbation(
 
         # Load anomaly source image
         if anomaly_source_path:
-            anomaly_source_img = cv2.imread(anomaly_source_path)
+            anomaly_source_img = cv2.imread(str(anomaly_source_path))
```
ashwinvaidya17, line 114: Should we be converting the image to RGB? Would it affect the distribution of the anomalous image?

src/anomalib/data/utils/transforms.py
```diff
@@ -25,18 +29,18 @@ class InputNormalizationMethod(str, Enum):
...
 
 def get_transforms(
-    config: str | A.Compose | None = None,
+    config: str | v2.Compose | None = None,
     image_size: int | tuple[int, int] | None = None,
     center_crop: int | tuple[int, int] | None = None,
     normalization: InputNormalizationMethod = InputNormalizationMethod.IMAGENET,
-    to_tensor: bool = True,
-) -> A.Compose:
+    _to_tensor: bool = True,
```
ashwinvaidya17, line 36: If this is marked as unused, can't we just drop this?

ashwinvaidya17 (commented):
I didn't run this yet but overall I like this over albumentations. It will be one less dependency.

src/anomalib/data/base/dataset.py
djdameln: Torchvision automatically applies the right transforms to the right datatypes. Since mask is a `Mask` instance, normalization will be skipped. See https://pytorch.org/vision/main/auto_examples/transforms/plot_transforms_getting_started.html

src/anomalib/data/utils/augmenter.py
djdameln: I don't think it would make a lot of difference since the source images serve as random noise, but I agree that technically we should read the images as RGB. Let's address this separately.

src/anomalib/data/base/dataset.py
```diff
@@ -113,28 +134,29 @@ def __getitem__(self, index: int) -> dict[str, str | torch.Tensor]:
...
+            mask = (
+                Mask(torch.zeros(image.shape[-2:]))
+                if label_index == 0
+                else Mask(to_tensor(Image.open(mask_path)).squeeze())
+            )
+            if self.transform:
+                item["image"], item["mask"] = self.transform(image, mask)
+            else:
+                item["image"] = image
+                item["mask"] = mask
```
samet-akcay, line 158:
```suggestion
            item["image"], item["mask"] = self.transform(image, mask) if self.transform else (image, mask)

```

src/anomalib/data/base/depth.py
```diff
@@ -35,30 +38,36 @@ def __getitem__(self, index: int) -> dict[str, str | torch.Tensor]:
...
 
         if self.task == TaskType.CLASSIFICATION:
-            transformed = self.transform(image=image, depth_image=depth_image)
-            item["image"] = transformed["image"]
-            item["depth_image"] = transformed["depth_image"]
+            if self.transform:
+                item["image"], item["depth_image"] = self.transform(image, depth_image)
+            else:
+                item["image"] = image
+                item["depth_image"] = depth_image
```
samet-akcay, line 55: same as above?
```suggestion
            item["image"], item["depth_image"] = self.transform(image, depth_image) if self.transform else (image, depth_image)
```

src/anomalib/data/base/depth.py
```diff
@@ -35,30 +38,36 @@ def __getitem__(self, index: int) -> dict[str, str | torch.Tensor]:
...
             # Therefore, create empty mask for Normal (0) images.
-            mask = np.zeros(shape=image.shape[:2]) if label_index == 0 else cv2.imread(mask_path, flags=0) / 255.0
-
-            transformed = self.transform(image=image, depth_image=depth_image, mask=mask)
-
-            item["image"] = transformed["image"]
-            item["depth_image"] = transformed["depth_image"]
+            mask = (
+                Mask(torch.zeros(image.shape[-2:]))
+                if label_index == 0
```
samet-akcay, line 61:
```suggestion
                if label_index == LabelName.NORMAL
```

src/anomalib/data/base/depth.py
```diff
@@ -35,30 +38,36 @@ def __getitem__(self, index: int) -> dict[str, str | torch.Tensor]:
...
+                else Mask(to_tensor(Image.open(mask_path)).squeeze())
+            )
+            if self.transform:
+                item["image"], item["depth_image"], item["mask"] = self.transform(image, mask)
+            else:
+                item["image"] = image
+                item["depth_image"] = depth_image
+                item["mask"] = mask
```
samet-akcay, line 69: same as above, how about using a conditional expression

src/anomalib/data/image/folder.py
```diff
@@ -251,8 +249,8 @@ def __init__(
         self.mask_dir = mask_dir
         self.extensions = extensions
 
-    def _setup(self) -> None:
-        """Assign samples."""
+        # def _setup(self) -> None:
+        #     """Assign samples."""
```
samet-akcay, line 253: is this intentional?

src/anomalib/data/utils/augmenter.py
samet-akcay: maybe on another pr, would it be possible to use PIL Image for this as well?

src/anomalib/data/base/depth.py
```diff
@@ -35,30 +38,36 @@ def __getitem__(self, index: int) -> dict[str, str | torch.Tensor]:
...
-        label_index = self._samples.iloc[index].label_index
-        depth_path = self._samples.iloc[index].depth_path
+        image_path = self.samples.iloc[index].image_path
+        mask_path = self.samples.iloc[index].mask_path
+        label_index = self.samples.iloc[index].label_index
+        depth_path = self.samples.iloc[index].depth_path
 
-        image = read_image(image_path)
-        depth_image = read_depth_image(depth_path)
+        image = to_tensor(Image.open(image_path))
```
samet-akcay, line 46: can we use `from torchvision.io import read_image` for this to create tensor directly

src/anomalib/deploy/export.py
```diff
@@ -47,20 +48,70 @@ class ExportType(str, Enum):
...
 
 
+class InferenceModel(nn.Module):
+    """Inference model for export.
+
+    The InferenceModel is used to wrap the model and transform for exporting to torch and ONNX/OpenVINO.
+
+    Args:
+        model (nn.Module): Model to export.
+        transform (Compose): Input transform for the model.
```
samet-akcay, line 58: can you add this and explain what it is `disable_antialias`

src/anomalib/deploy/export.py
```diff
@@ -47,20 +48,70 @@ class ExportType(str, Enum):
...
     """Export AnomalibModel to torch.
 
     Args:
         model (AnomalyModule): Model to export.
         export_root (Path): Path to the output folder.
-        transform (dict[str, Any] | AnomalibDataset | AnomalibDataModule | A.Compose): Data transforms (augmentations)
-            used for the model. When using ``dict``, ensure that the transform dict is in the format required by
-            Albumentations.
+        transform (Compose): Input transforms used for the model. If not provided, the transform is taken from the
+            model.
```
samet-akcay, line 114:
```suggestion
            model.
            Defaults to ``None``.
```

src/anomalib/deploy/export.py
```diff
@@ -188,10 +235,8 @@ def export_to_openvino(
     Args:
         export_root (Path): Path to the export folder.
         model (AnomalyModule): AnomalyModule to export.
-        input_size (tuple[int, int]): Input size of the model. Used for adding metadata to the IR.
-        transform (dict[str, Any] | AnomalibDataset | AnomalibDataModule | A.Compose): Data transforms (augmentations)
-            used for the model. When using dict, ensure that the transform dict is in the format required by
-            Albumentations.
+        transform (Compose): Input transforms used for the model. If not provided, the transform is taken from the
+            model.
```
samet-akcay, line 239:
```suggestion
            model.
            Defaults to ``None``.
```

tests/unit/engine/test_setup_transform.py
```diff
@@ -0,0 +1,197 @@
+"""Tests for the Anomalib Engine."""
+
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+
```
samet-akcay, line 6:
```suggestion
```

samet-akcay (changes requested):
thanks, massive effort! Got couple of questions/comments

src/anomalib/data/base/datamodule.py
```diff
@@ -74,6 +75,10 @@ def __init__(
         val_split_ratio: float,
         test_split_mode: TestSplitMode | str | None = None,
         test_split_ratio: float | None = None,
+        image_size: tuple[int, int] | None = None,
+        transform: Transform | None = None,
```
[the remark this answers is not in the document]
djdameln: I opted for `transform` to keep it consistent with the torchvision datasets API, see https://pytorch.org/vision/stable/datasets.html

src/anomalib/data/base/depth.py
djdameln: Unfortunately this does not work for all datasets. `torchvision.io` only supports `.png` and `.jpg` images, and some of our datasets have different image extensions (e.g. `.bmp` or `.tiff`).

samet-akcay: ah, yeah, would have been nice

src/anomalib/data/base/dataset.py
```diff
@@ -35,16 +36,38 @@
...
+    Note:
+        The example above is illustrative and may need to be adjusted based on the specific dataset structure.
+
     Args:
         task (str): Task type, either 'classification' or 'segmentation'
         transform (A.Compose): Albumentations Compose object describing the transforms that are applied to the inputs.
     """
 
-    def __init__(self, task: TaskType, transform: A.Compose) -> None:
+    def __init__(self, task: TaskType, transform: Transform) -> None:
```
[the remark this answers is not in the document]
djdameln: Do we still use the `v2.Transform` format anywhere? I personally prefer importing the transforms directly without the `v2` prefix. It would also make it easier to switch once the prefix will be deprecated.

src/anomalib/data/base/datamodule.py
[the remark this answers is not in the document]
samet-akcay: looks like torchvision does not have a consistent strategy for this either. Some datasets have transforms down in the page. 

Alright, let's stick to transform then

src/anomalib/data/base/dataset.py
samet-akcay: Can we use `read_image` here too?
```suggestion
        image = read_image(image_path)
```

src/anomalib/data/video/ucsd_ped.py
```diff
@@ -111,7 +108,7 @@ def get_mask(self, idx: int) -> np.ndarray | None:
         mask_frames = sorted(Path(mask_folder).glob("*.bmp"))
         mask_paths = [mask_frames[idx] for idx in frames.int()]
 
-        return np.stack([cv2.imread(str(mask_path), flags=0) / 255.0 for mask_path in mask_paths])
+        return torch.stack([read_image(mask_path, as_tensor=True) for mask_path in mask_paths]).squeeze()
```
samet-akcay, line 111: would it be an idea to have `read_mask` function that is similar to `read_image` but also does the `squeeze()` stuff. I see that `squeeze` is also done in the base dataset. Doing that in the function would be safer.

src/anomalib/utils/visualization/image.py
```diff
@@ -102,7 +102,8 @@ def _visualize_batch(self, batch: dict) -> Iterator[GeneratorResult]:
         for i in range(batch_size):
             if "image_path" in batch:
                 height, width = batch["image"].shape[-2:]
-                image = read_image(path=batch["image_path"][i], image_size=(height, width))
+                image = read_image(path=batch["image_path"][i])
+                image = cv2.resize(image, dsize=(width, height), interpolation=cv2.INTER_AREA)
```
samet-akcay, line 106: would it be an idea to keep the resize option in `read_image`. So we would not use cv2?

With PIL we could do it with `image.resize(new_size)` in `read_image`?

ashwinvaidya17 (approved):
Thanks for the efforts! This was a much needed improvement

src/anomalib/utils/visualization/image.py
djdameln: Let's address this later.

samet-akcay (approved):
Looking great! Thanks a lot!

PR #1726, opened 2024-02-09
PIMO

### LINKED ISSUE #1728: Add PIMO metric to anomalib
# Description: 

- ArXiv: https://arxiv.org/abs/2401.01984
- Medium: https://medium.com/p/c653ac30e802
- GSoC: https://gist.github.com/jpcbertoldo/12553b7eaa97cfbf3e55bfd7d1cafe88

Assignee: @jpcbertoldo

jpcbertoldo:
## 📝 Description

Replace https://github.com/openvinotoolkit/anomalib/pull/1557, which replaces the PRs from https://gist.github.com/jpcbertoldo/12553b7eaa97cfbf3e55bfd7d1cafe88 .

Implements refactors from https://github.com/jpcbertoldo/anomalib/blob/metrics/refactors/src/anomalib/utils/metrics/perimg/.refactors .

arxiv: https://arxiv.org/abs/2401.01984
medium post: https://medium.com/p/c653ac30e802
GSoC deliverable: https://gist.github.com/jpcbertoldo/12553b7eaa97cfbf3e55bfd7d1cafe88

Closes #1728 1728

## ✨ Changes

- [X] 🚀 New feature (non-breaking change which adds functionality)
- [ ] 📚 Documentation update
- [ ] 🔒 Security update

## ✅ Checklist

- [ ] 📋 **I have summarized my changes in the** [CHANGELOG](https://github.com/openvinotoolkit/anomalib/blob/main/CHANGELOG.md) and followed the guidelines for my type of change (skip for minor changes, documentation updates, and test enhancements).

For more information about code review checklists, see the [Code Review Checklist](../docs/source/markdown/guides/developer/code_review_checklist.md).


src/anomalib/metrics/per_image/binclf_curve.py
```diff
@@ -0,0 +1,178 @@
...
+# Original Code
+# Copyright (c) 2024 @jpcbertoldo
+# https://github.com/jpcbertoldo/aupimo
+# SPDX-License-Identifier: MIT
+#
+# Modified
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
```
jpcbertoldo, line 23: @ashwinvaidya17 
> Can we avoid this? Last I checked `__future__.annotations` does not behave well with jsonargparse

i thought it was necessary for the annotations like `range: tuple[float, float]`, is it not?

src/anomalib/metrics/per_image/_binclf_curve_numba.py
```diff
@@ -0,0 +1,119 @@
+"""Binary classification matrix curve (NUMBA implementation of low level functions).
```
jpcbertoldo, line 1: > What's the convention behind naming files starting with `_`?

it's similar to the _ attributes inside a module

a function _func in module.py is supposed to be used within the module scope, right?

similarly, the _validate is private to per_image/, so it can be used across other modules inside the subpackage per_image but not out of it

src/anomalib/metrics/per_image/_validate.py
```diff
@@ -0,0 +1,428 @@
...
+    lower, upper = bounds
+    rate(lower, zero_ok=False, one_ok=False)
+    rate(upper, zero_ok=False, one_ok=True)
+
+    if lower >= upper:
+        msg = f"Expected the upper bound to be larger than the lower bound, but got {upper=} <= {lower=}"
+        raise ValueError(msg)
+
+
+def file_path(file_path: str | Path, must_exist: bool, extension: str | None, pathlib_ok: bool) -> None:
```
jpcbertoldo, line 105: @ashwinvaidya17 
> Maybe we can move this method in data.utils.

i am trying to avoid things out of per_image for the sake of synchronizing the standalone repo more easily
could we eventually put this in a list of "refactors for later"?

src/anomalib/metrics/per_image/_validate.py
```diff
@@ -0,0 +1,428 @@
...
+
+    if extension is None:
+        return
+
+    if file_path.suffix != extension:
+        msg = f"Expected file path to have extension '{extension}', but got '{file_path.suffix}'"
+        raise ValueError(msg)
+
+
+def file_paths(file_paths: list[str | Path], must_exist: bool, extension: str | None, pathlib_ok: bool) -> None:
```
jpcbertoldo, line 142: @ashwinvaidya17 
> same here and maybe we can rename it to validate_path_are_files. A bit verbose but improves readibility imo.

so the idea of the _validate module is to have short names because the "validate" part is common to all the methods; the functions are always called with _validate.file_paths() to make the "validate" show up

sounds logical?

src/anomalib/metrics/per_image/binclf_curve_numpy.py
```diff
@@ -0,0 +1,439 @@
...
+
+from . import _validate
+
+logger = logging.getLogger(__name__)
+
+# =========================================== CONSTANTS ===========================================
+
+
+@dataclass
+class BinclfAlgorithm:
```
jpcbertoldo, line 44: @ashwinvaidya17 
> What's the advantage of using this over enums?

i just wanted to avoid enums for the sake of simplicity, yet i wanted to put these contants together for consistency
should i switch to enum or is that ok?

src/anomalib/metrics/per_image/pimo_numpy.py
```diff
@@ -0,0 +1,461 @@
...
+        shared_fpr,
+        fpr_upper_bound,
+    )
+
+    if not np.isclose(fpr_lower_bound_defacto, fpr_lower_bound, rtol=(rtol := 1e-2)):
+        msg = (
+            "The lower bound of the shared FPR integration range is not exactly achieved. "
+            f"Expected {fpr_lower_bound} but got {fpr_lower_bound_defacto}, which is not within {rtol=}."
+        )
+        warnings.warn(msg, RuntimeWarning, stacklevel=1)
```
jpcbertoldo, line 279: @ashwinvaidya17 
> What's the advantage of using both logger.warning and warnings.warn. What happens if we want to pipe the warnings to a file and keep the console empty? A file handler can be passed to the logging module in such a scenario.

tbh i didnt know what was the right policy here 😳
i put both cuz i was sure you'd find and tell me what to do :P hehe

so should i use the logger?

on a side note: warnings.warn allows one to make this raise an exception with warnings.filterwarnings("error");
maybe that could be interesting?

for context: this warning is due to having too few points to integrate the AUC curve, which may result in unprecise AUC values

src/anomalib/metrics/per_image/pimo_numpy.py
```diff
@@ -0,0 +1,461 @@
...
+        logger.warning(msg)
+
+    aucs: ndarray = np.trapz(per_image_tprs_bounded, x=shared_fpr_bounded_log, axis=1)
+
+    # normalize, then clip(0, 1) makes sure that the values are in [0, 1] in case of numerical errors
+    normalization_factor = aupimo_normalizing_factor(fpr_bounds)
+    aucs = (aucs / normalization_factor).clip(0, 1)
+
+    return threshs, shared_fpr, per_image_tprs, image_classes, aucs, num_points_integral
```
jpcbertoldo, line 363: @ashwinvaidya17 
> Do you think we should capture these in a dataclass?

there is a dataclass for the torch interface but not for the numpy interface
i couldnt figure out a nice, maintainable way to make the two versions
perhaps we can discuss this in a call? i'll explain better and maybe you'll know a good solution

jpcbertoldo:
another unresolved issue from the previous PR

[about "ATTENTION..." in docstrings]

@ashwinvaidya17 
> Same here. We need to consider how these docstrings will be rendered in sphinx.

how can i check that?

samet-akcay:
> another unresolved issue from the previous PR
> 
> [about "ATTENTION..." in docstrings]
> 
> @ashwinvaidya17
> 
> > Same here. We need to consider how these docstrings will be rendered in sphinx.
> 
> how can i check that?

The documentation is built here based on your changes

src/anomalib/metrics/per_image/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+"""
+
+# Original Code
+# Copyright (c) 2024 @jpcbertoldo
+# https://github.com/jpcbertoldo/aupimo
+# SPDX-License-Identifier: MIT
+#
+# Modified
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
```
samet-akcay, line 13: this is a good idea, but we might need to double check with GSoC guidelines in regards to who owns the code.

jpcbertoldo:
> > another unresolved issue from the previous PR
> > [about "ATTENTION..." in docstrings]
> > @ashwinvaidya17
> > > Same here. We need to consider how these docstrings will be rendered in sphinx.
> > 
> > how can i check that?
> 
> The documentation is built here based on your changes

there are some [metadata] stuff showing up, idk why

apparently sphinx doesnt like dataclasses? 
https://anomalib--1726.org.readthedocs.build/en/1726/markdown/guides/reference/metrics/index.html#id5
this field is in the class `AUPIMOResult` but it's showing as if it was a function or w/e in the root (?)
while `AUPIMOResult` doesnt show at all

samet-akcay:
> > > another unresolved issue from the previous PR
> > > [about "ATTENTION..." in docstrings]
> > > @ashwinvaidya17
> > > > Same here. We need to consider how these docstrings will be rendered in sphinx.
> > > 
> > > 
> > > how can i check that?
> > 
> > 
> > The documentation is built here based on your changes
> 
> there are some [metadata] stuff showing up, idk why
> 
> apparently sphinx doesnt like dataclasses? https://anomalib--1726.org.readthedocs.build/en/1726/markdown/guides/reference/metrics/index.html#id5 this field is in the class `AUPIMOResult` but it's showing as if it was a function or w/e in the root (?) while `AUPIMOResult` doesnt show at all

Tree structure is also messed up a bit. It might be an idea to split each metric into a separate section.

jpcbertoldo:
> > another unresolved issue from the previous PR
> > [about "ATTENTION..." in docstrings]
> > @ashwinvaidya17
> > > Same here. We need to consider how these docstrings will be rendered in sphinx.
> > 
> > 
> > how can i check that?
> 
> The documentation is built here based on your changes

https://anomalib--1726.org.readthedocs.build/en/1726/markdown/guides/reference/metrics/index.html

it's not quite working as expected

1) i expected `per_image` to show as submenu in metrics, how could I do that?

2) it seems not to like dataclasses; there are attributes of `PIMOResult` and `AUPIMOResult` showing as if it was a function (?) and the classes themselves dont' show

src/anomalib/metrics/per_image/binclf_curve.py
jpcbertoldo: So, (i think) i re-figured out why i used this. 
In the other file, the class `AUPIMOResult` has annotations using the class itself (eg `from_pimoresult(...) -> AUPIMOResult:`) and apparently the linter didnt like it, but adding `__future__.annotations` solves it.
Now, idk if this is a good justification. What do you prefer to do about this?

ashwinvaidya17 (changes requested):
A lot has changed since this PR was submitted, but I've finally gotten around to reviewing it. This is a huge PR with a lot of efforts behind it. However, I have some concerns. I've gone over it once but I think I'll require a few more passes for a more thorough review. But meanwhile we can start the discussions for the current opens.

src/anomalib/metrics/per_image/_binclf_curve_numba.py
```diff
@@ -0,0 +1,119 @@
...
+
+See docstring of `binclf_curve` or `binclf_curve_numpy` for more details.
+
+author: jpcbertoldo
+"""
+
+# Original Code
+# Copyright (c) 2024 @jpcbertoldo
+# https://github.com/jpcbertoldo/aupimo
+# SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
jpcbertoldo: @samet-akcay do you know?

requirements/base.txt
```diff
@@ -16,3 +16,4 @@ torch>=2,<2.2.0 # rkde export fails even with ONNX 17 (latest) with torch 2.2.0.
 torchmetrics==0.10.3
 rich-argparse
 open-clip-torch>=2.23.0
+numba>=0.58.1
```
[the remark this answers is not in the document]
jpcbertoldo: how should i do that? like create another optional install in the `setup.py`?

src/anomalib/metrics/per_image/binclf_curve.py
```diff
@@ -0,0 +1,178 @@
+"""Binary classification curve (torch interface).
```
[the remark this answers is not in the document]
jpcbertoldo: i tested all these variants back then
numba gives a good improvement, it is not neglectable
as for the tensor-ndarray-tensor conversion, it is also faster than using only torch

(from what i saw few months ago in the code and my own tests)
torchmetrics has two methods for computing this (both with cuda available)
A)  a fast one that computes the curves with all thresholds possible (ie all score values seen)
B) a slower one that takes pre-defined threholds

pimo needs B because all images are evaluated at the same thresholds
in that case, the methods here (even without numba) are faster because torch's and torchmetrics' stuff dont deal with computing same thresholds on instance-scoped binary classification matrices

i can find detailed results from my test again, but just for a reference (in the paper)

image

that numba method (+ the tensor-ndarray conversion + just-in-time compilation) is beating torchmetrics AUROC on GPU

src/anomalib/metrics/per_image/utils.py
```diff
@@ -0,0 +1,413 @@
...
+    outliers_policy: str | None = StatsOutliersPolicy.NONE,
+    repeated_policy: str | None = StatsRepeatedPolicy.AVOID,
+    repeated_replacement_atol: float = 1e-2,
+) -> list[dict[str, str | int | float]]:
+    """Torch-oriented interface for `per_image_scores_stats`. See its dscription for more details (below).
+
+    Numpy version docstring
+    =======================
+
+    {docstring}
```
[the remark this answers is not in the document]
jpcbertoldo: it is, the documentation is built programatically from the the other function because this is just a wrapper
i wasnt sure how this would interact with the documentation builder tho
i wanted to avoid repeating documentation to avoid mistakes

src/anomalib/metrics/per_image/binclf_curve_numpy.py
```diff
@@ -0,0 +1,439 @@
...
+    """Algorithm to use."""
+
+    PYTHON: ClassVar[str] = "python"
```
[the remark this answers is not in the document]
jpcbertoldo: the logic is that the low level operations (that make it faster) (L153-L157) are just python-native operations, not even numpy
and both versions use numpy in a way or another

src/anomalib/metrics/per_image/_binclf_curve_numba.py
```diff
@@ -0,0 +1,119 @@
...
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import numba
+import numpy as np
+from numpy import ndarray
+
+
+@numba.jit(nopython=True)
+def binclf_one_curve_numba(scores: ndarray, gts: ndarray, threshs: ndarray) -> ndarray:
```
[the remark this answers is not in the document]
jpcbertoldo: not really
the python version has `itertools.takewhile` and this one has an actual while loop
many things are not recognized by or not compatible with numba's compiler (eg `itertools.takewhile`)
in general, numba is a bit sensitve, it's very easy to break it so that kind of solution is generally not a great idea
i made sure (eg validations and in the higher level functions) that this one does work

tests/unit/metrics/per_image/test_utils.py
```diff
@@ -0,0 +1,312 @@
...
+    compare_models_pairwise_ttest_rel(
+        scores_per_model,
+        alternative=alternative,
+        higher_is_better=higher_is_better,
+    )
+
+
+def test_compare_models_pairwise_wilcoxon(scores_per_model: dict, alternative: str, higher_is_better: bool) -> None:
+    """Test `compare_models_pairwise_wilcoxon`."""
+    from anomalib.metrics.per_image import AUPIMOResult, compare_models_pairwise_wilcoxon
```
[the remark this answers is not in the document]
jpcbertoldo: i reverted this to usual imports on top of the file

src/anomalib/metrics/per_image/utils.py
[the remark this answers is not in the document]
jpcbertoldo: i reverted this

ashwinvaidya17 (changes requested):
Thanks! I think we are almost there. I have two major comments remaining. First, the `author` tag in the headers is not consistent with the rest of the repo. The current scans check `third-party-programs.txt` to verify the license and proper attribution for codes taken from other repositories. Second, `numba` adds complexity to the codebase. If it is not strictly necessary, let's not include it. But I am happy to hear other opinions.

jpcbertoldo:
> Thanks! I think we are almost there. I have two major comments remaining. First, the `author` tag in the headers is not consistent with the rest of the repo. The current scans check `third-party-programs.txt` to verify the license and proper attribution for codes taken from other repositories. Second, `numba` adds complexity to the codebase. If it is not strictly necessary, let's not include it. But I am happy to hear other opinions.

I removed the author tag and add the original code's repo to `third-party-programs.txt`.

I think the decision for `numba` is rather up to y`all as maintainers, but it does give a nice speedup. 

For the record (if i get it well), it only has two requirements: 

```python
install_requires = [
    'llvmlite >={},<{}'.format(min_llvmlite_version, max_llvmlite_version),
    'numpy >={},<{}'.format(min_numpy_run_version, max_numpy_run_version),
    'importlib_metadata; python_version < "3.9"',
]
```
https://github.com/numba/numba/blob/d4460feb8c91213e7b89f97b632d19e34a776cd3/setup.py#L369-L373

```
min_numpy_run_version = "1.22"
max_numpy_run_version = "1.27"
min_llvmlite_version = "0.41.0dev0"
max_llvmlite_version = "0.42"
```
https://github.com/numba/numba/blob/d4460feb8c91213e7b89f97b632d19e34a776cd3/setup.py#L25-L28

I copied these from `numba==0.58.1`, which is the minimum here.

djdameln (commented):
I'm getting the following error when I try to use the AUPIMO metric to evaluate an anomalib model. Can you check it out?

```
ValueError: The `.compute()` return of the metric logged as 'pixel_AUPIMO' must be a tensor. Found (PIMOResult(shared_fpr_metric='mean-per-image-fpr'), AUPIMOResult(shared_fpr_metric='mean-per-image-fpr', fpr_lower_bound=1e-05, fpr_upper_bound=0.0001, num_threshs=48480))
```

src/anomalib/metrics/per_image/pimo.py
```diff
@@ -0,0 +1,981 @@
...
+        paths: paths to the source images to which the AUPIMO scores correspond.
+
+    Returns:
+        tuple[PIMOResult, AUPIMOResult]: PIMO and AUPIMO results dataclass objects. See `PIMOResult` and `AUPIMOResult`.
+    """
+    _validate_is_anomaly_maps(anomaly_maps)
+    anomaly_maps_array = anomaly_maps.detach().cpu().numpy()
+
+    _validate_is_masks(masks)
+    masks_array = masks.detach().cpu().numpy()
```
jpcbertoldo, line 671:
```suggestion
    anomaly_maps_array = anomaly_maps.detach().cpu().numpy()
    masks_array = masks.detach().cpu().numpy()
```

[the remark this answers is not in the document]
jpcbertoldo: done

src/anomalib/__init__.py
```diff
@@ -8,6 +8,14 @@
 __version__ = "1.2.0dev"
 
 
+try:
+    import numba  # noqa: F401
+except ImportError:
+    HAS_NUMBA = False
+else:
+    HAS_NUMBA = True
```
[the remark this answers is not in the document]
jpcbertoldo: i dont personally dont see the difference, it's kind of arbitrary, no?
should i move it there?

src/anomalib/metrics/per_image/_binclf_curve_numba.py
```diff
@@ -0,0 +1,117 @@
...
+"""
+
+# Original Code
+# https://github.com/jpcbertoldo/aupimo
+#
+# Modified
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import numba
```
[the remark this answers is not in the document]
jpcbertoldo: the pytorch-only was slower than the one with numpy (which is slower than numba)
i see this is becoming a long issue, i can just drop numba, but can you confirm this is the decision to make? (just to avoid going back later)

jpcbertoldo:
> I'm getting the following error when I try to use the AUPIMO metric to evaluate an anomalib model. Can you check it out?
> 
> ```
> ValueError: The `.compute()` return of the metric logged as 'pixel_AUPIMO' must be a tensor. Found (PIMOResult(shared_fpr_metric='mean-per-image-fpr'), AUPIMOResult(shared_fpr_metric='mean-per-image-fpr', fpr_lower_bound=1e-05, fpr_upper_bound=0.0001, num_threshs=48480))
> ```

This is normal because aupimo returns many values so it was encapsulated in that dataclass.
We could create an option in the torchmetrics interface to optionally return this and (by default) return just the average aupimo instead. 
Sounds good?

src/anomalib/metrics/per_image/pimo_numpy.py
```diff
@@ -0,0 +1,420 @@
...
+
+logger = logging.getLogger(__name__)
+
+# =========================================== CONSTANTS ===========================================
+
+
+class PIMOSharedFPRMetric(Enum):
+    """Shared FPR metric (x-axis of the PIMO curve)."""
+
+    MEAN_PERIMAGE_FPR: str = "mean-per-image-fpr"
```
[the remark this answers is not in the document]
djdameln: I would be in favor of removing it. If we extend the metric with more possible values in the future we can always bring it back

djdameln:
> This is normal because aupimo returns many values so it was encapsulated in that dataclass.
> We could create an option in the torchmetrics interface to optionally return this and (by default) return just the average aupimo instead.
> Sounds good?

I think that would be a good idea. With the default setting the metric should be fully compatible with Anomalib's pipeline. So users should be able to enable it from the config/cli or API to have Anomalib report the average AUPIMO value.

src/anomalib/metrics/per_image/_binclf_curve_numba.py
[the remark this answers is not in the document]
djdameln: @jpcbertoldo would it be possible to use the pytorch-only implementation by default, and use the Numba implementation if the user has Numba installed in their environment? This way we don't force casual users to install another dependency, and advanced users who care more about speed can install Numba to speed up the computation. We could log a warning to notify the users that the computation can be made faster by installing Numba.

src/anomalib/__init__.py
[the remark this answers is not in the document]
jpcbertoldo: moved it

src/anomalib/metrics/per_image/pimo_numpy.py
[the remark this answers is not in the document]
jpcbertoldo: done

src/anomalib/metrics/per_image/binclf_curve.py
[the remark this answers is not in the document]
jpcbertoldo: i am resolving this one because it is a duplicate of https://github.com/openvinotoolkit/anomalib/pull/1726#discussion_r1654663073

src/anomalib/metrics/per_image/_binclf_curve_numba.py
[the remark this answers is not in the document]
jpcbertoldo: @djdameln this is already the case

this https://github.com/openvinotoolkit/anomalib/blob/440bf2f5f32c2a9230485a6df11ce3b026efa306/src/anomalib/metrics/per_image/binclf_curve_numpy.py#L23-L32 is checking if numba is installed and only importing the the numba-dependent version if it is the case

and this part https://github.com/openvinotoolkit/anomalib/blob/440bf2f5f32c2a9230485a6df11ce3b026efa306/src/anomalib/metrics/per_image/binclf_curve_numpy.py#L229-L239  is doing this version routing (with or without numba)

the installation of numba is already optional as well and it requires "opt-in" from the user: https://github.com/openvinotoolkit/anomalib/blob/440bf2f5f32c2a9230485a6df11ce3b026efa306/pyproject.toml#L87

@ashwinvaidya17 do you think that is ok like this?

jpcbertoldo:
> > This is normal because aupimo returns many values so it was encapsulated in that dataclass.
> > We could create an option in the torchmetrics interface to optionally return this and (by default) return just the average aupimo instead.
> > Sounds good?
> 
> I think that would be a good idea. With the default setting the metric should be fully compatible with Anomalib's pipeline. So users should be able to enable it from the config/cli or API to have Anomalib report the average AUPIMO value.

@djdameln Done : ) 

i think the only missing issue is about numba (here https://github.com/openvinotoolkit/anomalib/pull/1726#discussion_r1654663073)

i can remove it if that's better, but it is already optional like you suggested in your last comment

samet-akcay:
> @samet-akcay could you launch a code check here please?
> 
> About [CDO](https://github.com/openvinotoolkit/anomalib/pull/1726/checks?check_run_id=29066507534), I think it is ok to apply the correction? But I'm not 100% sure (don't want to mess up the commit history 😬 )

done

src/anomalib/metrics/per_image/_binclf_curve_numba.py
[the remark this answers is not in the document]
jpcbertoldo: @samet-akcay can we give a final push to decide about keeping `numba` or not?
i think it is the only thing blocking the merge
the code doesn't depend on it, for the moment it is optional to install/use it, so the question is if we keep it or just get rid of it

src/anomalib/metrics/per_image/binclf_curve_numpy.py
```diff
@@ -0,0 +1,430 @@
...
+
+from . import _validate
+
+logger = logging.getLogger(__name__)
+
+# =========================================== CONSTANTS ===========================================
+
+
+class BinclfAlgorithm(Enum):
+    """Algorithm to use (relates to the low-level implementation)."""
```
samet-akcay, line 42: is the name algorithm correct here? Do you mean the algorithm used by Python and numba?

src/anomalib/metrics/per_image/binclf_curve_numpy.py
```diff
@@ -0,0 +1,430 @@
...
+            "Notice that the performance will be slower. Consider installing Numba for faster computation.",
+        )
+
+    return _binclf_multiple_curves_python(scores_batch, gts_batch, threshs)
+
+
+# ========================================= PER-IMAGE BINCLF CURVE =========================================
+
+
+def _get_threshs_minmax_linspace(anomaly_maps: ndarray, num_threshs: int) -> ndarray:
```
samet-akcay, line 245: Not sure if using threshs over thresholds would massively reduce the verbosity. As a reader, i would prefer the full for clarity.

```suggestion
def _get_threshs_minmax_linspace(anomaly_maps: ndarray, num_thresholds: int) -> ndarray:
```

src/anomalib/metrics/per_image/pimo.py
```diff
@@ -0,0 +1,956 @@
...
+        num_threshs_auc=num_threshs_auc,
+        aupimos=aupimos,
+    )
+    return pimoresult, aupimoresult
+
+
+# =========================================== TORCHMETRICS ===========================================
+
+
+class PIMO(Metric):
```
samet-akcay, line 704: As a reader I would expect this class almost at the top. I feel we might need to split this module into multiple sub-modules to make easier to follow

src/anomalib/metrics/per_image/pimo.py
```diff
@@ -0,0 +1,956 @@
...
+        masks = torch.concat(self.masks, dim=0)
+        return pimo_curves(
+            anomaly_maps,
+            masks,
+            self.num_threshs,
+            binclf_algorithm=self.binclf_algorithm,
+        )
+
+
+class AUPIMO(PIMO):
```
samet-akcay, line 824: would it be an idea to have a dedicated module for this, even maybe sub-package if needed. 

Option I
```bash
metrics
  per_image
    aupimo.py
```

Option II
```bash
metrics
  per_image
    aupimo
      aupimo.py
      utils..
```

src/anomalib/metrics/per_image/pimo_numpy.py
```diff
@@ -0,0 +1,410 @@
...
+def _joint_validate_threshs_shared_fpr(threshs: ndarray, shared_fpr: ndarray) -> None:
+    if threshs.shape[0] != shared_fpr.shape[0]:
+        msg = (
+            "Expected `threshs` and `shared_fpr` to have the same number of elements, "
+            f"but got {threshs.shape[0]} != {shared_fpr.shape[0]}"
+        )
+        raise ValueError(msg)
+
+
+# =========================================== PIMO ===========================================
```
samet-akcay, line 68: Looks like these validators are used only once. Is it needed to create a function for this logic? Maybe just validate it within the place, which is `thresh_at_shared_fpr_level` in this case?

src/anomalib/metrics/per_image/utils.py
```diff
@@ -0,0 +1,521 @@
...
+        per_image_scores_array,
+        images_classes_array,
+        only_class=only_class,
+        outliers_policy=outliers_policy,
+        repeated_policy=repeated_policy,
+        repeated_replacement_atol=repeated_replacement_atol,
+    )
+
+
+def compare_models_pairwise_ttest_rel(
```
samet-akcay, line 319: Is this used anywhere within the codebase?

samet-akcay: Let's say a user wants to use this, how would they use this? For example, how to get `scores_per_model: dict[str, Tensor]`?

src/anomalib/metrics/per_image/utils.py
```diff
@@ -0,0 +1,521 @@
...
+        )
+        for model_name, scores in scores_per_model.items()
+    ]
+    cls = OrderedDict if isinstance(scores_per_model, OrderedDict) else dict
+    scores_per_model_with_arrays = cls(scores_per_model_items)
+
+    return utils_numpy.compare_models_pairwise_ttest_rel(scores_per_model_with_arrays, alternative, higher_is_better)
+
+
+def compare_models_pairwise_wilcoxon(
```
samet-akcay, line 387: same as above

src/anomalib/metrics/per_image/utils_numpy.py
```diff
@@ -0,0 +1,481 @@
...
+
+    None | "none": do not include outliers.
+    "hi": only include high outliers.
+    "lo": only include low outliers.
+    "both": include both high and low outliers.
+    """
+
+    NONE: str = "none"
+    HI: str = "hi"
+    LO: str = "lo"
```
samet-akcay, line 39: why not 
HIGH, instead of HI, and LOW, instead LO?

src/anomalib/metrics/per_image/utils_numpy.py
```diff
@@ -0,0 +1,481 @@
...
+                - 'q3': 75th percentile of the scores.
+                - 'whishi': Upper whisker value.
+                - 'whislo': Lower whisker value.
+                - 'outlo_i': low outlier value; `i` is a unique index for each low outlier.
+                - 'outhi_j': high outlier value; `j` is a unique index for each high outlier.
+            - 'stat_value': Value of the statistic (same units as `values`).
+            - 'image_idx': Index of the image in `per_image_scores` whose score is the closest to the statistic's value.
+            - 'score': The score of the image at index `image_idx` (not necessarily the same as `stat_value`).
+
+        The list is sorted by increasing `stat_value`.
```
samet-akcay, line 208: As a reader/user, I would love to see an example to understand how to use this

src/anomalib/metrics/per_image/pimo.py
```diff
@@ -0,0 +1,956 @@
...
+    masks = torch.concat(masks, dim=0)
+    device = masks.device
+    image_classes = pimo_numpy._images_classes_from_masks(masks.numpy())  # noqa: SLF001
+    return torch.from_numpy(image_classes, device=device)
+
+
+# =========================================== ARGS VALIDATION ===========================================
+
+
+def _validate_is_anomaly_maps(anomaly_maps: Tensor) -> None:
```
samet-akcay, line 74: would it be an idea to move these validation utils to _validate.py? these functions already use functions from validate.

src/anomalib/metrics/per_image/pimo.py
```diff
@@ -0,0 +1,956 @@
...
+        anomaly_maps: floating point anomaly score maps of shape (N, H, W)
+        masks: binary (bool or int) ground truth masks of shape (N, H, W)
+
+    Args:
+        num_threshs: number of thresholds to compute (K)
+        binclf_algorithm: algorithm to compute the binary classifier curve (see `binclf_curve_numpy.Algorithm`)
+
+    Returns:
+        PIMOResult: PIMO curves dataclass object. See `PIMOResult` for details.
+    """
```
samet-akcay, line 733: I think examples section here would be really useful. Like what do I need to run AUPIMO with some basic torch input and output

samet-akcay (commented):
@jpcbertoldo, thanks for creating this huuuge and amazing PR! Also, thanks for your patience. Took me a while to go through this.

I've got some comments/questions.

# Clarity vs Conciseness in Naming
- When reading the code, I would prefer clarity over conciseness. I find the abbreviations hard to follow, which overall slows me down reading the code. Overall, I would prefer full words over abbreviations. Something like;
`classification_curve.py` or `binary_classification_curve.py`
`thresholds` instead of `threshs`

# Validation
- I'm wondering how important the validation stuff for this metric evaluation?
  - Is it possible to carry out these checks multiple times? Or just in the correct place.
  - Is it possible to store validators in more organized way. currently there is a `_validation.py`, but there are quite a bit of validators scattered across the sub-package.

# Structure of the sub-package
- I feel we could organise the sub-package a bit more. For example, instead of placing multiple modules horizontally like `pimo.py`, `pimo_numpy.py`, we could potentially create a pimo sub package that organises these, which would be easier to follow. Similarly, it might be an idea to create something like utils/... , even maybe `curves/..`?

- For example, the actual PIMO implementation starts in Line 700ish

I'm working on something similar in PR #2305 to structure sub packages in data, for easier navigation 

```bash
data
├── __init__.py
├── dataclasses
│   ├── __init__.py
│   ├── generic.py
│   ├── numpy
│   │   ├── __init__.py
│   │   ├── base.py
│   │   ├── depth.py
│   │   ├── image.py
│   │   └── video.py
│   └── torch
│       ├── __init__.py
│       ├── base.py
│       ├── depth.py
│       ├── image.py
│       └── video.py
├── datamodules
│   ├── __init__.py
│   ├── base
│   │   ├── __init__.py
│   │   ├── image.py
│   │   └── video.py
│   ├── depth
│   │   ├── __init__.py
│   │   ├── folder_3d.py
│   │   └── mvtec_3d.py
│   ├── image
│   │   ├── __init__.py
│   │   ├── btech.py
│   │   ├── folder.py
│   │   ├── kolektor.py
│   │   ├── mvtec.py
│   │   └── visa.py
│   └── video
│       ├── __init__.py
│       ├── avenue.py
│       ├── shanghaitech.py
│       └── ucsd_ped.py
├── datasets
│   ├── __init__.py
│   ├── base
│   │   ├── __init__.py
│   │   ├── depth.py
│   │   ├── image.py
│   │   └── video.py
│   ├── depth
│   │   ├── __init__.py
│   │   ├── folder_3d.py
│   │   └── mvtec_3d.py
│   ├── image
│   │   ├── __init__.py
│   │   ├── btech.py
│   │   ├── folder.py
│   │   ├── kolektor.py
│   │   ├── mvtec.py
│   │   └── visa.py
│   └── video
│       ├── __init__.py
│       ├── avenue.py
│       ├── shanghaitech.py
│       └── ucsd_ped.py
├── errors.py
├── predict.py
├── transforms
│   ├── __init__.py
│   └── center_crop.py
├── utils
│   ├── __init__.py
│   ├── augmenter.py
│   ├── boxes.py
│   ├── download.py
│   ├── generators
│   │   ├── __init__.py
│   │   └── perlin.py
│   ├── image.py
│   ├── label.py
│   ├── path.py
│   ├── split.py
│   ├── synthetic.py
│   ├── tiler.py
│   └── video.py
└── validators
    ├── __init__.py
    ├── numpy
    │   ├── __init__.py
    │   ├── depth.py
    │   ├── image.py
    │   └── video.py
    ├── path.py
    └── torch
        ├── __init__.py
        ├── depth.py
        ├── image.py
        └── video.py
```

Given your other commitments these days, one possibility would be to work on it together by merging it to a feature branch? Any thoughts? @jpcbertoldo, @ashwinvaidya17, @djdameln ?

ashwinvaidya17 (commented):
I agree with Samet’s suggestion to move all validation checks to a separate module, as it would significantly improve the clarity and organization of `pimo.py`. Furthermore, I propose we also consider removing the NumPy-based computations. Since we aren’t using NumPy for other metrics in Anomalib, and there doesn’t seem to be a current need for NumPy in our metric computations, eliminating it could further simplify the codebase.
As Samet suggested, we can target this change to a feature branch so that the rest of the team can assist in reducing your workload, especially given the significant amount of work you’ve already done.

jpcbertoldo:
> I agree with Samet’s suggestion to move all validation checks to a separate module, as it would significantly improve the clarity and organization of `pimo.py`. Furthermore, I propose we also consider removing the NumPy-based computations. Since we aren’t using NumPy for other metrics in Anomalib, and there doesn’t seem to be a current need for NumPy in our metric computations, eliminating it could further simplify the codebase. As Samet suggested, we can target this change to a feature branch so that the rest of the team can assist in reducing your workload, especially given the significant amount of work you’ve already done.

Ok, no worries. So shoul I remove the `*_numpy.py` stuff and move the code to the torch onlye versions ?

Just one warning. PIMO is fast to compute thanks to 

https://github.com/jpcbertoldo/anomalib/blob/241df0e2aea82373bb69c216e73a7e43106aee22/src/anomalib/metrics/per_image/binclf_curve_numpy.py#L98

which is in numpy, so that specific one would be worth keeping.

Using torchmetrics (even on GPU) is slower. The constraint is that the thresholds are shared across images, but the binary classification (for multiple thresholds) are computed per image.

If this one is removed, then we need a torch-based implementation, which I'm sure is significantly slower for evaluation on the original resolution of images (1000-ish), which was a major argument in the paper. This is a topic we went through last year, but I can try to find the graphs to put numbers in this difference of execution time. 

So, do we keep or remove  this one? (`_binclf_one_curve_python`)

ashwinvaidya17:
Okay, then let's keep `bin_clfcurve_numpy.py`. I wonder how will this affect the computation when we support distributed training. Anyway, we can look into that later. Let's get this merged.

jpcbertoldo:
> Okay, then let's keep `bin_clfcurve_numpy.py`. I wonder how will this affect the computation when we support distributed training. Anyway, we can look into that later. Let's get this merged.

(just guessing) In the torchmetrics class, there isn't much happening in the `update()`, it basically stores score maps and masks. The `compute()` method will do everything. It' pretty much what torchmetrics does with AUROC. So it should be ok?

PR #1794, opened 2024-02-29
Update setup logic in model and datamodule

djdameln:
## 📝 Description

Minor change in how we check if `self._setup` should be called in `setup`. The change is repeated in both the datamodule and the anomaly module. This is mainly to ensure that we can use the modules in a standalone fashion, without/before being attached to a trainer. For example, we could use the following to quickly inspect the images in our datamodule.
```python
>>> from anomalib.data import MVTec
>>>
>>> datamodule = MVTec(image_size=(512, 512))
>>> datamodule.setup()
>>>
>>> sample = next(iter(datamodule.train_dataloader()))
>>> sample["image"].shape
torch.Size([32, 3, 512, 512])
```
Similarly, we could inspect the structure of the torch model of a given Lightning model after invoking `setup`:
```python
>>> from anomalib.models import Ganomaly
>>>
>>> model = Ganomaly()
>>> model.setup()
>>>
>>> model.model.generator.encoder1.input_layers
Sequential(
  (initial-conv-3-64): Conv2d(3, 64, kernel_size=(4, 4), stride=(2, 2), padding=(4, 4), bias=False)
  (initial-relu-64): LeakyReLU(negative_slope=0.2, inplace=True)
)
```

Without the current change, this would lead to problems downstream when the entrypoints of the engine are called.

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 🔒 Security update

## ✅ Checklist

For more information about code review checklists, see the [Code Review Checklist](../docs/source/markdown/guides/developer/code_review_checklist.md).

### DIFF
--- a/src/anomalib/data/base/datamodule.py
+++ b/src/anomalib/data/base/datamodule.py
@@ -120,7 +120,8 @@ class AnomalibDataModule(LightningDataModule, ABC):
             stage: str | None:  Train/Val/Test stages.
                 Defaults to ``None``.
         """
-        if not any(hasattr(self, subset) for subset in ["train_data", "val_data", "test_data"]) or not self._is_setup:
+        has_subset = any(hasattr(self, subset) for subset in ["train_data", "val_data", "test_data"])
+        if not has_subset or not self._is_setup:
             self._setup(stage)
             self._create_test_split()
             self._create_val_split()
--- a/src/anomalib/data/base/datamodule.py
+++ b/src/anomalib/data/base/datamodule.py
@@ -9,6 +9,7 @@ from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any
 
 from lightning.pytorch import LightningDataModule
+from lightning.pytorch.trainer.states import TrainerFn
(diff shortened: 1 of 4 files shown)


src/anomalib/data/base/datamodule.py
```diff
@@ -110,18 +111,22 @@ def __init__(
...
+
     def setup(self, stage: str | None = None) -> None:
         """Set up train, validation and test data.
 
         Args:
             stage: str | None:  Train/Val/Test stages.
                 Defaults to ``None``.
         """
-        if not self.is_setup:
+        if not any(hasattr(self, subset) for subset in ["train_data", "val_data", "test_data"]) or not self._is_setup:
```
samet-akcay, line 123: would it be possible to split this condition with an easier-to-read variable names?

samet-akcay (changes requested):
Thanks, only a single comment...

src/anomalib/data/base/datamodule.py
djdameln: done

samet-akcay (approved):
Great, thanks!

PR #2060, opened 2024-05-15
🚀 Anomalib Pipelines

ashwinvaidya17:
# Description

- Experimental pipelines


src/anomalib/pipelines/benchmark/job.py
```diff
@@ -0,0 +1,108 @@
...
+        for result in results:
+            for key, value in result.items():
+                output[key].append(value)
+        return pd.DataFrame(output)
+
+    @staticmethod
+    def save(result: pd.DataFrame) -> None:
+        """Save the result to a csv file."""
+        BenchmarkJob._print_tabular_results(result)
+        file_path = Path("runs") / BenchmarkJob.name / datetime.now().strftime("%Y-%m-%d-%H:%M:%S") / "results.csv"
```
blaz-r, line 92: I think the format should be changed to `%Y-%m-%d-%H_%M_%S` or something that doesn't use colons, as those are invalid on windows.

ashwinvaidya17, line 92:
```suggestion
        file_path = Path("runs") / BenchmarkJob.name / datetime.now().strftime("%Y-%m-%d-%H_%M_%S") / "results.csv"
```

samet-akcay (approved):

PR #2467, opened 2024-12-12
Apply transforms in PreProcessor

djdameln:
## 📝 Description

- Delegate the responsibility of applying the input transforms to the `PreProcessor`. This is more in line with the other auxiliary components such as post-processing and evaluation, where the full functionality of the aux operation is contained in a single class. 
- To account for datasets with heterogeneous image shapes, we need to perform an additional check in the collate method, and resize the images before collating if not all images have the same shape.
- Thorough manual testing may be required to check for any unwanted side-effects.

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 🔒 Security update

## ✅ Checklist


tests/unit/pre_processing/test_pre_processing.py
```diff
@@ -91,6 +91,7 @@ def test_different_stage_transforms() -> None:
         assert isinstance(processed_batch, torch.Tensor)
         assert processed_batch.shape == (1, 3, 288, 288)
 
+    @pytest.skip
```
samet-akcay, line 94: why are these skipped?

src/anomalib/pre_processing/pre_processing.py
```diff
@@ -109,63 +99,49 @@ def __init__(
...
-    def setup(self, trainer: Trainer, pl_module: LightningModule, stage: str) -> None:
-        """Configure transforms at the start of each stage.
-
-        Args:
-            trainer: The Lightning trainer.
-            pl_module: The Lightning module.
-            stage: The stage (e.g., 'fit', 'validate', 'test', 'predict').
-        """
-        stage = TrainerFn(stage).value  # Ensure stage is str
+    def on_train_batch_start(self, trainer: Trainer, pl_module: LightningModule, batch: Batch, batch_idx: int) -> None:
```
samet-akcay, line 102: more like a styling issue.. it might be an idea to use multi-lines here as well, similar to other hooks.

src/anomalib/data/dataclasses/generic.py
```diff
@@ -656,5 +658,20 @@ def batch_size(self) -> int:
...
+        if torch.unique(im_shapes, dim=0).size(0) != 1:  # check if batch has heterogeneous shapes
+            target_shape = im_shapes[
+                torch.unravel_index(im_shapes.argmax(), im_shapes.shape)[0],
+                :,
+            ]  # shape of image with largest H or W
+            for item in items:
+                for key in keys:
+                    value = getattr(item, key)
+                    if isinstance(value, Image | Mask):
+                        setattr(item, key, resize(value, target_shape))
```
alexriedel1, line 673: the image transforms are supposed to be applied before the batch collate function right? So this resize operation will only apply if no resizing is defined in the image transforms?

djdameln: With the changes in this PR, the transforms are actually applied by the new `PreProcessor` class using its `on_before_<stage>_batch` hooks, so after collating. This has the following advantages:
- The transforms are contained within the `PreProcessor`. Previously, we needed complicated logic to ensure that the transforms in the datamodule/dataloaders were equal to the transforms stored in (or defined by) the model. Now, the Preprocessor is the only class concerned with the transforms, leading to a much better separation of responsibilities.
- The `PreProcessor` is a submodule of the Lightning model, so we have direct access to the transforms from within the model. This helps with model saving/loading (transforms are automatically stored in the model's state dict), and exporting (transforms are included in the forward pass so that they end up in the model graph).

The only reason why we need resizing in the collate method is to allow collating image batches with heterogeneous image sizes. We can only stack a sequence of image tensors into a single tensor if all images have the same shape. This is why we check the image shapes in the batch, and if they are not equal, we resize all of them to the shape of the image with the biggest width or height. This is only to ensure that we can succesfully collate the images into a single batch, and is completely independent of any further resizing defined in the transforms. (In general, it is recommended to apply resizing in the image transforms, and most of our models include this in their default set transforms).

tests/unit/pre_processing/test_pre_processing.py
djdameln: They are not needed anymore. These tests check the setting of the transforms in the datamodule and dataloader from the preprocessor, which we don't do anymore. I skipped them for now, but once we settle on the final design they could be removed.

samet-akcay: ah right, I see

src/anomalib/data/dataclasses/generic.py
alexriedel1: Yes I see! Resizing can be a critical function for image processing models when using different interpolations modes.
What will happen if a user uses a different interpolation mode for their custom transform function:
- the batches with different shaped images are resized in the collate function via the (default) bilinear transform, and later resized using the users custom interpolation mode
- the batches with equal shaped images are only resized using the users custom interpolation mode

There's a good chance that this won't affect anything in anomalib but if accuracy of models drop this is also some place to look after..

djdameln: We hadn't considered that, thanks for pointing it out!

We could potentially add an optional resize argument to the `AnomalibDatamodule` to give the user control over the resizing before collating. But this would require some further investigation and internal discussion.

alexriedel1: Ok! Yes maybe its not even a problem for the model accuracy. Still it's a bit computational expensive to have one extra resizing operation..

alexriedel1:
Yes I also thought about the problem and what you now did, adding augmentations to the dataset, is the onyl solution to this.. 
I also would say it's common and good practice in other frameworks.

djdameln:
> Yes I also thought about the problem and what you now did, adding augmentations to the dataset, is the onyl solution to this.. I also would say it's common and good practice in other frameworks.

@alexriedel1 As you've seen, I re-introduced transforms on the dataset/datamodule side, but now named as `augmentations`. 

**Motivation:**

In addition to the transforms applied by the `PreProcessor`, we now also have augmentation transforms on the dataset/datamodule side. The two serve different purposes: Augmentations are optional and can be used for enriching the dataset, e.g. random transforms on training set or test-time augmentations on the test set. `PreProcessor` transforms are model-specific and are used to ensure that the input images are read correctly by the model. Usually this involves resizing and normalization, but other transforms are possible as well. 

This way, we keep the model-specific transforms on the model side, which has the advantages mentioned in [my earlier comment](https://github.com/openvinotoolkit/anomalib/pull/2467#discussion_r1883750755). At the same time, the augmentation transforms on the datamodule side make it easier for the user to define their custom augmentation pipeline. The main advantage over earlier designs is that the user now only has to define the augmentations, and does not need to include the model-specific transforms such as resizing and normalization (because these are handled separately by the `PreProcessor`).

**Concerning resizing:** 
- When no augmentations are supplied, resizing only happens on the `PreProcessor` side, using the model-specific transforms defined for the model, or any custom resize/transform passed to the `PreProcessor` by the user. When the user wants to change the image size, **as seen by the model**, this is the resize transform that they need to change.
- Applying augmentations to the images at full resolution may be costly, so the user may want to include a `Resize` transform in their augmentation pipeline to speed up the data loading. In this case, resizing will happen twice. Once when the dataset applies the augmentations before collating, and once by the `PreProcessor` after collating. This may be computationally expensive as you mentioned earlier, but only when the two resize transforms use a different output size. When the resize transforms use the same size, the second one will be skipped.
- In the case of heterogeneous image size in the dataset, the user can now just pass a `Resize` transform to the `augmentations` argument of the datamodule (or dataset). This will ensure that the dataset resizes the images before collating, and the additional resize operation in the collate method will not be called. (We still keep the resize operation in the collate method as an additional safeguard). This gives the user control over the interpolation method that is used for resizing the images.

Please let me know what you think, any suggestions are more than welcome :)

alexriedel1:
> > Yes I also thought about the problem and what you now did, adding augmentations to the dataset, is the onyl solution to this.. I also would say it's common and good practice in other frameworks.
> 
> @alexriedel1 As you've seen, I re-introduced transforms on the dataset/datamodule side, but now named as `augmentations`.
> 
> **Motivation:**
> 
> In addition to the transforms applied by the `PreProcessor`, we now also have augmentation transforms on the dataset/datamodule side. The two serve different purposes: Augmentations are optional and can be used for enriching the dataset, e.g. random transforms on training set or test-time augmentations on the test set. `PreProcessor` transforms are model-specific and are used to ensure that the input images are read correctly by the model. Usually this involves resizing and normalization, but other transforms are possible as well.
> 
> This way, we keep the model-specific transforms on the model side, which has the advantages mentioned in [my earlier comment](https://github.com/openvinotoolkit/anomalib/pull/2467#discussion_r1883750755). At the same time, the augmentation transforms on the datamodule side make it easier for the user to define their custom augmentation pipeline. The main advantage over earlier designs is that the user now only has to define the augmentations, and does not need to include the model-specific transforms such as resizing and normalization (because these are handled separately by the `PreProcessor`).
> 
> **Concerning resizing:**
> 
> * When no augmentations are supplied, resizing only happens on the `PreProcessor` side, using the model-specific transforms defined for the model, or any custom resize/transform passed to the `PreProcessor` by the user. When the user wants to change the image size, **as seen by the model**, this is the resize transform that they need to change.
> * Applying augmentations to the images at full resolution may be costly, so the user may want to include a `Resize` transform in their augmentation pipeline to speed up the data loading. In this case, resizing will happen twice. Once when the dataset applies the augmentations before collating, and once by the `PreProcessor` after collating. This may be computationally expensive as you mentioned earlier, but only when the two resize transforms use a different output size. When the resize transforms use the same size, the second one will be skipped.
> * In the case of heterogeneous image size in the dataset, the user can now just pass a `Resize` transform to the `augmentations` argument of the datamodule (or dataset). This will ensure that the dataset resizes the images before collating, and the additional resize operation in the collate method will not be called. (We still keep the resize operation in the collate method as an additional safeguard). This gives the user control over the interpolation method that is used for resizing the images.
> 
> Please let me know what you think, any suggestions are more than welcome :)

Yes I'm (and most users too I guess) am happy with this! You should make clear ion the docs that preprocessing transforms are stored in the model after export while augmentations are not, so people know exactly which one to use for their use case.

src/anomalib/data/datamodules/base/image.py
```diff
@@ -81,8 +93,12 @@ def __init__(
         train_batch_size: int,
         eval_batch_size: int,
         num_workers: int,
-        val_split_mode: ValSplitMode | str,
-        val_split_ratio: float,
+        train_augmentations: Transform | None = None,
+        val_augmentations: Transform | None = None,
+        test_augmentations: Transform | None = None,
+        augmentations: Transform | None = None,
```
samet-akcay, line 99: Looking at the naming, are we 100% clear that `augmentations` is the best naming in this case? The reason for asking is the annotation for `augmentations` is, obviously, `Transform`. 

Would it make absolute sense to users?

alexriedel1: Ok so there are transforms that can be pre-defined model specific and they are stored in the exported model. 
And there are augmentations that are only applied during training.
Both can be defined separately for training, testing and validation. 


Maybe it would make sense to call the transforms something like `input_image_transforms` ? Or something that makes clearer that it is "baked" into the model and the last thing that transforms images before being processed by the model..

djdameln: On the datamodule side, I think `augmentations` is a more accurate and descriptive name than `transforms`. The `augmentations` naming signals that these are meant to define the data augmentation pipeline. `transforms` is more ambiguous and can refer to both data augmentations and model-specific input image pre-processing.

On the model/pre-processor side, we could think of a more descriptive name than `transform`, maybe`input_transforms`?  In any case we should ensure that there is a thorough how-to guide that clearly explains these concepts.

---

> Both can be defined separately for training, testing and validation.

@alexriedel1  Please note that in the latest version of this PR, we only keep 1 transform on the `PreProcessor` side. Since the model always requires the inputs in the same format, we figured there was no longer a use-case for having separate model transforms between the different subsets (but please let us know if you disagree).

alexriedel1: > @alexriedel1 Please note that in the latest version of this PR, we only keep 1 transform on the `PreProcessor` side. Since the model always requires the inputs in the same format, we figured there was no longer a use-case for having separate model transforms between the different subsets (but please let us know if you disagree).

Yes thats a good idea!

ashwinvaidya17 (approved):
Thanks for the massive efforts here. Looks good to me.

src/anomalib/utils/attrs.py
```diff
@@ -0,0 +1,24 @@
+"""Utility functions for working with attributes."""
+
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any
+
+
+def get_nested_attr(obj: Any, attr_path: str, default: Any | None = None) -> Any:  # noqa: ANN401
```
samet-akcay, line 9: Can you elaborate why this is needed? Maybe add more description to the docstring?

src/anomalib/data/transforms/utils.py
```diff
@@ -0,0 +1,23 @@
+"""Utility functions for working with Torchvision Transforms."""
+
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+from torchvision.transforms.v2 import Compose, Transform
+
+
+def get_transforms_of_type(input_transform: Transform | None, transform_type: type[Transform]) -> list[type[Transform]]:
```
samet-akcay, line 9:
```suggestion
def get_transforms_by_type(input_transform: Transform | None, transform_type: type[Transform]) -> list[type[Transform]]:
```

or `filter_transforms_by_type`?

djdameln: For `filter_transforms_by_type` I would expect the input to be a sequence of transform objects.

This is not entirely accurate, since the function extracts the transforms from a single other transform, which could be an actual transform or a Compose.

Maybe `extract_transforms_by_type`?

samet-akcay (approved):

PR #2508, opened 2025-01-14
Add non-strict mode to AnomalibMetric

djdameln:
## 📝 Description

This PR adds a non-strict mode to `AnomalibMetric` class, which reduces boilerplate code in applications where the batches passed to `update` may or may not contain the required fields.

- Add `strict` keyword argument to `AnomalibMetric` class that defines the behaviour when the batch passed to update has missing fields.
- When `strict==True` (default), the behaviour will be the same as it was previously. The metric will raise an error if the field is not present in the passed instance, or if the value of the field is None.
- When `strict==False`, the metric will skip the update call of the internal metric, but not raise an error. When `compute` is called, the metric will return `None` if all calls to update were unsuccessful.
- The changes in `OneClassPostProcessor` are an example how the non-strict mode can reduce boilerplate code. Setting the evaluator's pixel metrics to non-strict also resolves issues like [this one](https://github.com/openvinotoolkit/anomalib/issues/2510).

Other changes:

- Add unit tests for `AnomalibMetric`.
- Add integration tests that test the behaviour of aux components across different task types. Although we got rid of the task type parameter, we still have implicit task types in the datamodule and model, depending on which gt data is available and what type of predictions are returned. In the past we often encountered bugs due to mismatch between expected and available gt and prediction types. These new integration tests should catch this type of bug in the future.
- Convert `MinMax` and `F1AdaptiveThreshold` to `AnomalibMetric`, and use in `OneClassPostProcessor`
- Remove persistent states in `MinMax` metric. This is no longer needed in v2.

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 🔒 Security update

## ✅ Checklist


ashwinvaidya17 (approved):
Thanks

src/anomalib/metrics/base.py
```diff
@@ -40,13 +40,34 @@
...
+        >>> batch = ImageBatch(
+        ...     image=torch.rand(4, 3, 256, 256),
+        ...     gt_label=torch.tensor([0, 0, 1, 1])
+        ... )
+        >>>
+        >>> f1_score_strict.update(batch)  # ValueError
+        >>> f1_score_strict.compute()  # UserWarning, tensor(0.)
+        >>>
+        >>> f1_score_nonstrict.update(batch)  # No error
+        >>> f1_score_nonstrict.compute()  # None
```
samet-akcay, line 62: `# Copyright (C) 2024-2025 Intel Corporation` :)

src/anomalib/metrics/base.py
```diff
@@ -109,6 +132,8 @@ def __init__(
             raise ValueError(msg)
         self.fields = fields
         self.name = prefix + self.__class__.__name__
+        self.strict = strict
+        self.__update_count = 0  # keeps track of the update calls of the wrapper class
```
samet-akcay, line 136: is double underscore `__` intentional?

src/anomalib/metrics/base.py
```diff
@@ -130,13 +155,43 @@ def update(self, batch: Batch, *args, **kwargs) -> None:
...
+        """Compute the metric value.
+
+        If the metric has not been updated, and metric is not in strict mode, return None.
+
+        Returns:
+            torch.Tensor: Computed metric value or None.
+        """
+        if self._update_count == 0 and not self.strict:  # type: ignore[attr-defined]
+            return None
+        return super().compute()  # type: ignore[misc]
```
samet-akcay, line 193: Can you elaborate the types are ignored? `compute` always expected to return tensor?

samet-akcay (commented):
thanks, just some minor comments..

src/anomalib/metrics/base.py
djdameln: yes, but it's no longer needed, so I'll remove it

src/anomalib/metrics/base.py
djdameln: Without the ignore, mypy raises the following error:
```
src/anomalib/metrics/base.py:191: error: "compute" undefined in superclass  [misc]
```
Mypy does not recognize the parent class, because this class can be added to any Metric class, sort of like a mixin.

PR #2548, opened 2025-02-04
Fix latest torch and torchvision compatibility

djdameln:
## 📝 Description

This PR fixes two problems that emerge when upgrading to PyTorch 2.6 and Torchvision 0.21.

- Fixes a problem with model loading in the torch inferencer, leading to CI failures. The problem was caused by a breaking change in PyTorch 2.6, where the default value of `weights_only` param of `torch.load` was changed from `False` to `True`. See https://pytorch.org/docs/main/notes/serialization.html#torch-load-with-weights-only-true for more information.
- Note that we may need to revisit how we load models in the torch inferencer, because loading with `weights_only==False` may lead to arbitrary code execution.
- Fixes a problem caused by the internal renaming of `Transform._transform` to `Transform.transform` in Torchvision 0.21.

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 🔒 Security update

## ✅ Checklist

### DIFF
--- a/src/anomalib/data/transforms/center_crop.py
+++ b/src/anomalib/data/transforms/center_crop.py
@@ -142,3 +142,17 @@ class ExportableCenterCrop(Transform):
         """
         del params
         return center_crop_image(inpt, output_size=self.size)
+
+    def transform(self, inpt: torch.Tensor, params: dict[str, Any]) -> torch.Tensor:
+        """Wrapper for self._transform.
+
+        This is to ensure compatibility with Torchvision 2.6+, where the `_transform` method was renamed to `transform`.
+
+        Args:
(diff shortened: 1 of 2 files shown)


ashwinvaidya17:
Is this backward compatible? Do you think we need to set the minimum pytorch version?

djdameln:
> Is this backward compatible? Do you think we need to set the minimum pytorch version?

This change does not affect backward compatibility. In 2.5 and earlier, the `weights_only` parameter was already there, just with a different default value. Explicitly setting the parameter to `False` on our side ensures correct loading in both 2.6 and <2.5.

I see that there is another problem caused by a breaking change on the torchvision side. I'll add a fix for that as well.

samet-akcay (approved):

PR #2562, opened 2025-02-12
🚀 Add `MVTecAD2` dataset

samet-akcay:
## 📝 Description

This PR adds support for the MVTec AD 2 dataset, which is an advanced version of the original MVTec AD dataset designed for unsupervised anomaly detection. The implementation includes support for three different test sets:

- Public test set (`test_public/`): Contains both normal and anomalous samples with ground truth masks
- Private test set (`test_private/`): Contains unseen test samples without ground truth
- Private mixed test set (`test_private_mixed/`): Contains unseen test samples with mixed anomalies without ground truth

## Features
- Full implementation of the MVTecAD2 dataset class with support for all test types
- Proper handling of dataset splits (train, validation, test)
- Support for image augmentations and transforms
- Comprehensive test coverage for all test types and edge cases
- Integration with existing Anomalib data pipeline

## ✨ Changes

- Added `MVTecAD2` datamodule in `src/anomalib/data/datamodules/image/mvtecad2.py`
- Added `MVTecAD2Dataset` class in `src/anomalib/data/datasets/image/mvtecad2.py`
- Added `TestType` enum to handle different test set types
- Added unit tests for the dataset implementation
- Updated dummy dataset generation for testing
- Added configuration example in `examples/configs/data/mvtecad2.yaml`

- [ ] 📚 Documentation update
- [ ] 🔒 Security update

## Testing
The implementation includes comprehensive tests that verify:
- Dataset loading and preprocessing
- All three test types (public, private, mixed)
- Image augmentations and transforms
- Error handling for invalid test types
- Dataset split functionality
- Integration with the Anomalib pipeline

## Usage Example
```python
from anomalib.data import MVTecAD2

# Create datamodule with public test set
datamodule = MVTecAD2(
    root="./datasets/MVTec_AD_2",
    category="sheet_metal",
    train_batch_size=32,
    eval_batch_size=32,
)

# Use private test set
datamodule = MVTecAD2(
    root="./datasets/MVTec_AD_2",
    category="sheet_metal",
    test_type="private",
)

# Access different test sets
datamodule.setup()
public_loader = datamodule.test_dataloader()  # returns loader based on test_type
private_loader = datamodule.test_dataloader(test_type="private")
mixed_loader = datamodule.test_dataloader(test_type="private_mixed")
```

## License
The MVTec AD 2 dataset will be released under the Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International License (CC BY-NC-SA 4.0).

## Notes
- This implementation requires proper dataset access and follows the MVTec AD 2 dataset license terms
- Users need to download the dataset separately from the MVTec website
- The implementation maintains compatibility with the existing Anomalib API


ashwinvaidya17 (approved):

src/anomalib/data/datasets/base/image.py
```diff
@@ -273,23 +270,24 @@ def __getitem__(self, index: int) -> DatasetItem:
...
-            # Only Anomalous (1) images have masks in anomaly datasets
-            # Therefore, create empty mask for Normal (0) images.
+            # Create empty mask for:
+            # - Normal samples (label_index = 0)
+            # - Unknown samples (label_index = -1)
+            # Only use mask for anomalous samples (label_index = 1)
             mask = (
                 Mask(torch.zeros(image.shape[-2:])).to(torch.uint8)
-                if label_index == LabelName.NORMAL
+                if label_index in {LabelName.NORMAL, LabelName.UNKNOWN}
```
djdameln, line 279: Does it make sense to provide an empty mask as gt_mask in the case of unknown label type? I'm concerned it could lead to some unexpected behaviour when users try to evaluate on the private test set.

Maybe it would be better to provide no mask at all for the unknown label? This would prevent running an evaluation sequence on the private test set. (Not sure if this would lead to any practical issues at runtime).

src/anomalib/data/datasets/base/image.py
```diff
@@ -273,23 +270,24 @@ def __getitem__(self, index: int) -> DatasetItem:
...
-
         else:
             msg = f"Unknown task type: {self.task}"
             raise ValueError(msg)
 
         return ImageItem(
             image=item["image"],
             gt_mask=item.get("gt_mask"),
-            gt_label=int(label_index),
+            gt_label=torch.tensor(label_index),  # Convert to tensor to match type hints
```
djdameln, line 290: In case of `UNKNOWN` label, the `-1` label_index value will get converted to a positive (`True`) boolean by the dataclass validator. When combined with the all-zero gt_mask that I commented on above, this leads to the weird situation where an `UNKNOWN` test sample will have a positive (anomalous) GT class label, but an empty (all-normal) GT mask. It may be better to have consistency between GT-mask and -label.

djdameln (commented):
Thanks for the effort. I have some comments related to the new `UNKNOWN` label introduced in this PR. My feeling is that it may be better to leave the label and mask empty for samples of the private test set, to prevent users from running an evaluation sequence and getting meaningless metric values as a result. Maybe we could warn the user in case they try to run evaluation on a test set that contains `UNKNOWN` labels.

djdameln (approved):

PR #2629, opened 2025-03-26
🚀 feat(model): add GLASS model into Anomalib

### LINKED ISSUE #2619: 📋 [TASK]  Implement GLASS - ECCV 2024
### User Story

I want to implement a new anomaly detection model - GLASS based on [this paper](https://arxiv.org/abs/2407.09359). This model combines the use global and local anomaly synthesis to improve accuracy.

### Acceptance Criteria

1. The new algorithm is implemented and integrated into the anomalib framework
2. Unit tests are written and pass for the new implementation
3. Performance benchmarks show improvement over existing methods on specified datasets
4. Documentation is updated to include usage instructions and theory behind the new algorithm
5. An example notebook is provided demonstrating the algorithm's application

### Story Points

2

### Related Epic

_No response_

### Category

Anomaly Detection Algorithms

### Additional Context

_No response_

code-dev05:
## 📝 Description

- This PR will add the GLASS model for synthesis of anomalies on a global and local level.
- 🛠️ Fixes #2619 

## ✨ Changes

- [x] 📚 Documentation update

## TODO
- [x] Implement the model from this [research paper](https://arxiv.org/pdf/2407.09359).
- [x] Write a trainer for the above model.
- [x] Add comments to the code for easier understanding.
- [ ] Write test cases for verification.
- [ ] Update documentation for the above model.
- [ ] Any optimizations if needed?

I will keep updating the PR as I implement the rest of the list.

### DIFF
--- a/src/anomalib/models/image/glass/components/rescale_segmentor.py
+++ b/src/anomalib/models/image/glass/components/rescale_segmentor.py
@@ -17,7 +17,7 @@ class RescaleSegmentor:
         smoothing (int): The standard deviation used for Gaussian smoothing.
     """
 
-    def __init__(self, target_size: int = 288) -> None:
+    def __init__(self, target_size: tuple[int, int] = (288, 288)) -> None:
         """Initializes the RescaleSegmentor.
 
         Args:
@@ -48,7 +48,10 @@ class RescaleSegmentor:
             scores = patch_scores.to(device)
             scores = scores.unsqueeze(1)  # [N, 1, H, W]
             scores = f.interpolate(
-                scores, size=self.target_size, mode="bilinear", align_corners=False,
+                scores,
+                size=self.target_size,
+                mode="bilinear",
+                align_corners=False,
             )
             patch_scores = scores.squeeze(1)  # [N, H, W]
 
--- a/src/anomalib/models/image/glass/lightning_model.py
+++ b/src/anomalib/models/image/glass/lightning_model.py
@@ -135,6 +135,11 @@ class Glass(AnomalibModule):
         if layers is None:
             layers = ["layer2", "layer3"]
 
+        if anomaly_source_path is not None:
+            dtd_dir = Path(anomaly_source_path)
+            if not dtd_dir.is_dir():
+                download_and_extract(dtd_dir, DTD_DOWNLOAD_INFO)
+
         self.model = GlassModel(
             input_shape=input_shape,
             anomaly_source_path=anomaly_source_path,
@@ -154,6 +159,7 @@ class Glass(AnomalibModule):
         )
 
         self.learning_rate = learning_rate
+        self.pre_trained = pre_trained
 
         if pre_projection > 0:
             self.projection_opt = optim.Adam(
@@ -164,7 +170,7 @@ class Glass(AnomalibModule):
         else:
             self.projection_opt = None
 
-        if not pre_trained:
+        if not self.pre_trained:
             self.backbone_opt = optim.AdamW(
                 self.model.forward_modules["feature_aggregator"].backbone.parameters(),
                 self.learning_rate,
@@ -173,11 +179,6 @@ class Glass(AnomalibModule):
             self.backbone_opt = None
 
         self.automatic_optimization = False
-
-        if anomaly_source_path is not None:
-            dtd_dir = Path(anomaly_source_path)
-            if not dtd_dir.is_dir():
-                download_and_extract(dtd_dir, DTD_DOWNLOAD_INFO)
 
     @classmethod
     def configure_pre_processor(
@@ -247,9 +248,11 @@ class Glass(AnomalibModule):
             STEP_OUTPUT: Dictionary containing loss values and metrics
         """
         del batch_idx
+
         discriminator_opt = self.optimizers()
 
-        self.model.forward_modules.eval()
+        if not self.pre_trained:
+            self.model.forward_modules["feature_aggregator"].train()
         if self.model.pre_projection > 0:
             self.model.projection.train()
         self.model.discriminator.train()
@@ -295,7 +298,7 @@ class Glass(AnomalibModule):
         predictions = self.model(batch.image)
         return batch.update(**predictions._asdict())
 
-    def on_train_start(self) -> None:
+    def on_train_epoch_start(self) -> None:
         """Initialize model by computing mean feature representation across training dataset.
 
         This method is called at the start of training and computes a mean feature vector
--- a/src/anomalib/models/image/glass/loss.py
+++ b/src/anomalib/models/image/glass/loss.py
@@ -75,7 +75,7 @@ class FocalLoss(nn.Module):
     def __init__(
         self,
         apply_nonlinearity: nn.Module | None = None,
-        alpha: float | list | np.ndarray | None = None,
+        alpha: float | torch.Tensor | np.ndarray | None = None,
         gamma: float = 2,
         balance_index: int = 0,
         smooth: float = 1e-5,
@@ -129,13 +129,12 @@ class FocalLoss(nn.Module):
         target = target.view(-1, 1)
 
         alpha = self.alpha
-        if alpha is None:
+        if self.alpha is None:
             alpha = torch.ones(num_classes, 1)
-        elif isinstance(alpha, (list | np.ndarray)):
-            assert len(alpha) == num_classes
+        elif isinstance(self.alpha, (list | np.ndarray)):
             alpha = torch.FloatTensor(alpha).view(num_classes, 1)
             alpha = alpha / alpha.sum()
-        elif isinstance(alpha, float):
+        elif isinstance(self.alpha, float):
             alpha = torch.ones(num_classes, 1)
             alpha = alpha * (1 - self.alpha)
             alpha[self.balance_index] = self.alpha
--- a/src/anomalib/models/image/glass/torch_model.py
+++ b/src/anomalib/models/image/glass/torch_model.py
@@ -128,7 +128,7 @@ class GlassModel(nn.Module):
 
         self.patch_maker = PatchMaker(patchsize, stride=patchstride)
 
-        self.anomaly_segmentor = RescaleSegmentor(target_size=input_shape[:])
+        self.anomaly_segmentor = RescaleSegmentor(target_size=input_shape)
 
     def calculate_center(self, dataloader: dataloader, device: torch.device) -> None:
         """Calculates and updates the center embedding from a dataset.
@@ -164,7 +164,8 @@ class GlassModel(nn.Module):
                 else:
                     self.center += torch.mean(outputs, dim=0)
 
-    def calculate_features(self,
+    def calculate_features(
+        self,
         img: torch.Tensor,
         aug: torch.Tensor,
         evaluation: bool = False,
@@ -193,8 +194,10 @@ class GlassModel(nn.Module):
             true_feats = true_feats[0] if len(true_feats) == 2 else true_feats
         else:
             fake_feats = self.generate_embeddings(aug, evaluation=evaluation)[0]
+            assert isinstance(fake_feats, torch.Tensor)
             fake_feats.requires_grad = True
             true_feats = self.generate_embeddings(img, evaluation=evaluation)[0]
+            assert isinstance(true_feats, torch.Tensor)
             true_feats.requires_grad = True
 
         return true_feats, fake_feats
@@ -227,7 +230,7 @@ class GlassModel(nn.Module):
         """
         if not evaluation and not self.pre_trained:
             self.forward_modules["feature_aggregator"].train()
-            features = self.forward_modules["feature_aggregator"](images)
+            features = self.forward_modules["feature_aggregator"](images, eval=evaluation)
         else:
             self.forward_modules["feature_aggregator"].eval()
             with torch.no_grad():
@@ -315,7 +318,7 @@ class GlassModel(nn.Module):
     def forward(
         self,
         img: torch.Tensor,
-    ) -> tuple[torch.Tensor, torch.Tensor]:
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor] | InferenceBatch:
         """Forward pass to compute patch-wise feature embeddings for original and augmented images.
 
         Depending on whether a pre-projection module is used, this method optionally applies it to the


samet-akcay:
@code-dev05, you could use Anomalib's [PreProcessor](https://anomalib.readthedocs.io/en/v2.0.0/markdown/guides/how_to/models/pre_processor.html) class. Anomalib also provides [PerlinAnomalyGenerator](PerlinAnomalyGenerator) class that you could create as a transform and pass into `PreProcessor`

Let me know if the documentation is not clear, or you would need more clarification.

samet-akcay:
Alternatively, you could use [perlin_noise_generator](https://anomalib.readthedocs.io/en/v2.0.0/markdown/guides/reference/data/utils/synthetic.html#anomalib.data.utils.generators.generate_perlin_noise) function in `on_train_batch_start` method in your `Glass` implementation that inherits `AnomalibModule`

What this would do is to apply the perlin noise when the model receives the input on batch start.

You could prototype any one of them. We could always polish it later on once your pr is ready for review

code-dev05:
@samet-akcay I have created one new commit. Please check it. I have noticed a few issues, 
- One is computing the center of the dataset. Since we are using the Batch dataclass, in one training step we only compute the outputs for one batch, and for the center we need the whole dataset. One solution for this would be to somehow calculate the center for the whole dataset during initialisation and other way would be to calculate the center of the batch for that training step. This approach would be more inaccurate than the other.
- Second is the the use of two different datasets. In the original code, they have used DTD dataset along with MVTec for different textures. How to implement that? In my code I have used batch.image and batch.aug so I am assuming we are storing both the image and the texture in the object. I think we have to make a new dataclass for this? If so, we can directly integrate the perlin noise in the image instead of during training step.
Thank you for answering my queries.

code-dev05:
@samet-akcay I will also need to create a datamodule for the [DTD dataset](https://www.robots.ox.ac.uk/~vgg/data/dtd/). Should I add that in this PR only or create a new PR?

src/anomalib/models/image/glass/__init__.py
```diff
@@ -0,0 +1 @@
+from .lightning_model import Glass
```
ashwinvaidya17, line 1: Can you install `pre-comit` hooks and run it again? It will sort out such issues

src/anomalib/models/image/glass/backbones.py
```diff
@@ -0,0 +1,50 @@
+import torchvision.models as models
```
ashwinvaidya17, line 1: It would be nice to attribute files like these to the original implementation. Also, include an entry in `third-party.txt`
https://github.com/open-edge-platform/anomalib/blob/main/third-party-programs.txt

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,161 @@
...
+from .loss import FocalLoss
+from .torch_model import GlassModel
+
+class Glass(AnomalibModule):
+    def __init__(
+            self,
+            backbone,
+            input_shape,
+            pretrain_embed_dim,
+            target_embed_dim,
```
ashwinvaidya17, line 24: Can you type these as well? Same for the other files.

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,161 @@
+import torch
```
ashwinvaidya17, line 1: Can you add license and docstring to the top of all the files?

ashwinvaidya17 (changes requested):

src/anomalib/models/image/glass/lightning_model.py
djdameln: Also please provide default values. Users should be able to instantiate the model without passing any args.

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,161 @@
...
+from anomalib.models.components import AnomalibModule
+from anomalib.metrics import Evaluator
+from anomalib.post_processing import PostProcessor
+from anomalib.pre_processing import PreProcessor
+from anomalib.visualization import Visualizer
+
+from .loss import FocalLoss
+from .torch_model import GlassModel
+
+class Glass(AnomalibModule):
```
djdameln, line 18: The class is missing the following mandatory properties: `learning_type` and `trainer_arguments`. Without the implementation of these properties, it won't run. You can have a look at the other models for example implementations.

djdameln (changes requested):

djdameln:
Hi @code-dev05

Thanks a lot for your effort! Here are some comments that may help you address your issues related to center computation and noise generation:

> - One is computing the center of the dataset. Since we are using the Batch dataclass, in one training step we only compute the outputs for one batch, and for the center we need the whole dataset. One solution for this would be to somehow calculate the center for the whole dataset during initialisation and other way would be to calculate the center of the batch for that training step. This approach would be more inaccurate than the other.

We want to reproduce the algorithm from the original paper as closely as possible, so I would suggest to compute the center of the dataset in an initialization step. You could try using the `setup` or `on_fit_start`/`on_train_start` lightning hooks for this (see [lightning docs](https://lightning.ai/docs/pytorch/latest/common/lightning_module.html#hooks) for further reading on model hooks).

> - Second is the the use of two different datasets. In the original code, they have used DTD dataset along with MVTec for different textures. How to implement that? In my code I have used batch.image and batch.aug so I am assuming we are storing both the image and the texture in the object. I think we have to make a new dataclass for this? If so, we can directly integrate the perlin noise in the image instead of during training step.

Looking at your code, it is not entirely clear to me how you are intending to use `batch.aug`. The current `(Image)-Batch` class does not contain the `aug` key. Were you planning on adding it? This would involve fundamental changes to Anomalib's dataset and dataclasses interface, which I'm not sure is the way to go here.

My personal preference would be to apply the augmentation within the `training_step`. This is similar to how Perlin noise augmentations are used in DRAEM model (see [here](https://github.com/open-edge-platform/anomalib/blob/main/src/anomalib/models/image/draem/lightning_model.py#L156)). In my view, this is the most flexible approach, as the training step retains access to both the original and the augmented image. At the same time, we ensure that the augmentation is always applied before passing the images to the model during training (contrary to adding the noise as an augmentation step on the dataset side, which could lead to problems if the dataset is not configured correctly by the user).

> I will also need to create a datamodule for the [DTD dataset](https://www.robots.ox.ac.uk/~vgg/data/dtd/). Should I add that in this PR only or create a new PR?

This might not be necessary. As far as I understand (but please correct me if I'm wrong), we only use the DTD dataset as a source of textures for the synthetic anomaly generation step. This means that we don't need any ground truth information for this dataset, just the raw images. So instead of implementing a dedicated dataset/datamodule, we can just randomly sample some images that we read from the file system in every step. In fact, this mechanism has already been implemented for DRAEM, which uses a similar approach. Please have a look at [this class](https://github.com/open-edge-platform/anomalib/blob/main/src/anomalib/data/utils/generators/perlin.py#L154) to see if you can use this for the augmentation step of your GLASS model (note that you can just pass the root of the DTD dataset on your file system, and the augmenter will automatically samples and read the images).

code-dev05:
> Looking at your code, it is not entirely clear to me how you are intending to use `batch.aug`. The current `(Image)-Batch` class does not contain the `aug` key. Were you planning on adding it? This would involve fundamental changes to Anomalib's dataset and dataclasses interface, which I'm not sure is the way to go here.

Yes, I was planning on adding a new dataclass interface but i will see the way it is done in the DRAEM model and try to get the same done here and will add the usage of the hooks for computing the center as soon as possible.

I will also make the suggested changes as soon as possible.

Thanks for the help.

src/anomalib/models/image/glass/__init__.py
code-dev05: the pre-commit hook is suggesting to change the line to `from .lightning_model import Glass as Glass
`. Is this correct?

src/anomalib/models/image/glass/lightning_model.py
code-dev05: should i put a default value for `input_shape` also? I think this one should be left for the user.

code-dev05:
@samet-akcay @ashwinvaidya17 @djdameln 
The PR is ready for review. Can you please review it?

src/anomalib/models/image/glass/loss.py
```diff
@@ -0,0 +1,88 @@
+# Copyright (C) 2022-2025 Intel Corporation
```
ashwinvaidya17, line 1: For files like these please follow a similar approach https://github.com/open-edge-platform/anomalib/blob/9887e1e6d302486d17419486c237e3e66e377dc0/src/anomalib/models/image/fastflow/torch_model.py#L3

src/anomalib/models/image/glass/loss.py
```diff
@@ -0,0 +1,88 @@
...
+    This is a implementation of Focal Loss with smooth label cross entropy supported which is proposed in
+    'Focal Loss for Dense Object Detection. (https://arxiv.org/abs/1708.02002)'
+        Focal_Loss= -1*alpha*(1-pt)*log(pt)
+    :param num_class:
+    :param alpha: (tensor) 3D or 4D the scalar factor for this criterion
+    :param gamma: (float,double) gamma > 0 reduces the relative loss for well-classified examples (p>0.5) putting more
+                    focus on hard misclassified example
+    :param smooth: (float,double) smooth value when cross entropy
+    :param balance_index: (int) balance class index, should be specific when alpha is float
+    :param size_average: (bool, optional) By default, the losses are averaged over each loss element in the batch.
```
ashwinvaidya17, line 20: We use the Google format for docstring in Anomalib

src/anomalib/models/image/glass/loss.py
```diff
@@ -0,0 +1,88 @@
...
+        self.gamma = gamma
+        self.balance_index = balance_index
+        self.smooth = smooth
+        self.size_average = size_average
+
+        if self.smooth is not None:
+            if self.smooth < 0 or self.smooth > 1.0:
+                raise ValueError("smooth value should be in [0,1]")
+
+    def forward(self, logit, target):
```
ashwinvaidya17, line 36: Please type these methods

ashwinvaidya17 (changes requested):
Thanks for the update. There are still a few things that need polishing

src/anomalib/models/components/feature_extractors/network_feature_extractor.py
samet-akcay, line 1: I'm not sure why this module is needed? Can you elaborate?

Is it not possible to use the existing feature extractors to achieve this? (anomalib, timm, torchvision?)

src/anomalib/models/image/__init__.py
```diff
@@ -84,4 +86,6 @@
     "Uflow",
     "VlmAd",
     "WinClip",
+    "Padim",
+    "Glass",
```
samet-akcay, line 90: I guess these are redundant?
```suggestion
```

src/anomalib/models/image/glass/backbones.py
```diff
@@ -0,0 +1,56 @@
+# Copyright (C) 2022-2025 Intel Corporation
```
samet-akcay, line 1:
```suggestion
# Copyright (C) 2025 Intel Corporation
```
It's because this module was created in 2025

src/anomalib/models/image/glass/backbones.py
```diff
@@ -0,0 +1,56 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import torchvision.models as models
+import timm
+
+"""copy from: https://github.com/cqylunlun/GLASS/blob/main/backbones.py
+This provides mechanism to import any of the given backbones using its name.
+"""
+_BACKBONES = {
```
samet-akcay, line 10: can you explain why we need this dict? Is it not possible to pass `backbone=<name-of-the-backbone>`, which would load the backbone model via timm or models? 

I think it would be good to stick to one package like timm, and use it throughout for simplicity.
So it would be like 

```python
backbone= "resnetv2_50x3_bitm"

....
backbone_model = timm.create_model(backbone, pretrained=True)
...
```

src/anomalib/models/image/glass/torch_model.py
```diff
@@ -0,0 +1,318 @@
...
+        pre_proj: int = 1,
+        dsc_layers: int = 2,
+        dsc_hidden: int = 1024,
+        dsc_margin: float = 0.5,
+    ) -> None:
+        super().__init__()
+
+        self.backbone = process_backbone(backbone)
+        self.layers = layers
+        self.input_shape = input_shape
```
samet-akcay, line 199: is `self.input_shape` used anywhere?

src/anomalib/models/image/glass/perlin.py
samet-akcay, line 1: Can you explain why this file is needed? Does Anomalib's perlin noise not work properly?

samet-akcay (commented):
Thanks a lot for your effort! I've got some initial feedback.

src/anomalib/models/image/glass/perlin.py
code-dev05: Oh, I did infact use the existing perlin noise implementation. I forgot to delete this file. Will do it as soon as posssible.

src/anomalib/models/image/glass/backbones.py
code-dev05: dict is used to provide the user the choice of either passing the name of the backbone model to be used or directly pass in a backbone model.
If it is a string, we can read from the dict. This is purely for user simplicity. I can remove it if you want.

src/anomalib/models/components/feature_extractors/network_feature_extractor.py
code-dev05: yes i believe the existing feature extractor can also be used. I will promptly make the change.

code-dev05:
Yes, I can see them. I will address those also and see what I can do to solve them.

samet-akcay:
> Yes, I can see them. I will address those also and see what I can do to solve them.

Great thanks

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,365 @@
...
+
+        This method is called at the start of training and computes a mean feature vector
+        that serves as a reference point for the normal class distribution.
+        """
+        dataloader = self.trainer.train_dataloader
+
+        with torch.no_grad():
+            for i, batch in enumerate(dataloader):
+                if i == 0:
+                    self.c = self.model.calculate_mean(batch.image)
```
samet-akcay:
```suggestion
                    self.c = self.model.calculate_mean(batch.image.to(self.device)
```

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,365 @@
...
+                else:
+                    self.c += self.model.calculate_mean(batch.image)
```
samet-akcay:
```suggestion
                    self.c += self.model.calculate_mean(batch.image.to(self.device)
```

samet-akcay: Alternatively another approach would be something like this?

```python
with torch.no_grad():
    means = []
    for batch in dataloader:
        images = batch.image.to(self.device)
        means.append(self.model.calculate_mean(images))
    self.c = torch.stack(means).mean(dim=0)
```

samet-akcay (changes requested):

hgaiser:
Hi @code-dev05 . I'm interested in trying out this implementation and possibly provide some feedback or fixes. Does it make sense to give this a go already or is it better to wait?

code-dev05:
hi @hgaiser , thank you for your interest in this implementation. I would suggest waiting for sometime until all the issues are fixed and the pr is merged.

hgaiser:
> hi @hgaiser , thank you for your interest in this implementation. I would suggest waiting for sometime until all the issues are fixed and the pr is merged.

Alright, thanks for replying. Let me know if you need an extra pair of hands on this.

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,326 @@
...
+            mining=mining,
+            noise=noise,
+            radius=radius,
+            p=p,
+        )
+
+        self.c = torch.tensor([1])
+        self.lr = lr
+
+        self.dev = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
samet-akcay: @hgaiser, there is a bug here to be fixed. For some reason, gpu is not detected for this model, which is why @code-dev05 had to add this line. However, this is temporary and not generic enough to run on any other GPUs (Intel, AMD or mac)

src/anomalib/models/components/feature_extractors/__init__.py
```diff
@@ -28,7 +28,6 @@
 
 from .timm import TimmFeatureExtractor
 from .utils import dryrun_find_featuremap_dims
-
```
samet-akcay, line 31: why is this line break removed?

src/anomalib/models/image/__init__.py
```diff
@@ -76,12 +77,13 @@
...
     "Ganomaly",
+    "Glass",
     "Padim",
     "Patchcore",
     "ReverseDistillation",
     "Stfpm",
     "Supersimplenet",
     "Uflow",
     "VlmAd",
-    "WinClip",
```
samet-akcay, line 89:
```suggestion
    "WinClip",
```

samet-akcay: No need for this change

src/anomalib/models/image/glass/__init__.py
```diff
@@ -0,0 +1,23 @@
+"""GLASS - Unsupervised anomaly detection via Gradient Ascent for Industrial Anomaly detection and localization.
```
samet-akcay, line 4:
```suggestion
# Copyright (C) 2025 Intel Corporation
# SPDX-License-Identifier: Apache-2.0

"""GLASS - Unsupervised anomaly detection via Gradient Ascent for Industrial Anomaly detection and localization.
```

src/anomalib/models/image/glass/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+      truncated projection
+    - A Local Anomaly Synthesis (LAS) module that overlays augmented textures onto images using Perlin noise masks
+    - A shared discriminator trained with features from normal, global, and local synthetic samples
+
+Paper: `A Unified Anomaly Synthesis Strategy with Gradient Ascent for Industrial Anomaly Detection and Localization
+<https://arxiv.org/pdf/2407.09359>`
+"""
+
+# Copyright (C) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
```
samet-akcay:
```suggestion

```

samet-akcay: We have merged a new PR, now that places the license at the top of the module. If you could update your PR, that'd be great

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,348 @@
+"""GLASS - Unsupervised anomaly detection via Gradient Ascent for Industrial Anomaly detection and localization.
```
samet-akcay, line 4:
```suggestion
# Copyright (C) 2025 Intel Corporation
# SPDX-License-Identifier: Apache-2.0

"""GLASS - Unsupervised anomaly detection via Gradient Ascent for Industrial Anomaly detection and localization.
```

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,348 @@
...
+    - A shared discriminator trained with features from normal, global, and local synthetic samples
+
+Paper: `A Unified Anomaly Synthesis Strategy with Gradient Ascent for Industrial Anomaly Detection and Localization
+<https://arxiv.org/pdf/2407.09359>`
+"""
+
+# Copyright (C) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any
```
samet-akcay, line 22:
```suggestion
"""

from typing import Any
```

samet-akcay: One line break between the docstring and the imports

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,348 @@
...
+            Defaults to `0.0001`.
+        step (int, optional): Number of gradient ascent steps for anomaly synthesis.
+            Defaults to `20`.
+        svd (int, optional): Flag to enable SVD-based feature projection.
+            Defaults to `0`.
+    """
+
+    def __init__(
+        self,
+        input_shape: tuple[int, int],
```
samet-akcay: Is `input_shape` still needed? Can it be read from self.pre_processor?

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,348 @@
...
+        anomaly_source_path: str,
```
samet-akcay:
```suggestion
        anomaly_source_path: str | None = None,
```

samet-akcay: Ideally a user should be able to import the model with the default parameters via `model = Glass()`

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,348 @@
...
+            dsc_margin=dsc_margin,
+            step=step,
+            svd=svd,
+            mining=mining,
+            noise=noise,
+            radius=radius,
+            p=p,
+        )
+
+        self.c = torch.tensor([1])
```
samet-akcay: Can you explain what `self.c` is? centroid? 

Would it make sense to move it inside the torch model as well? I see that you pass it to torch model like `self.model(batch.image, self.c)`?

samet-akcay: I see that you use this value `on_train_start` to compute? I think it can still be part of the torch model and you can compute it on train start. 

This way you would not need to pass the centroid every time in the `forward` method ?

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,348 @@
...
+        self.lr = lr
+
+        if pre_proj > 0:
```
samet-akcay: Can you use more descriptive name than `pre_proj`? I'm not sure what it exactly means

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,348 @@
...
+            self.proj_opt = optim.AdamW(
```
samet-akcay: same here, more descriptive name for `self.proj_opt` please. I'm terrible at decoding abbreviations  🙃

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,348 @@
...
+        Returns:
+            Optimizer: AdamW Optimizer for the discriminator.
+        """
+        return optim.AdamW(self.model.discriminator.parameters(), lr=self.lr * 2)
+
+    def training_step(
+        self,
+        batch: Batch,
+        batch_idx: int,
+    ) -> STEP_OUTPUT:
```
samet-akcay:
```suggestion
def training_step(self, batch: Batch, batch_idx: int) -> STEP_OUTPUT:
```

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,348 @@
...
+
+        Args:
+            batch (Batch): Input batch containing images and metadata
+            batch_idx (int): Index of the current batch
+
+        Returns:
+            STEP_OUTPUT: Dictionary containing loss values and metrics
+        """
+        del batch_idx
+        dsc_opt = self.optimizers()
```
samet-akcay: Same here for `dsc_opt` and the rest of the abbreviations.

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,348 @@
...
+        self.model.discriminator.train()
+
+        dsc_opt.zero_grad()
+        if self.proj_opt is not None:
+            self.proj_opt.zero_grad()
+        if self.backbone_opt is not None:
+            self.backbone_opt.zero_grad()
+
+        img = batch.image
+        true_loss, gaus_loss, bce_loss, focal_loss, loss = self.model(img, self.c)
```
samet-akcay:
```suggestion
        true_loss, gaus_loss, bce_loss, focal_loss, loss = self.model(batch.image, self.c)
```

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,348 @@
...
+        self.log("gaus_loss", gaus_loss, prog_bar=True)
+        self.log("bce_loss", bce_loss, prog_bar=True)
+        self.log("focal_loss", focal_loss, prog_bar=True)
+        self.log("loss", loss, prog_bar=True)
+
+    def validation_step(
+        self,
+        batch: Batch,
+        batch_idx: int,
+    ) -> STEP_OUTPUT:
```
samet-akcay:
```suggestion
    def validation_step(self, batch: Batch, batch_idx: int) -> STEP_OUTPUT:
```

src/anomalib/models/image/glass/loss.py
```diff
@@ -0,0 +1,169 @@
+"""Focal Loss for multi-class classification with optional label smoothing and class weighting.
```
samet-akcay, line 10: similar to above, license header at the top now

src/anomalib/models/image/glass/loss.py
```diff
@@ -0,0 +1,169 @@
...
+        gamma (float): Focusing parameter that reduces the relative loss for well-classified examples (gamma > 0).
+        smooth (float): Label smoothing factor for cross-entropy.
+        balance_index (int): Index of the class to balance when `alpha` is a float.
+        size_average (bool, optional): If True (default), the loss is averaged over the batch; otherwise, the loss is
+          summed.
+    """
+
+    def __init__(
+        self,
+        apply_nonlin: nn.Module | None = None,
```
samet-akcay:
```suggestion
        apply_nonlinearity: nn.Module | None = None,
```

src/anomalib/models/image/glass/loss.py
```diff
@@ -0,0 +1,169 @@
...
+            logit (torch.Tensor): The predicted logits of shape (B, C, ...) where B is batch size and C is the number of
+              classes.
+            target (torch.Tensor): The ground-truth class indices of shape (B, 1, ...) or broadcastable to logit.
+
+        Returns:
+            torch.Tensor: Computed focal loss value (averaged or summed depending on `size_average`).
+        """
+        if self.apply_nonlin is not None:
+            logit = self.apply_nonlin(logit)
+        num_class = logit.shape[1]
```
samet-akcay:
```suggestion
        num_classes = logit.shape[1]
```

samet-akcay: a bit nitpicky. rest of the `num_class` variables to be refactored in the method

src/anomalib/models/image/glass/loss.py
```diff
@@ -0,0 +1,169 @@
...
+        self.gamma = gamma
+        self.balance_index = balance_index
+        self.smooth = smooth
+        self.size_average = size_average
+
+        if self.smooth is not None and (self.smooth < 0 or self.smooth > 1.0):
+            msg = "smooth value should be in [0,1]"
+            raise ValueError(msg)
+
+    def forward(self, logit: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
```
samet-akcay: again, nitpicky comment, but would it make sense to use the plural form of `logit`, like `logits?`

src/anomalib/models/image/glass/torch_model.py
```diff
@@ -0,0 +1,741 @@
+"""GLASS - Unsupervised anomaly detection via Gradient Ascent for Industrial Anomaly detection and localization.
```
samet-akcay, line 4: license at the top of the module as above

src/anomalib/models/image/glass/torch_model.py
```diff
@@ -0,0 +1,741 @@
...
+            x (torch.Tensor): Input tensor of shape (B * N, ...).
+            batchsize (int): Original batch size.
+
+        Returns:
+            torch.Tensor: Reshaped tensor of shape (B, N, ...).
+        """
+        return x.reshape(batchsize, -1, *x.shape[1:])
+
+    @staticmethod
+    def score(x: torch.Tensor) -> torch.Tensor:
```
samet-akcay: How about?
```suggestion
    def compute_score(x: torch.Tensor) -> torch.Tensor:
```

src/anomalib/models/image/glass/torch_model.py
```diff
@@ -0,0 +1,741 @@
...
+        return [s.squeeze(0) for s in smoothed_stack]  # List of [H, W] tensors
+
+
+class GlassModel(nn.Module):
+    """PyTorch Implementation of the GLASS Model."""
+
+    def __init__(
+        self,
+        input_shape: tuple[int, int],  # (H, W)
+        anomaly_source_path: str,
```
samet-akcay: would it be possible to use default values for these as well?

ashwinvaidya17:
Can you also test if the model runs via the CLI?

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,348 @@
...
+            pre_processor=pre_processor,
+            post_processor=post_processor,
+            evaluator=evaluator,
+            visualizer=visualizer,
+        )
+
+        if layers is None:
+            layers = ["layer1", "layer2", "layer3"]
+
+        self.augmentor = PerlinAnomalyGenerator(anomaly_source_path)
```
ashwinvaidya17: This isn't used anywhere

src/anomalib/models/image/glass/torch_model.py
ashwinvaidya17:
```suggestion
        anomaly_source_path: str | None = None,
```

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,348 @@
...
+        dsc_hidden: int = 1024,
+        dsc_margin: float = 0.5,
+        pre_processor: PreProcessor | bool = True,
+        post_processor: PostProcessor | bool = True,
+        evaluator: Evaluator | bool = True,
+        visualizer: Visualizer | bool = True,
+        mining: int = 1,
+        noise: float = 0.015,
+        radius: float = 0.75,
+        p: float = 0.5,
```
ashwinvaidya17: Nitpicky but can you rename this to something more descriptive like `random_selection_prob` or something.

src/anomalib/models/image/glass/loss.py
```diff
@@ -0,0 +1,169 @@
...
+            alpha = torch.ones(num_classes, 1)
+        elif isinstance(alpha, (list | np.ndarray)):
+            assert len(alpha) == num_classes
+            alpha = torch.FloatTensor(alpha).view(num_classes, 1)
+            alpha = alpha / alpha.sum()
+        elif isinstance(alpha, float):
+            alpha = torch.ones(num_classes, 1)
+            alpha = alpha * (1 - self.alpha)
+            alpha[self.balance_index] = self.alpha
+        else:
```
ashwinvaidya17, line 141: What happens if `self.alpha` is of type `torch.Tensor`

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,326 @@
...
+        else:
+            self.projection_opt = None
+
+        if not pre_trained:
+            self.backbone_opt = optim.AdamW(
+                self.model.forward_modules["feature_aggregator"].backbone.parameters(),
+                self.learning_rate,
+            )
+        else:
+            self.backbone_opt = None
```
ashwinvaidya17, line 179: Since you are using manual step, it is better to explicitly set automatic optimization to false.
https://lightning.ai/docs/pytorch/stable/common/optimization.html#automatic-optimization
Like some of these models https://github.com/search?q=repo%3Aopen-edge-platform%2Fanomalib%20self.automatic_optimization&type=code

ashwinvaidya17, line 179: Maybe this can be moved to `configure_optimizers`. Have you tried whether `configure_optimizers` can return list with Nones? Like `list[optim, optim | None, optim | None]`. This way you can call `discriminator_opt, projection_opt, backbone_opt = self.optimizers()`.

src/anomalib/models/image/glass/torch_model.py
```diff
@@ -0,0 +1,742 @@
...
+
+        Returns:
+            None: The method updates `self.center` in-place with the computed center tensor.
+        """
+        self.forward_modules.eval()
+        self.center = torch.tensor([1])
+        with torch.no_grad():
+            for i, batch in enumerate(dataloader):
+                if self.pre_projection > 0:
+                    outputs = self.projection(self.generate_embeddings(batch.image.to(device))[0])
```
ashwinvaidya17, line 154: Have you confirmed that batch isn't on GPU to begin with? I believe lightning should automatically move the batch to the target training backend. In either case we can drop the `device` parameter to this method as `nn.Module` already has `self.device` property.

src/anomalib/models/image/glass/torch_model.py
```diff
@@ -0,0 +1,742 @@
...
+    dimensions_mapping = dryrun_find_featuremap_dims(
+        feature_extractor,
+        input_size,
+        layers,
+    )
+
+    return [dimensions_mapping[layer]["num_features"] for layer in layers]
+
+
+class Preprocessing(torch.nn.Module):
```
ashwinvaidya17: How about we split this into multiple files. We can introduce a components folder. Something like this https://github.com/open-edge-platform/anomalib/tree/main/src/anomalib/models/image/uninet/components
We can group dependant classes in the same file. Like `MiniMapper` can be in the same place as `Preprocessing`.

code-dev05: yes thats a good idea.  I will implement it.

src/anomalib/models/image/glass/loss.py
code-dev05: right now it will return an error message.
I will remove `torch.Tensor` from the data type in arguments and add `list` and `np.ndarray`.

src/anomalib/models/image/glass/lightning_model.py
code-dev05: `configure_optimizers` does not allow None in list.

src/anomalib/models/image/glass/torch_model.py
code-dev05: yes the batch is not on gpu in this method. I am not sure of the reason, thats why i had to include the device parameter.
also `GlassModel` does not have a `self.device`.

src/anomalib/models/image/glass/components/patch_maker.py
```diff
@@ -0,0 +1,90 @@
...
+    """Handles patch-based processing of feature maps.
+
+    This class provides utilities for converting feature maps into patches,
+    reshaping patch scores back to original dimensions, and computing global
+    anomaly scores from patch-wise predictions.
+
+    Attributes:
+        patchsize (int): Size of each patch (patchsize x patchsize).
+        stride (int or None): Stride used for patch extraction. Defaults to patchsize if None.
+        top_k (int): Number of top patch scores to consider. Used for score reduction.
```
rajeshgangireddy: Minor/nitpick : Please make sure the docstrings are in proper format and also that the params are in the order  `__init__(self, patchsize: int, top_k: int = 0, stride: int | None = None)`

src/anomalib/models/image/glass/components/patch_maker.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+
+    def __init__(self, patchsize: int, top_k: int = 0, stride: int | None = None) -> None:
+        self.patchsize = patchsize
+        self.stride = stride if stride is not None else patchsize
+        self.top_k = top_k
```
rajeshgangireddy: I do not see top_k being used in any of the methods. Am I missing something ?

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,328 @@
...
+            mining=mining,
+            noise=noise,
+            radius=radius,
+            random_selection_prob=random_selection_prob,
+        )
+
+        self.learning_rate = learning_rate
+
+        if pre_projection > 0:
+            self.projection_opt = optim.AdamW(
```
ashwinvaidya17:
```suggestion
            self.projection_opt = optim.Adam(
```

This is what the official implementation says https://github.com/cqylunlun/GLASS/blob/6af03b9d7f7b33a1aebd69cd4c30a41bf020a2d1/glass.py#L99

Though in the paper they use Adam for both feature adapter and the discriminator

Image

src/anomalib/models/image/glass/components/rescale_segmentor.py
```diff
@@ -0,0 +1,62 @@
...
+            scores = scores.unsqueeze(1)  # [N, 1, H, W]
+            scores = f.interpolate(
+                scores, size=self.target_size, mode="bilinear", align_corners=False,
+            )
+            patch_scores = scores.squeeze(1)  # [N, H, W]
+
+        patch_stack = patch_scores.unsqueeze(1)  # [N, 1, H, W]
+        smoothed_stack = kf.gaussian_blur2d(
+            patch_stack,
+            kernel_size=(5, 5),
```
ashwinvaidya17: In the original implementation I see that they use numpy's gaussian filter. https://github.com/cqylunlun/GLASS/blob/6af03b9d7f7b33a1aebd69cd4c30a41bf020a2d1/common.py#L65
According to the documentation, it accepts radius parameter https://docs.scipy.org/doc/scipy-1.16.0/reference/generated/scipy.ndimage.gaussian_filter.html#gaussian-filter
If it is None (as in the original implementation), then it is computed as
`2*(truncate*sigma)+ 1`. So, if my calculations are correct, the kernel size should be (33,33). It does not make a lot of difference but the anomaly maps are slightly better.

ashwinvaidya17 (commented):
Can you also add Readme.md and look at the merge conflicts in third-party.txt?

ashwinvaidya17:
Oh and also include a config file with default parameters. https://github.com/open-edge-platform/anomalib/tree/main/examples/configs/model

src/anomalib/models/image/glass/components/preprocessing.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (C) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""Maps input features to a fixed dimension using adaptive average pooling."""
+
+import torch
+import torch.nn.functional as f
+
+
+class MeanMapper(torch.nn.Module):
```
rajeshgangireddy, line 10: This looks very similar to the Aggregator class. 
Could you please tell me the difference between their usage ?
Could one be redundant ?

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,328 @@
...
+        post_processor: PostProcessor | bool = True,
+        evaluator: Evaluator | bool = True,
+        visualizer: Visualizer | bool = True,
+        mining: int = 1,
+        noise: float = 0.015,
+        radius: float = 0.75,
+        random_selection_prob: float = 0.5,
+        learning_rate: float = 0.0001,
+        step: int = 20,
+        svd: int = 0,
```
rajeshgangireddy: It looks like there are too many inputs args. Are there any params that can be removed (if they have least effect on results)?

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,328 @@
...
+        pre_trained: bool = True,
+        layers: list[str] | None = None,
+        pre_projection: int = 1,
+        discriminator_layers: int = 2,
+        discriminator_hidden: int = 1024,
+        discriminator_margin: float = 0.5,
+        pre_processor: PreProcessor | bool = True,
+        post_processor: PostProcessor | bool = True,
+        evaluator: Evaluator | bool = True,
+        visualizer: Visualizer | bool = True,
```
rajeshgangireddy, line 126: For uniformity, please move the following params to the end
```
        pre_processor: PreProcessor | bool = True,
        post_processor: PostProcessor | bool = True,
        evaluator: Evaluator | bool = True,
        visualizer: Visualizer | bool = True,
```

src/anomalib/models/image/glass/lightning_model.py
```diff
@@ -0,0 +1,328 @@
...
+            ValueError: If at least one dimension of ``center_crop_size`` is larger
+                than correspondent ``image_size`` dimension.
+
+        Example:
+            >>> pre_processor = Glass.configure_pre_processor(
+            ...     image_size=(256, 256)
+            ... )
+            >>> transformed_image = pre_processor(image)
+        """
+        image_size = image_size or (256, 256)
```
rajeshgangireddy: in the original implementation run (to reproduce the results) [here](https://github.com/cqylunlun/GLASS/blob/6af03b9d7f7b33a1aebd69cd4c30a41bf020a2d1/shell/run-mvtec.sh#L36)  , the authors use image size of (288, 288).

In the paper: 
image

rajeshgangireddy (commented):
Thanks for the effort. 
 I have some minor questions and comments

PR #2731, opened 2025-06-02
Docs: update custom pre-processing class to import from origin correctly

lucianchauvin:
Currently the docs import `get_exportable_transform` by means of `from anomalib.utils.transform import get_exportable_transform` which no longer exists as of `v2.0.0`. This file is now at `anomalib.pre_processing.utils.transform` as per 084331dad4e320c7dc200823aa8ae8857ec94d4b.


rajeshgangireddy (approved):
Thanks for the contribution.

PR #2813, opened 2025-07-07
🔄 refactor(model): memory usage optimisation

alfieroddan:
## 📝 Description

The memory bank models currently keep training embeddings, model's stored embeddings on device instead of releasing.

This can be seen most dramatically in Patchcore and PaDiM. We expect peak memory to drop in validation but actually we keep the training embeddings in memory:

Screenshot 2025-07-07 at 15 46 19

Screenshot 2025-07-07 at 15 46 30

As an example here is Patchcore's lightning fit function, and I've commented where we should delete duplicate embeddings:

```python
def fit(self) -> None:
        """Apply subsampling to the embedding collected from the training set.

        This method:
        1. Aggregates embeddings from all training batches
        2. Applies coreset subsampling to reduce memory requirements
        """
        logger.info("Aggregating the embedding extracted from the training set.")
        embeddings = torch.vstack(self.embeddings)

        logger.info("Applying core-set subsampling to get the embedding.")
        self.model.subsample_embedding(embeddings, self.coreset_sampling_ratio)

        # del self.embeddings <-- we should remove the extra ones
```

In addition to duplicate embeddings we should also be storing embeddings on the model not in lightning. So that the model can be used stand alone.

Lastly, the memory bank models are not designed for DDP or multi-gpu use. I have put default args to fix this.

- 🛠️ Fixes # (issue number)
- https://github.com/open-edge-platform/anomalib/issues/2673
- https://github.com/open-edge-platform/anomalib/discussions/2752

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [ ] 🔒 Security update

## ✅ Checklist


alfieroddan:
Ran some benchmarks just to check (bear in mind patchcore runs out of memory on benchmark script).

Screenshot 2025-07-10 at 15 20 57

Screenshot 2025-07-10 at 15 21 04

src/anomalib/models/image/dfm/torch_model.py
```diff
@@ -153,18 +154,33 @@ def __init__(
...
         Args:
-            dataset (torch.Tensor): Input dataset with shape
-                ``(n_samples, n_features)``.
+            dataset (torch.Tensor, optional): **[DEPRECATED]**
+                Input dataset with shape ``(n_samples, n_features)``.
+                This argument is deprecated and will be ignored. Instead,
+                the method now uses `self.memory_bank` as the input dataset.
         """
-        self.pca_model.fit(dataset)
+        if dataset is not None:
```
rajeshgangireddy: I am slightly concerned that users can pass a dataset to be used for training but it will not be used. Would replacing the memory bank with the passed dataset's feature be a good idea? Let's see what the rest of the team says.

src/anomalib/models/image/dfm/torch_model.py
```diff
@@ -202,17 +218,16 @@ def get_features(self, batch: torch.Tensor) -> torch.Tensor:
...
-        features = features.view(batch_size, -1).detach()
-        return features if self.training else (features, feature_shapes)
+        with torch.no_grad():
+            features = self.feature_extractor(batch)[self.layer]
+            batch_size = len(features)
+            if self.pooling_kernel_size > 1:
+                features = F.avg_pool2d(input=features, kernel_size=self.pooling_kernel_size)
+            feature_shapes = features.shape
+            features = features.view(batch_size, -1)
+        return features, feature_shapes
```
rajeshgangireddy, line 214: update method's return type to Tuple[torch.Tensor, torch.Size]

src/anomalib/models/image/patchcore/torch_model.py
```diff
@@ -33,6 +33,10 @@
         Coreset subsampling using k-center-greedy approach
 """
 
+# Copyright (C) 2022-2025 Intel Corporation
```
rajeshgangireddy: duplicated copyright header?

src/anomalib/models/image/dfm/torch_model.py
alfieroddan: This is the same in the old code; https://github.com/open-edge-platform/anomalib/blob/main/src/anomalib/models/image/dfm/torch_model.py#L156

I would also much prefer to call `fit` and not have a dataset arg.

src/anomalib/models/image/patchcore/torch_model.py
```diff
@@ -239,25 +253,37 @@ def reshape_embedding(embedding: torch.Tensor) -> torch.Tensor:
...
 
         Example:
-            >>> embedding = torch.randn(1000, 512)
-            >>> model.subsample_embedding(embedding, sampling_ratio=0.1)
+            >>> model.memory_bank = torch.randn(1000, 512)
+            >>> model.subsample_embedding(sampling_ratio=0.1)
             >>> model.memory_bank.shape
             torch.Size([100, 512])
         """
+        if embeddings is not None:
```
alfieroddan, line 272: Should we do a deprecated using decorators instead?

Like this:
```python
def set_theme(
    context: Literal["paper", "notebook", "talk", "poster"] | dict[str, Any] = "notebook",
    style: Literal["white", "dark", "whitegrid", "darkgrid", "ticks"] | dict[str, Any] = "darkgrid",
    palette: str | Sequence[ColorType] | None = "deep",
    font: str = "sans-serif",
    font_scale: float = 1,
    color_codes: bool = True,
    rc: dict[str, Any] | None = None,
) -> None: ...
@deprecated("Function `set` is deprecated in favor of `set_theme`")
def set(
    context: Literal["paper", "notebook", "talk", "poster"] | dict[str, Any] = "notebook",
    style: Literal["white", "dark", "whitegrid", "darkgrid", "ticks"] | dict[str, Any] = "darkgrid",
    palette: str | Sequence[ColorType] | None = "deep",
    font: str = "sans-serif",
    font_scale: float = 1,
    color_codes: bool = True,
    rc: dict[str, Any] | None = None,
) -> None: ...
```

(from seaborn)

rajeshgangireddy: Yes. 
There is a deprecate decorator in our utils: 
`from anomalib.utils.deprecation import deprecate`

alfieroddan: That deprecates functions not args, unless I put an explanation in `use`. I tried to keep the same methods in the refactor. I could update `from anomalib.utils.deprecation` so it does something like this?

```python
@deprecate(
    since="2.0.0",
    remove="2.1.0",
    use="fit(y=...)",
    reason="Argument `x` was renamed to `y` for clarity.",
    deprecated_args={"x": "y"},
)
def fit(y):
    print(f"Received y = {y}")

if __name__ ==
fit(y=100)
# Output:
# Received y = 100
fit(x=100)
# Output:
# DeprecationWarning: Argument `x` is deprecated since v1.2.0 and will be removed in v2.0.0; use `y` instead.
# DeprecationWarning: fit is deprecated since v1.2.0 and will be removed in v2.0.0. Use fit(y=...) instead.
# Argument `x` was renamed to `y` for clarity.
# Received y = 100
```

Let me know what you think

src/anomalib/models/image/dfkde/lightning_model.py
```diff
@@ -133,18 +131,15 @@ def training_step(self, batch: Batch, *args, **kwargs) -> None:
...
         # Return a dummy loss tensor
         return torch.tensor(0.0, requires_grad=True, device=self.device)
 
     def fit(self) -> None:
         """Fit KDE model to collected embeddings from the training set."""
-        embeddings = torch.vstack(self.embeddings)
-
         logger.info("Fitting a KDE model to the embedding collected from the training set.")
-        self.model.classifier.fit(embeddings)
+        self.model.fit_classifier()
```
samet-akcay: Would it be an idea to use `fit` to match the naming convention?
```suggestion
        self.model.fit()
```

src/anomalib/models/image/dfkde/torch_model.py
```diff
@@ -141,8 +142,34 @@ def forward(self, batch: torch.Tensor) -> torch.Tensor | InferenceBatch:
...
+            else:
+                new_bank = torch.cat((self.memory_bank, features), dim=0).to(self.memory_bank)
+                self.memory_bank = new_bank
             return features
 
         # 2. apply density estimation
         scores = self.classifier(features)
         return InferenceBatch(pred_score=scores)
+
+    def fit_classifier(self) -> None:
```
samet-akcay: same as above

```suggestion
    def fit(self) -> None:
```

src/anomalib/models/image/dfm/torch_model.py
samet-akcay: I think we can drop this `dataset`. It's quite the low-level implementation detail, no one is probably aware

src/anomalib/models/image/padim/lightning_model.py
```diff
@@ -154,19 +151,15 @@ def training_step(self, batch: Batch, *args, **kwargs) -> None:
...
         return torch.tensor(0.0, requires_grad=True, device=self.device)
 
     def fit(self) -> None:
         """Fit a Gaussian to the embedding collected from the training set."""
-        logger.info("Aggregating the embedding extracted from the training set.")
-        embeddings = torch.vstack(self.embeddings)
-
         logger.info("Fitting a Gaussian to the embedding collected from the training set.")
-        self.stats = self.model.gaussian.fit(embeddings)
+        self.model.fit_gaussian()
```
samet-akcay: same as above, let's use more generic fit so classifier can be modified if needed.
```suggestion
        self.model.fit()
```

src/anomalib/models/image/padim/torch_model.py
```diff
@@ -217,3 +223,23 @@ def generate_embedding(self, features: dict[str, torch.Tensor]) -> torch.Tensor:
         # subsample embeddings
         idx = self.idx.to(embeddings.device)
         return torch.index_select(embeddings, 1, idx)
+
+    def fit_gaussian(self) -> None:
```
samet-akcay:
```suggestion
    def fit(self) -> None:
```

src/anomalib/models/image/patchcore/torch_model.py
```diff
@@ -123,9 +124,8 @@ def __init__(
         ).eval()
         self.feature_pooler = torch.nn.AvgPool2d(3, 1, 1)
         self.anomaly_map_generator = AnomalyMapGenerator()
-
-        self.register_buffer("memory_bank", torch.empty(0))
         self.memory_bank: torch.Tensor
+        self.register_buffer("memory_bank", torch.Tensor())
```
samet-akcay: is `torch.Tensor()` really needed? `semgrep` and `bandit` don't like it from security and linter perspective.

src/anomalib/models/image/patchcore/torch_model.py
samet-akcay: yeah, current `@deprecate` handles classes and functions only. It would be good to support args as well. Instead of `deprecated_args`, I would prefer `args`. So, it would be something like this;

```python
@deprecate(args={"old_param": "new_param"}, since="2.1.0")
  def my_function(new_param, old_param=None):
      pass
```

samet-akcay (changes requested):
Looking great, thanks for the refactor!

Only have few minor comments.

ashwinvaidya17 (approved):

src/anomalib/models/image/dfm/torch_model.py
```diff
@@ -153,18 +154,29 @@ def __init__(
             layers=[layer],
         ).eval()
 
-    def fit(self, dataset: torch.Tensor) -> None:
+        self.memory_bank = torch.empty(0)
+
+    @deprecate(args={"dataset": None}, since="2.1.0")
+    def fit(self, dataset: torch.Tensor = None) -> None:
```
rajeshgangireddy: I thought we would completely remove the dataset param. 
Not many  uses/knows about it

alfieroddan: Ah I see, apologies for the confusion! Will change now :smile:

src/anomalib/models/image/dfm/torch_model.py
```diff
@@ -33,6 +33,7 @@
 
 from anomalib.data import InferenceBatch
 from anomalib.models.components import PCA, DynamicBufferMixin, TimmFeatureExtractor
+from anomalib.utils.deprecation import deprecate
```
samet-akcay:
```suggestion
```

samet-akcay (changes requested):
pre-commit fails with this line

PR #2843, opened 2025-07-23
fix(deploy): Fix discrepancy between lightning and standalone inferencers

ashwinvaidya17:
## 📝 Description

- 🛠️ Fixes https://github.com/open-edge-platform/anomalib/issues/2747

## Current status 
```bash
┏━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┓
┃ Model Name           ┃ Status ┃
┡━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━┩
│ cflow                │ ✅     │
│ dfm                  │ ✅     │
│ supersimplenet       │ ❌     │
│ ai_vad               │ ❌     │
│ padim                │ ✅     │
│ dfkde                │ ✅     │
│ fre                  │ ✅     │
│ win_clip             │ ❌     │
│ csflow               │ ✅     │
│ stfpm                │ ✅     │
│ patchcore            │ ✅     │
│ efficient_ad         │ ✅     │
│ draem                │ ✅     │
│ uflow                │ ✅     │
│ vlm_ad               │ ❌     │
│ cfa                  │ ❌     │
│ fastflow             │ ✅     │
│ dsr                  │ ❌     │
│ reverse_distillation │ ✅     │
│ ganomaly             │ ✅     │
│ fuvas                │ ❌     │
└──────────────────────┴────────┘
```

## ✨ Changes

- [x] 🧪 Tests (adding/modifying tests)
- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [ ] 🔒 Security update

## ✅ Checklist

### DIFF
--- a/tests/unit/deploy/test_inferencer.py
+++ b/tests/unit/deploy/test_inferencer.py
@@ -170,7 +170,12 @@ def test_inference_similarity(
     engine = Engine(logger=False, default_root_dir=project_path, devices=1)
 
     predict_dataset = PredictDataset(test_image_path)
-    predict_dataloader = DataLoader(predict_dataset, batch_size=1, collate_fn=predict_dataset.collate_fn)
+    predict_dataloader = DataLoader(
+        predict_dataset,
+        batch_size=1,
+        collate_fn=predict_dataset.collate_fn,
+        pin_memory=True,
+    )
     engine_pred: list[ImageBatch] = engine.predict(model, dataloaders=predict_dataloader, ckpt_path=ckpt_path("Padim"))
     engine_pred = engine_pred[0]
 


samet-akcay:
Maybe a different PR... but

Would it be an idea to follow [AHA](https://kentcdodds.com/blog/aha-programming) (Avoid Hasty Abstractions) principal for the model implementation? 

This would mean that we would implement the module hooks for each model instead of implementing them in `AnomalibModule`.

This would be inline with your suggestion that users/readers should be able to follow the training workflow for each model. 

Any thoughts? @ashwinvaidya17 @rajeshgangireddy

src/anomalib/models/image/fre/lightning_model.py
```diff
@@ -188,7 +188,7 @@ def trainer_arguments(self) -> dict[str, Any]:
                 - ``max_epochs``: ``220``
```
rajeshgangireddy: remove this - `max_epochs: 220`

tests/unit/deploy/test_inferencer.py
```diff
@@ -108,3 +111,77 @@ def test_openvino_inference(ckpt_path: Callable[[str], Path]) -> None:
...
+    predict_dataset = PredictDataset(test_image_path)
+    predict_dataloader = DataLoader(predict_dataset, batch_size=1, collate_fn=predict_dataset.collate_fn)
+    engine_pred: list[ImageBatch] = engine.predict(model, dataloaders=predict_dataloader, ckpt_path=ckpt_path("Padim"))
+    engine_pred = engine_pred[0]
+
+    torch_path = model.to_torch(export_root=project_path, model_file_name="torch")
+    torch_inferencer = TorchInferencer(torch_path, device="cpu")
+    torch_pred = torch_inferencer.predict(test_image_path)
+
+    openvino_path = model.to_openvino(export_root=project_path, model_file_name="openvino")
```
rajeshgangireddy: some of the transformer models (for example dinomaly) needs to be exported first to onnx and then to openvino (like how engine.export already does). 
This will resolve some of the unsupported operators.

rajeshgangireddy: I feel we should suggest the users to use engine.export() as standard way to export models and hence should compre these predictions  : 

engine.predict() 
engine.export(export_type=ExportType.OPENVINO) + openvino inferencer)


I think, if openvino model gives correct results, then onnx conversion was also correct. (I will think a bit more about this).

ashwinvaidya17: We convert all models to onnx first before converting to openvino. https://github.com/open-edge-platform/anomalib/blob/73073a0461feaf388de6211000e8cc24ad2ae2f6/src/anomalib/models/components/base/export_mixin.py#L254

ashwinvaidya17:
> Maybe a different PR... but
> 
> Would it be an idea to follow [AHA](https://kentcdodds.com/blog/aha-programming) (Avoid Hasty Abstractions) principal for the model implementation?
> 
> This would mean that we would implement the module hooks for each model instead of implementing them in `AnomalibModule`.
> 
> This would be inline with your suggestion that users/readers should be able to follow the training workflow for each model.
> 
> Any thoughts? @ashwinvaidya17 @rajeshgangireddy

Yeah. I was thinking that maybe we should try to move away from anomalib module as much as possible. This way the lightning models might be more transparent to the users

src/anomalib/models/image/fre/lightning_model.py
```diff
@@ -185,10 +185,9 @@ def trainer_arguments(self) -> dict[str, Any]:
         Returns:
             dict[str, Any]: Dictionary of trainer arguments:
                 - ``gradient_clip_val``: ``0``
-                - ``max_epochs``: ``220``
                 - ``num_sanity_val_steps``: ``0``
         """
-        return {"gradient_clip_val": 0, "max_epochs": 220, "num_sanity_val_steps": 0}
+        return {"gradient_clip_val": 0, "num_sanity_val_steps": 0}
```
samet-akcay, line 190: Is this change relevant to the issue?

ashwinvaidya17: No but I saw that this parameter was overriding the max epochs value in the tests. And, according to our current design, this isn't mandatory (more like recommended). So, like other models recommended `max_epochs` is set in the model config.

samet-akcay (approved):

rajeshgangireddy (approved):

PR #2968, opened 2025-09-24
🚀 feat(model): Make coreset selection for patchcore faster

rajeshgangireddy:
## 📝 Description

Coreset selection is the most time-consuming part of patchcore. Upon profiling, it is understood that there are two problems  
- F.pairwise_distance() is not fast and is not required as well. It is therefore replaced with  torch.linalg.norm(). 
- CPU-GPU/XPU Transfer Overhead: `urEnqueueKernelLaunch` (20.94% CPU time) and `urEnqueueUSMMemcpy` (5.67% CPU time) indicate significant kernel launch overhead and memory copy operations. To this, i have tried to reduce the calls to .item() and using indices on cpu while features are on a different device. 

## Results 
These are the results on different hardware with different dataset sizes. 
Numbers are averaged across 5 runs with different seeds.

image

## Additional info 

There is scope for cleaning up the code in `class KCenterGreedy` and to make it more better suited to patchcore. For example, `def select_coreset_idxs(self, selected_idxs: list[int] | None = None) ` is always called without any params when used in patchcore.  

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [ ] 🔒 Security update

## ✅ Checklist


rajeshgangireddy:
Setting it to draft yet, as I want to check and remove if we even need the case where centers are more than 1. 
I am 90% sure that this will not be needed - but i haven't spent time to fully confirm this.

rajeshgangireddy:
image

There is no difference between the accuracy metrics. 
The above numbers are averaged across 5 runs with different seed numbers.

rajeshgangireddy:
> Setting it to draft yet, as I want to check and remove if we even need the case where centers are more than 1. I am 90% sure that this will not be needed - but i haven't spent time to fully confirm this.

I have made some changes to cleanup dead code (multiple centers case).

src/anomalib/models/components/sampling/k_center_greedy.py
```diff
@@ -58,84 +57,82 @@ def reset_distances(self) -> None:
         """Reset minimum distances to None."""
         self.min_distances = None
 
-    def update_distances(self, cluster_centers: list[int]) -> None:
-        """Update minimum distances given cluster centers.
+    def update_distances(self, cluster_centers: int | torch.Tensor | None) -> None:
```
ashwinvaidya17: Nitpicky comment but cluster_center~s~

PR #2984, opened 2025-10-01
🚀 feat(inspect): Allow user to upload images, train model and display training progress

dwesolow:
## 📝 Description

This PR introduces a few features:
- allow user to upload images (temporary feature, to be removed when we support stream capture);
- list uploaded images;
- allow user to train a model when there is enough number of normal images;
- display proper message while training.

This is just starting point, we will improve that once we know that training is stable.

image

## ✨ Changes

- [X] 🚀 New feature (non-breaking change which adds functionality)
- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [ ] 🔒 Security update

## ✅ Checklist

- [X] 🏷️ My PR title follows conventional commit format.

### DIFF
--- a/app/ui/src/features/inspect/dataset/dataset-item/dataset-item.component.tsx
+++ b/app/ui/src/features/inspect/dataset/dataset-item/dataset-item.component.tsx
@@ -29,6 +29,7 @@ export const DatasetItem = ({ mediaItem }: DatasetItemProps) => {
         return <DatasetItemPlaceholder />;
     }
 
+    // TODO: Replace with thumbnail once supported by the backend
     const mediaUrl = `/api/projects/${mediaItem.project_id}/images/${mediaItem.id}/full`;
 
     return (


app/ui/src/features/inspect/dataset/dataset-list.component.tsx
```diff
@@ -0,0 +1,31 @@
...
+    mediaItems: MediaItem[];
+}
+
+export const DatasetList = ({ mediaItems }: DatasetItemProps) => {
+    const mediaItemsToRender =
+        mediaItems.length >= REQUIRED_NUMBER_OF_NORMAL_IMAGES_TO_TRIGGER_TRAINING
+            ? mediaItems
+            : Array.from({ length: REQUIRED_NUMBER_OF_NORMAL_IMAGES_TO_TRIGGER_TRAINING }).map((_, index) =>
+                  index <= mediaItems.length - 1 ? mediaItems[index] : undefined
+              );
```
MarkRedeman, line 17: np: I think this is a bit more readable, (it makes it easier to see that we render the media items by default, and potentially add some undefined placeholders)
```suggestion
    const mediaItemsToRender = [
        ...mediaItems,
        ...Array.from({ length: Math.max(0, REQUIRED_NUMBER_OF_NORMAL_IMAGES_TO_TRIGGER_TRAINING - mediaItems.length) })
    ]
```

app/ui/src/features/inspect/dataset/dataset-status-panel.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+        </InlineAlert>
+    );
+};
+
+const useAvailableModels = () => {
+    const AVAILABLE_MODELS = [
+        {
+            id: 'padim',
+            name: 'PADIM',
+        },
```
MarkRedeman, line 32: @ashwinvaidya17 can you comment on what other models should be available? The id that is used here will be used in the [model training service](https://github.com/open-edge-platform/anomalib/blob/feature/geti-inspect/app/backend/src/services/training_service.py#L115).

app/ui/src/features/inspect/dataset/dataset-status-panel.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+                <Content>
+                    <Text>{jobProgress.message}</Text>
+                </Content>
+            </InlineAlert>
+        );
+    }
+
+    return (
+        <InlineAlert variant='negative'>
+            <Heading>Training canceled</Heading>
```
MarkRedeman, line 151: I'd maybe also add a `if (jobProgres.status === 'cancelled')` here instead of relying that type inference guarantees that the status must be `cancelled`.
Having the explicit if statement here makes it a bit easier to read when reviewing outside of an IDE (I don't have to make sure that in the if statements above we handled all except the cancelled status)

app/ui/src/features/inspect/dataset/dataset-status-panel.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+    const startTrainingMutation = $api.useMutation('post', '/api/jobs:train');
+
+    const handleStartTraining = (body: { project_id: string; model_name: string }) => {
+        startTrainingMutation.mutate({
+            body,
+        });
+    };
+
+    // TODO: Investigate how to handle that case (local storage, poll for job status) after refreshing a page.
+    if (startTrainingMutation.data !== undefined) {
```
MarkRedeman, line 173: Can we use the  `'/api/jobs'` endpoint for this? Perhaps we can make it so that if the jobs endpoint returns 1 or more jobs for this project, then we render all of those jobs' progresses.

app/ui/src/features/inspect/dataset/dataset.component.tsx
```diff
@@ -1,55 +1,90 @@
...
+                    project_id: projectId,
+                },
+            },
+        });
+    };
+
+    const captureImages = (files: FileList | null) => {
+        if (files === null) return;
+
+        Array.from(files).forEach((file) => captureImage(file));
```
MarkRedeman, line 49: It might be better to change this function to an async function and await each mutation. Otherwise if the user selects 1000 images then we would immediately try to upload all of them.

Though given that we will likely remove this feature once the source pipeline is integrated I guess it won't be worth it to optimize this?

app/ui/src/features/inspect/dataset/dataset-status-panel.component.tsx
dwesolow: Yeah I was wondering implementing this but I wasn't sure but if u also agree then I'll implement that ;).

app/ui/src/features/inspect/dataset/dataset.component.tsx
dwesolow: Yeah, I wanted to keep it really simple since it's a temporary feature.

app/ui/src/features/inspect/dataset/dataset-list.component.tsx
dwesolow: Nice, will add it, thanks!

app/ui/src/features/inspect/dataset/dataset-status-panel.component.tsx
MarkRedeman: Let's go with this on the short term.
Long term we have to see how long training takes and how frequently we expect new models to be trained (and if they are trained in bulk), if they are then I think a job management panel like we have in Geti would be a good approach.

PR #3045, opened 2025-10-23
🚀 feat(inspect): Add progressbar

### LINKED ISSUE #3073: 📋 [TASK] Add training progress
### Describe the task

- Finalize decision between ws or SSE.
- Integrate progress bar with the UI

### Acceptance Criteria

Working progress bar with a clean design

### Priority

Low

### Related Epic

_No response_

### Estimated Time

_No response_

### Current Status

Not Started

### Additional Information

_No response_

ashwinvaidya17:
## 📝 Description

- 🛠️ Fixes #3073

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [ ] 🔒 Security update

## ✅ Checklist

### DIFF
--- a/application/ui/package-lock.json
+++ b/application/ui/package-lock.json
@@ -79,6 +79,7 @@
                 "typescript-eslint": "^8.31.1",
                 "vite": "^7.0.6",
                 "vite-plugin-svgr": "^4.3.0",
+                "vite-tsconfig-paths": "^5.1.4",
                 "vitest": "^3.2.4"
             },
             "engines": {
@@ -3208,13 +3209,13 @@
             }
         },
         "node_modules/@eslint/plugin-kit": {
-            "version": "0.3.5",
-            "resolved": "https://registry.npmjs.org/@eslint/plugin-kit/-/plugin-kit-0.3.5.tgz",
-            "integrity": "sha512-Z5kJ+wU3oA7MMIqVR9tyZRtjYPr4OC004Q4Rw7pgOKUOKkJfZ3O24nz3WYfGRpMDNmcOi3TwQOmgm7B7Tpii0w==",
+            "version": "0.3.2",
+            "resolved": "https://registry.npmjs.org/@eslint/plugin-kit/-/plugin-kit-0.3.2.tgz",
+            "integrity": "sha512-4SaFZCNfJqvk/kenHpI8xvN42DMaoycy4PzKc5otHxRswww1kAt82OlBuwRVLofCACCTZEcla2Ydxv8scMXaTg==",
             "dev": true,
             "license": "Apache-2.0",
             "dependencies": {
-                "@eslint/core": "^0.15.2",
+                "@eslint/core": "^0.15.0",
                 "levn": "^0.4.1"
             },
             "engines": {
@@ -3222,9 +3223,9 @@
             }
         },
         "node_modules/@eslint/plugin-kit/node_modules/@eslint/core": {
-            "version": "0.15.2",
-            "resolved": "https://registry.npmjs.org/@eslint/core/-/core-0.15.2.tgz",
-            "integrity": "sha512-78Md3/Rrxh83gCxoUc0EiciuOHsIITzLy53m3d9UyiW8y9Dj2D29FeETqyKA+BRK76tnTp6RXWb3pCay8Oyomg==",
+            "version": "0.15.0",
+            "resolved": "https://registry.npmjs.org/@eslint/core/-/core-0.15.0.tgz",
+            "integrity": "sha512-b7ePw78tEWWkpgZCDYkbqDOP8dmM6qe+AOC6iuJqlq1R/0ahMAeH3qynpnqKFGkMltrp44ohV4ubGyvLX28tzw==",
             "dev": true,
             "license": "Apache-2.0",
             "dependencies": {
@@ -4049,13 +4050,13 @@
             "license": "MIT"
         },
         "node_modules/@playwright/test": {
-            "version": "1.56.1",
-            "resolved": "https://registry.npmjs.org/@playwright/test/-/test-1.56.1.tgz",
-            "integrity": "sha512-vSMYtL/zOcFpvJCW71Q/OEGQb7KYBPAdKh35WNSkaZA75JlAO8ED8UN6GUNTm3drWomcbcqRPFqQbLae8yBTdg==",
+            "version": "1.54.1",
+            "resolved": "https://registry.npmjs.org/@playwright/test/-/test-1.54.1.tgz",
+            "integrity": "sha512-FS8hQ12acieG2dYSksmLOF7BNxnVf2afRJdCuM1eMSxj6QTSE6G4InGF7oApGgDb65MX7AwMVlIkpru0yZA4Xw==",
             "dev": true,
             "license": "Apache-2.0",
             "dependencies": {
-                "playwright": "1.56.1"
+                "playwright": "1.54.1"
             },
             "bin": {
                 "playwright": "cli.js"
@@ -13722,6 +13723,12 @@
                 "url": "https://github.com/sponsors/ljharb"
             }
         },
+        "node_modules/globrex": {
+            "version": "0.1.2",
+            "resolved": "https://registry.npmjs.org/globrex/-/globrex-0.1.2.tgz",
+            "integrity": "sha512-uHJgbwAMwNFf5mLst7IWLNg14x1CkeqglJb/K3doi4dw6q2IvAAmM/Y81kevy83wP+Sst+nutFTYOGg3d1lsxg==",
+            "dev": true
+        },
         "node_modules/gopd": {
             "version": "1.2.0",
             "resolved": "https://registry.npmjs.org/gopd/-/gopd-1.2.0.tgz",
@@ -16923,13 +16930,13 @@
             }
         },
         "node_modules/playwright": {
-            "version": "1.56.1",
-            "resolved": "https://registry.npmjs.org/playwright/-/playwright-1.56.1.tgz",
-            "integrity": "sha512-aFi5B0WovBHTEvpM3DzXTUaeN6eN0qWnTkKx4NQaH4Wvcmc153PdaY2UBdSYKaGYw+UyWXSVyxDUg5DoPEttjw==",
+            "version": "1.54.1",
+            "resolved": "https://registry.npmjs.org/playwright/-/playwright-1.54.1.tgz",
+            "integrity": "sha512-peWpSwIBmSLi6aW2auvrUtf2DqY16YYcCMO8rTVx486jKmDTJg7UAhyrraP98GB8BoPURZP8+nxO7TSd4cPr5g==",
             "dev": true,
             "license": "Apache-2.0",
             "dependencies": {
-                "playwright-core": "1.56.1"
+                "playwright-core": "1.54.1"
             },
             "bin": {
                 "playwright": "cli.js"
@@ -16942,9 +16949,9 @@
             }
         },
         "node_modules/playwright-core": {
-            "version": "1.56.1",
-            "resolved": "https://registry.npmjs.org/playwright-core/-/playwright-core-1.56.1.tgz",
-            "integrity": "sha512-hutraynyn31F+Bifme+Ps9Vq59hKuUCz7H1kDOcBs+2oGguKkWTU50bBWrtz34OUWmIwpBTWDxaRPXrIXkgvmQ==",
+            "version": "1.54.1",
+            "resolved": "https://registry.npmjs.org/playwright-core/-/playwright-core-1.54.1.tgz",
+            "integrity": "sha512-Nbjs2zjj0htNhzgiy5wu+3w09YetDx5pkrpI/kZotDlDUaYk0HVA5xrBVPdow4SAUIlhgKcJeJg4GRKW6xHusA==",
             "dev": true,
             "license": "Apache-2.0",
             "bin": {
@@ -19231,14 +19238,14 @@
             "license": "MIT"
         },
         "node_modules/tinyglobby": {
-            "version": "0.2.15",
-            "resolved": "https://registry.npmjs.org/tinyglobby/-/tinyglobby-0.2.15.tgz",
-            "integrity": "sha512-j2Zq4NyQYG5XMST4cbs02Ak8iJUdxRM0XI5QyxXuZOzKOINmWurp3smXu3y5wDcJrptwpSjgXHzIQxR0omXljQ==",
+            "version": "0.2.14",
+            "resolved": "https://registry.npmjs.org/tinyglobby/-/tinyglobby-0.2.14.tgz",
+            "integrity": "sha512-tX5e7OM1HnYr2+a2C/4V0htOcSQcoSTH9KgJnVvNm5zm/cyEWKJ7j7YutsH9CxMdtOkkLFy2AHrMci9IM8IPZQ==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
-                "fdir": "^6.5.0",
-                "picomatch": "^4.0.3"
+                "fdir": "^6.4.4",
+                "picomatch": "^4.0.2"
             },
             "engines": {
                 "node": ">=12.0.0"
@@ -19248,14 +19255,11 @@
             }
         },
         "node_modules/tinyglobby/node_modules/fdir": {
-            "version": "6.5.0",
-            "resolved": "https://registry.npmjs.org/fdir/-/fdir-6.5.0.tgz",
-            "integrity": "sha512-tIbYtZbucOs0BRGqPJkshJUYdL+SDH7dVM8gjy+ERp3WAUjLEFJE+02kanyHtwjWOnwrKYBiwAmM0p4kLJAnXg==",
+            "version": "6.4.6",
+            "resolved": "https://registry.npmjs.org/fdir/-/fdir-6.4.6.tgz",
+            "integrity": "sha512-hiFoqpyZcfNm1yc4u8oWCf9A2c4D3QjCrks3zmoVKVxpQRzmPNar1hUJcBG2RQHvEVGDN+Jm81ZheVLAQMK6+w==",
             "dev": true,
             "license": "MIT",
-            "engines": {
-                "node": ">=12.0.0"
-            },
             "peerDependencies": {
                 "picomatch": "^3 || ^4"
             },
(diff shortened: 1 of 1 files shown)


application/backend/src/pydantic_models/job.py
```diff
@@ -28,6 +28,7 @@ class Job(BaseIDModel):
     type: JobType = JobType.TRAINING
     progress: int = Field(default=0, ge=0, le=100, description="Progress percentage from 0 to 100")
     status: JobStatus = JobStatus.PENDING
+    stage: str = "idle"
```
maxxgx: Please create a `JobStage` enum so that pydantic can automatically validate this field

application/backend/src/api/endpoints/job_endpoints.py
```diff
@@ -42,3 +42,21 @@ async def get_job_logs(
...
+@job_router.get("/{job_id}/progress")
+async def get_job_progress(
+    job_id: Annotated[UUID, Depends(get_job_id)],
+    job_service: Annotated[JobService, Depends(get_job_service)],
+) -> StreamingResponse:
+    """Endpoint to get the progress of a job by its ID"""
+    return StreamingResponse(job_service.stream_progress(job_id=job_id), media_type="text/event-stream")
+
+
+@job_router.post("/{job_id}/cancel")
```
maxxgx:
```suggestion
@job_router.post("/{job_id}:cancel", status_code=status.HTTP_202_ACCEPTED)
```

For POST endpoints we use the colon convention `:cancel` instead of path `/cancel`. This is to distinguish actions (POST) from queries (GET).

Response should return `HTTP_202_ACCEPTED` because cancelling is not a synchronous operation, i.e. job is not truly cancelled yet when response is returned.

application/backend/src/pydantic_models/job.py
```diff
@@ -46,6 +47,10 @@ class JobSubmitted(BaseModel):
     job_id: UUID
 
 
+class JobCancelled(BaseModel):
+    job_id: UUID
```
maxxgx, line 50: It would be nice to also return a message. You add a pydantic computed field:
```python
from pydantic import BaseModel, computed_field

class JobCancelled(BaseModel):
    job_id: UUID
    
    @computed_field
    def message(self) -> str:
        return f" Job with ID `{self.job_id}' marked as cancelled."
```

application/backend/src/db/schema.py
```diff
@@ -59,6 +59,7 @@ class JobDB(Base):
     id: Mapped[str] = mapped_column(primary_key=True, default=lambda: str(uuid4()))
     project_id: Mapped[str] = mapped_column(ForeignKey("projects.id"))
     type: Mapped[str] = mapped_column(String(64), nullable=False)
+    stage: Mapped[str] = mapped_column(String(64), nullable=False)  # training, validation, test
```
MarkRedeman: I see it's also `idle` when the job hasn't started yet. Perhaps we can make this an enum? I.e. do something like `stage: Mapped[JobStage]` where `JobStage = 'idle' | 'training' | 'validation' | 'test'.

application/backend/src/services/job_service.py
```diff
@@ -107,3 +118,37 @@ async def is_job_still_running():
...
+            job = await repo.get_by_id(job_id)
+            if job is None:
+                raise ResourceNotFoundException(resource_id=job_id, resource_name="job")
+            while still_running:
+                now = loop.time()
+                if now - last_status_check > status_check_interval:
+                    cached_still_running = await cls.is_job_still_running(job_id=job_id)
+                    last_status_check = now
+                still_running = cached_still_running
+                yield json.dumps({"progress": job.progress, "stage": job.stage})
```
maxxgx: Is JSON necessary here? I'd expect `media_type="text/event-stream"` would take care of the conversion. If so, it would be nicer to return a pydantic model instead of constructing a dictionary here

application/backend/src/services/training_service.py
```diff
@@ -159,6 +174,24 @@ def _train_model(model: Model) -> Model | None:
...
+        synchronization_parameters: ProgressSyncParams,
+    ) -> None:
+        while True:
+            progress = synchronization_parameters.get_progress()
+            stage = synchronization_parameters.get_stage()
+            if not await job_service.is_job_still_running(job_id=job_id):
+                logger.info("Job cancelled, stopping progress sync")
+                synchronization_parameters.set_cancel_training_event()
+                break
+            logger.info(f"Syncing progress with db: {progress}% - {stage}")
```
maxxgx: I'd lower the log level to at least debug

application/backend/src/services/job_service.py
```diff
@@ -107,3 +118,37 @@ async def is_job_still_running():
...
+    async def stream_progress(cls, job_id: UUID | str) -> Coroutine[Any, Any, AsyncContentStream]:
+        """Stream the progress of a job by its ID"""
+        loop = asyncio.get_running_loop()
+        status_check_interval = 2.0  # seconds
+        last_status_check = 0.0
+        cached_still_running = True
+        still_running = True
+        async with get_async_db_session_ctx() as session:
+            repo = JobRepository(session)
+            job = await repo.get_by_id(job_id)
```
MarkRedeman: I think it might be safer to do `await self.get_job_by_id(job_id=job_id)`. This way you can avoid holding the `session` for too long (i.e. this might lock the database and therefore not allow other processes to make changes to it).

application/backend/src/services/training_service.py
```diff
@@ -159,6 +174,24 @@ def _train_model(model: Model) -> Model | None:
...
         return model
 
+    @classmethod
+    async def _sync_progress_with_db(
+        cls,
+        job_service: JobService,
+        job_id: UUID,
+        synchronization_parameters: ProgressSyncParams,
+    ) -> None:
+        while True:
```
MarkRedeman: Could this not be done whenever `GetiInspectProgressCallback` is called? I.e. instead of checking that something changed every 0.1 second, could you not make the change to the database anytime that you see it changes at the callback code? 

Perhaps to avoid calling the job service too much you could keep a local copy of the `job` model, and then compare it to the new data. If the new data is different from your local version of `job` then you send an update to the service.

application/backend/src/services/training_service.py
```diff
@@ -159,6 +174,24 @@ def _train_model(model: Model) -> Model | None:
...
+            progress = synchronization_parameters.get_progress()
```
maxxgx: Note that job progress is an `int`. I think it's fine if `synchronization_parameters.get_progress()` returns a more granular float value. However, you can avoid to update the job progress if the `int` values doesn't change.

application/backend/src/services/training_service.py
```diff
@@ -159,6 +174,24 @@ def _train_model(model: Model) -> Model | None:
...
+            stage = synchronization_parameters.get_stage()
+            if not await job_service.is_job_still_running(job_id=job_id):
+                logger.info("Job cancelled, stopping progress sync")
+                synchronization_parameters.set_cancel_training_event()
+                break
+            logger.info(f"Syncing progress with db: {progress}% - {stage}")
+            await job_service.update_job_status(job_id=job_id, status=JobStatus.RUNNING, progress=progress, stage=stage)
+            await asyncio.sleep(0.1)
```
maxxgx: 10 updates and 20 DB operations per second seems excessive. 

Maybe 0.5s is already responsive enough. 

This scenario makes good case to use web sockets 😃

application/backend/src/utils/callbacks.py
```diff
@@ -0,0 +1,182 @@
...
+    def get_stage(self) -> str:
+        return self.stage.value.decode("utf-8")
+
+    def set_progress(self, progress: float) -> None:
+        with self.progress.get_lock():
+            self.progress.value = progress
+        logger.info("Set progress: %s", progress)
+
+    def get_progress(self) -> float:
+        return self.progress.value
```
maxxgx: Nitpick: prefer `@property` instead of getters and setters

application/backend/src/db/schema.py
ashwinvaidya17: I was thinking about it but we also have fit, predict, etc. So one option is to have explicit enum or we can just forward the string from the trainer since it is used as a label anyways.

application/backend/src/services/training_service.py
ashwinvaidya17: job service uses async so I am not sure how it will play inside the callback. I'll try updating job inside the callback.

application/backend/src/services/job_service.py
maxxgx: I agree with @MarkRedeman, you can use `await self.get_job_by_id` instead of creating the repo. 

You don't need the loop to be inside of a db session context

application/backend/src/services/training_service.py
```diff
@@ -133,15 +146,17 @@ def _train_model(model: Model) -> Model | None:
...
+            callbacks=[GetiInspectProgressCallback(synchronization_parameters)],
         )
 
         # Execute training and export
         export_format = ExportType.OPENVINO
 
         # Capture pytorch stdout logs into logger
         with redirect_stdout(LoggerStdoutWriter()):  # type: ignore[type-var]
-            engine.train(model=anomalib_model, datamodule=datamodule)
+            engine.fit(model=anomalib_model, datamodule=datamodule)
```
MarkRedeman, line 180: What difference does this change make?

ashwinvaidya17: train calls fit + test. We don't need test as we don't have anomalous images. Also, unfortunately, the default behaviour of test is to write the predictions to the file. So, it takes more time and we don't need that in Geti Inspect

application/backend/src/services/training_service.py
ashwinvaidya17: No, I can't move it inside the callback.

application/backend/src/db/schema.py
```diff
@@ -59,6 +61,7 @@ class JobDB(Base):
     id: Mapped[str] = mapped_column(primary_key=True, default=lambda: str(uuid4()))
     project_id: Mapped[str] = mapped_column(ForeignKey("projects.id"))
     type: Mapped[str] = mapped_column(String(64), nullable=False)
+    stage: Mapped[JobStage] = mapped_column(Enum(JobStage), nullable=False)
```
maxxgx: What's the advantage of declaring an Enum column in the DB schema?

What is the values that is actually stored in DB, e.g. int or str?

I wonder what happens if you add/remove job stages, will queries still work?

application/backend/src/pydantic_models/job.py
```diff
@@ -23,11 +23,26 @@ class JobStatus(StrEnum):
...
+
+    See ``lightning.pytorch.trainer.states.RunningStage`` for more details.
+    """
+
+    IDLE = "idle"
+    TRAINING = "train"
+    SANITY_CHECKING = "sanity_check"
+    VALIDATING = "validate"
+    TESTING = "test"
+    PREDICTING = "predict"
```
maxxgx: Not sure if we are going to have other types of jobs in the future, but these are only related to training and there is overlap with `job.status`. Perhaps, it's more appropriate to store the training stage as part of `job.message` string

application/backend/src/pydantic_models/job.py
```diff
@@ -46,7 +61,15 @@ class JobSubmitted(BaseModel):
     job_id: UUID
 
 
+class JobCancelled(BaseModel):
+    job_id: UUID
+
+    @computed_field
+    def message(self) -> str:
+        return f" Job with ID `{self.job_id}' marked as cancelled."
```
maxxgx:
```suggestion
        return f"Job with ID `{self.job_id}' marked as cancelled."
```

application/backend/src/services/training_service.py
```diff
@@ -63,31 +65,46 @@ async def _run_training_job(cls, job: Job, job_service: JobService) -> Model:
...
+            logger.debug("Syncing progress with db stopped")
+            synchronization_task.cancel()
             return await model_service.create_model(trained_model)
         except Exception as e:
             logger.exception("Failed to train pending training job: %s", e)
             await job_service.update_job_status(
                 job_id=job.id, status=JobStatus.FAILED, message=f"Failed with exception: {str(e)}"
             )
+            logger.debug("Syncing progress with db stopped")
+            synchronization_task.cancel()
```
maxxgx: You could put this 2 lines in a `finally` block

application/backend/src/utils/callbacks.py
```diff
@@ -0,0 +1,184 @@
...
+        self._check_cancel_training(trainer)
+
+    def on_validation_end(self, trainer: Trainer, pl_module: LightningModule) -> None:
+        """Called when validation ends."""
+        self._check_cancel_training(trainer)
+
+    # Test callbacks
+    def on_test_start(self, trainer: Trainer, pl_module: LightningModule) -> None:
+        """Called when testing starts."""
+        self._send_progress(0, JobStage.TESTING)
```
maxxgx: How come the progress is `0` here?

application/backend/src/pydantic_models/job.py
ashwinvaidya17: We will have to extract it from the string if we want to show it in the progress bar.

application/backend/src/utils/callbacks.py
ashwinvaidya17: The idea is to treat each stage as having different progress instead of clubbing train->test as a single progress.
Currently the progressbar will first show the training progress, and then the testing progress. This just makes the progress a bit verbose.

application/backend/src/db/schema.py
ashwinvaidya17: Hmm not sure how it will respond when we modify it. I was following Mark's suggestion https://github.com/open-edge-platform/anomalib/pull/3045#discussion_r2454204524. Maybe we can achieve this in a different manner?

application/ui/src/routes/inspect/inspect.tsx
```diff
@@ -8,30 +8,34 @@ import { InferenceProvider } from '../../features/inspect/inference-provider.com
...
-        >
-            <SelectedMediaItemProvider>
-                <InferenceProvider>
-                    <Toolbar />
-                    <InferenceResult />
-                    <Sidebar />
-                </InferenceProvider>
-            </SelectedMediaItemProvider>
-        </Grid>
+        <div style={{ display: 'flex', flexDirection: 'column', height: '100%' }}>
```
MarkRedeman: Let's revert this change, the extra div is not needed.

application/ui/src/routes/inspect/inspect.tsx
```diff
@@ -8,30 +8,34 @@ import { InferenceProvider } from '../../features/inspect/inference-provider.com
...
+            <Grid
+                areas={['toolbar sidebar', 'canvas sidebar']}
```
MarkRedeman:
```suggestion
                areas={['toolbar sidebar', 'canvas sidebar', 'footer sidebar']}
```

The `Grid` component is essentially a wrapper around css's grid properties we can add a new footer grid cell that is used for displaying the sidebar.

application/ui/src/routes/inspect/inspect.tsx
```diff
@@ -8,30 +8,34 @@ import { InferenceProvider } from '../../features/inspect/inference-provider.com
...
+                rows={['size-800', 'minmax(0, 1fr)']}
```
MarkRedeman: With the above change this should make it so that the footer takes up a minimal height, (do check this though, we might need to tweak this)
```suggestion
                rows={['size-800', 'minmax(0, 1fr)', 'auto']}
```

application/ui/src/routes/inspect/inspect.tsx
```diff
@@ -8,30 +8,34 @@ import { InferenceProvider } from '../../features/inspect/inference-provider.com
...
+            >
+                <SelectedMediaItemProvider>
+                    <InferenceProvider>
+                        <Toolbar />
+                        <InferenceResult />
+                        <Sidebar />
+                    </InferenceProvider>
+                </SelectedMediaItemProvider>
+            </Grid>
+            <StatusBar />
```
MarkRedeman:
```suggestion
                        <Sidebar />
                        <Footer />
                    </InferenceProvider>
                </SelectedMediaItemProvider>
            </Grid>

```

Let's move the statusbar into a `Footer` component and put it inside of the grid.

application/ui/src/features/inspect/statusbar/statusbar.component.tsx
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Flex, View } from '@geti/ui';
+
+import { ProgressBarItem } from './items/progressbar.component';
+
+export const StatusBar = () => {
+    return (
+        <View gridArea={'statusbar'} backgroundColor={'gray-100'} width={'100%'} height={'30px'} overflow={'hidden'}>
```
MarkRedeman:
```suggestion
export const Footer = () => {
    return (
        <View gridArea={'footer'} backgroundColor={'gray-100'} width={'100%'} height={'size-400'} overflow={'hidden'}>
```

For spacing we try to adhere to [spectrum's styling](https://react-spectrum.adobe.com/react-spectrum/styling.html) guidelines. This makes it so that the applications' UI is more consistent.

application/ui/src/features/inspect/statusbar/items/progressbar.component.tsx
MarkRedeman, line 1: Currently this component feels very non-idomatic react code. I will send you a component that should follow more of our guidelines.

application/backend/src/db/schema.py
maxxgx: I've run some testing with `Enum` columns. It seems to be safe, at least with `StrEnum`, it saves the enum as a string. Adding/removing values in the code does not seem to require the DB to be updated.

application/backend/src/pydantic_models/job.py
maxxgx: The job message can be reset based on the current stage. So you could just use the whole message.

ATM, it seems you're not setting `job.message` at all.

In Geti classic, `job.step.message` was displayed: 
image

Looks like `JobStage` corresponds to `job.step` in the image above. However, our job progress is global, so we don't have this mechanism of tracking individual stage progress. Not sure how well this works in terms of UX: users would see progress resetting from 100 to 0 (e.g. from stage "training" to "validate"), but they don't know the total number of stages elapsed or remaining.

application/backend/src/utils/callbacks.py
maxxgx: Added a comment above regarding this "staged" progress tracking approach.

application/backend/src/services/training_service.py
```diff
@@ -70,18 +72,32 @@ async def _run_training_job(cls, job: Job, job_service: JobService) -> Model:
...
+                synchronization_parameters=synchronization_parameters,
+            )
             if trained_model is None:
                 raise ValueError("Training failed - model is None")
 
             await job_service.update_job_status(
                 job_id=job.id, status=JobStatus.COMPLETED, message="Training completed successfully"
             )
+            logger.debug("Syncing progress with db stopped")
+            synchronization_task.cancel()
```
maxxgx: I think the log would show twice because finally block is run in any case. You can move `synchronization_task.cancel()` under finally and remove the conditional task cancellation there.

maxxgx (approved):

AlexanderBarabanov (approved):

PR #3053, opened 2025-10-24
feat(inspect): Add train model dialog, job logs and models dataset

MarkRedeman:
## 📝 Description

This PR contains the changes I made for yesterday's demo. Please check the individual commits for more specific changes.
In summary this:
- Adds a "Train model" button that opens a dialog to let the user choose their model
- Adds a "Show logs" button that can be used to see training logs in the UI
- Use query parameters to determine which tab is opened
- Adds the models tab showing all trained models and non-completed jobs
- Invoke inference when changing to a different model

On the backend I made two changes:
- Update the end time of a job so that we can compute its duration (not yet used in the UI)
- Change the formatting of the SSE endpoint

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [ ] 🔒 Security update

## ✅ Checklist

### DIFF
--- a/application/ui/src/features/inspect/dataset/dataset.component.tsx
+++ b/application/ui/src/features/inspect/dataset/dataset.component.tsx
@@ -7,6 +7,7 @@ import { useQueryClient } from '@tanstack/react-query';
 
 import { TrainModelButton } from '../train-model/train-model-button.component';
 import { DatasetList } from './dataset-list.component';
+import { REQUIRED_NUMBER_OF_NORMAL_IMAGES_TO_TRIGGER_TRAINING } from './utils';
 
 const useMediaItems = () => {
     const { projectId } = useProjectIdentifier();
@@ -53,7 +54,7 @@ const UploadImages = () => {
         await queryClient.invalidateQueries({ queryKey: imagesOptions.queryKey });
         const images = await queryClient.ensureQueryData(imagesOptions);
 
-        if (images.media.length >= 20) {
+        if (images.media.length >= REQUIRED_NUMBER_OF_NORMAL_IMAGES_TO_TRIGGER_TRAINING) {
             toast({
                 title: 'Train',
                 type: 'info',
--- a/application/ui/src/features/inspect/inference-provider.component.tsx
+++ b/application/ui/src/features/inspect/inference-provider.component.tsx
@@ -70,10 +70,8 @@ export const InferenceProvider = ({ children }: InferenceProviderProps) => {
     const onSetSelectedModelId = (modelId: string | undefined) => {
         setSelectedModelId(modelId);
 
-        if (modelId) {
-            if (selectedMediaItem) {
-                onInference(selectedMediaItem, modelId);
-            }
+        if (modelId && selectedMediaItem) {
+            onInference(selectedMediaItem, modelId);
         }
     };
     return (
--- a/application/ui/src/features/inspect/jobs/show-job-logs.component.tsx
+++ b/application/ui/src/features/inspect/jobs/show-job-logs.component.tsx
@@ -96,7 +96,7 @@ const JobLogsDialog = ({ close, jobId }: { close: () => void; jobId: string }) =
                     padding='size-200'
                     backgroundColor={'gray-50'}
                     UNSAFE_style={{
-                        fontSize: '11px',
+                        fontSize: 'var(--spectrum-global-dimension-static-size-130)',
                     }}
                 >
                     <Suspense fallback={<Loading mode='inline' />}>
--- a/application/ui/src/features/inspect/models/models.component.tsx
+++ b/application/ui/src/features/inspect/models/models.component.tsx
@@ -41,7 +41,7 @@ interface ModelData {
     name: string;
     timestamp: string;
     startTime: number;
-    duration: number | null; // seconds
+    durationInSeconds: number | null;
     status: 'Training' | 'Completed' | 'Failed';
     architecture: string;
     progress: number;
@@ -62,12 +62,11 @@ export const ModelsView = () => {
             }
 
             let timestamp = '';
-            let duration = 0;
+            let durationInSeconds = 0;
             const start = job.start_time ? new Date(job.start_time) : new Date();
             if (job) {
                 const end = job.end_time ? new Date(job.end_time) : new Date();
-                // Job duration in seconds
-                duration = Math.floor((end.getTime() - start.getTime()) / 1000);
+                durationInSeconds = Math.floor((end.getTime() - start.getTime()) / 1000);
                 timestamp = dateFormatter.format(start);
             }
 
@@ -78,7 +77,7 @@ export const ModelsView = () => {
                 architecture: model.name!,
                 startTime: start.getTime(),
                 timestamp,
-                duration,
+                durationInSeconds,
                 progress: 1.0,
                 job,
             };
@@ -100,7 +99,7 @@ export const ModelsView = () => {
                 timestamp,
                 startTime: start.getTime(),
                 progress: 1.0,
-                duration: null,
+                durationInSeconds: null,
                 job,
             };
         });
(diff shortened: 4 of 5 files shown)


application/ui/src/features/inspect/dataset/dataset.component.tsx
```diff
@@ -47,9 +47,23 @@ const UploadImages = () => {
...
 
-        await queryClient.invalidateQueries({
-            queryKey: ['get', '/api/projects/{project_id}/images'],
+        const imagesOptions = $api.queryOptions('get', '/api/projects/{project_id}/images', {
+            params: { path: { project_id: projectId } },
         });
+        await queryClient.invalidateQueries({ queryKey: imagesOptions.queryKey });
+        const images = await queryClient.ensureQueryData(imagesOptions);
+
+        if (images.media.length >= 20) {
```
camiloHimura: Is 20 going to be the constant number used for training?

application/ui/src/features/inspect/jobs/show-job-logs.component.tsx
```diff
@@ -0,0 +1,129 @@
...
+                    try {
+                        const data = JSON.parse(message);
+                        if (data['text']) {
+                            yield data['text'];
+                        }
+                    } catch {
+                        console.error('Could not parse message:', message);
+                    }
+
+                    ({ promise, resolve, reject } = Promise.withResolvers<string>());
```
camiloHimura, line 62: Nice!!!

application/ui/src/features/inspect/jobs/show-job-logs.component.tsx
```diff
@@ -0,0 +1,129 @@
...
+    return (
+        <Dialog>
+            <Heading>Logs</Heading>
+            <Divider />
+            <Content>
+                <View
+                    padding='size-200'
+                    backgroundColor={'gray-50'}
+                    UNSAFE_style={{
+                        fontSize: '11px',
```
camiloHimura: Shouldn’t we use the Spectrum CSS variables instead?

application/ui/src/features/inspect/models/models.component.tsx
```diff
@@ -0,0 +1,203 @@
...
+                return null;
+            }
+
+            let timestamp = '';
+            let duration = 0;
+            const start = job.start_time ? new Date(job.start_time) : new Date();
+            if (job) {
+                const end = job.end_time ? new Date(job.end_time) : new Date();
+                // Job duration in seconds
+                duration = Math.floor((end.getTime() - start.getTime()) / 1000);
```
camiloHimura: Let’s rename it to `durationInSeconds` so the comment isn’t needed anymore

application/ui/src/features/inspect/models/models.component.tsx
```diff
@@ -0,0 +1,203 @@
...
+                timestamp = dateFormatter.format(start);
+            }
+
+            return {
+                id: model.id!,
```
camiloHimura, line 74: If possible, let’s avoid using the Non-Null Assertion Operator. When we’re sure about the value, we can just use String(model.id) instead

application/ui/src/features/inspect/train-model/trainable-model-list-box.component.tsx
```diff
@@ -0,0 +1,197 @@
...
+    return (
+        <div aria-label={`Attribute rating for ${name} is ${rating}`} style={{ height: '100%' }}>
+            <Flex direction={'column'} gap={'size-100'} justifyContent={'space-between'} height={'100%'}>
+                <Heading margin={0} UNSAFE_className={classes.attributeRatingTitle}>
+                    {name}
+                </Heading>
+                <Flex alignItems={'center'} gap={'size-100'}>
+                    {RateColors[rating].map((color, idx) => (
+                        <View
+                            key={idx}
```
camiloHimura: Declare `const rateLabels = Object.keys(RateColors);` outside the component

```suggestion
    key={`rate-${rateLabels[idx]}`}
```

application/ui/src/features/inspect/inference-provider.component.tsx
```diff
@@ -64,14 +65,25 @@ export const InferenceProvider = ({ children }: InferenceProviderProps) => {
     const [selectedModelId, setSelectedModelId] = useState<string | undefined>(undefined);
     const [inferenceOpacity, setInferenceOpacity] = useState<number>(0.75);
 
+    const { selectedMediaItem } = useSelectedMediaItem();
+
+    const onSetSelectedModelId = (modelId: string | undefined) => {
+        setSelectedModelId(modelId);
+
+        if (modelId) {
+            if (selectedMediaItem) {
```
camiloHimura: Any particular reason for having them in separate if statements instead of combining them into one?

MarkRedeman: Not anymore, will combine them. I think before I had some extra stuff in the first body but that's no longer needed.

application/ui/src/features/inspect/models/models.component.tsx
MarkRedeman: Yeah, the one thing I wanted to check as well is if we can make these ids non-nullable as that would save us a lot of effort. @maxxgx can you look into this?

application/ui/src/features/inspect/dataset/dataset.component.tsx
MarkRedeman: Yes, we already have a constant for it but I forgot to use it here, will update.

We might also change this to a lower number in the future.

application/ui/src/features/inspect/jobs/show-job-logs.component.tsx
MarkRedeman: First time using this method, felt like a nice use case. :D 
I did worry a bit that maybe this code is a bit too weird, but I expect we'd only need it here so I don't see too many issues with it.

application/backend/src/services/job_service.py
```diff
@@ -105,5 +110,8 @@ async def is_job_still_running():
                         continue
                     # No more lines are expected
                     else:
+                        yield "data: DONE\n\n"
                         break
-                yield line
+
+                # Format as an SSE message
+                yield f"data: {line.rstrip()}\n\n"
```
maxxgx, line 117: I think we can better encapsulate the log line into a `JSONServerSentEvent` object instead of string like this. FastAPI streaming is based on [sse-starlette](https://github.com/sysid/sse-starlette). But we don't have to address this in the current PR.

application/ui/src/features/inspect/models/models.component.tsx
maxxgx: I don't think I understand where nullable `model.id` is coming from? Where does the UI get the schema from?

MarkRedeman: We get this from the OpenAPI schema from the server. If you have the server running (on the default port) then you can run,
```
npm run build:api
```

from the `ui` folder and this will generate two files: a `openapi-spec.json` and `openapi-spec.d.ts`. The first is taken directly from the backend's OpenAPI json schema. The second is typescript types inferred based on that.

In this case the type inference happens on the models endpoint which returns a [ModelList](https://github.com/open-edge-platform/anomalib/blob/feature/geti-inspect/application/backend/src/pydantic_models/model.py#L57) which in turn returns the [Model](https://github.com/open-edge-platform/anomalib/blob/feature/geti-inspect/application/backend/src/pydantic_models/model.py#L20) pydantic model.
The nullable propery comes from [BaseIDModel's ID field](https://github.com/open-edge-platform/anomalib/blob/feature/geti-inspect/application/backend/src/pydantic_models/base.py#L14) which takes a `default_factory=uuid` which makes it nullable. Removing that default factory should solve it from the UI's point of view, but we might need to double check that it doesn't introduce regressions in the other models.

application/backend/src/services/job_service.py
MarkRedeman: Nice that sounds like a better solution. Let's try this out in a separate PR once this one is merged so we can easily test both backend and UI at the same time.

maxxgx: Made a PR https://github.com/open-edge-platform/anomalib/pull/3060. Will wait for this PR to be merged first

Issue #3061, opened 2025-10-28
🎯 [EPIC] Geti Inspect: Inference pipeline integration


MarkRedeman:
### Describe the epic

The inference pipeline of Geti Inspect allows the user to configure a source (example: a video stream from a camera), a trained model and a sink (example: send inference results to an MQTT message bus).
Users should be able to change the sources, sinks, models and inference pipeline status.

### Goals

Allow users to easily test anomaly detection pipelines.

### Initial inference pipeline

-   video<sub>file</sub> / webcam source
-   folder sink
-   inference
-   status
-   capture

### Select source

This issue keeps track of the progress of all individual source types.

1.  Disconnected (not needed, just set `pipeline.source` to `None`?)
2.  Webcam
    **NOTE**: Server side sync up with Arend Jan for FrameSource
    **NOTE**: Server side add a /sources/cameras/discover endpoint? (check with prompt / tune)
3.  IP camera
4.  Video file (server)
    **NOTE**: Video file and images folder can be combined into a single source.

5.  Images folder (server)
6.  Video file (client)
7.  Images folder (client)

### Select sink

This issue keeps track of the progress of all individual sink types.

1.  Disconnected (not needed, just set `pipeline.sink` to `None`?)
2.  Folder
3.  MQTT
4.  ROS
5.  Webhook

### Select model for inference

Update the pipeline&rsquo;s model.

PR #3078, opened 2025-10-30
feat(inspect): add, remove list and connect sources

camiloHimura:
## 📝 Description

This PR introduces the initial implementation of pipeline source integration. The new UI allows users to create, edit, and connect sources to a pipeline. Due to the number of files involved, this functionality will be split across multiple PRs. The upcoming one will focus on UX improvements

#3061 

https://github.com/user-attachments/assets/5f7840f5-d2b4-4105-9048-6a6911591785

- 🛠️ Fixes # (issue number)

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [ ] 🔒 Security update

## ✅ Checklist

### DIFF
--- a/application/ui/src/features/inspect/toolbar/sources/add-source/add-source.component.tsx
+++ b/application/ui/src/features/inspect/toolbar/sources/add-source/add-source.component.tsx
@@ -4,8 +4,8 @@
 import { ReactNode } from 'react';
 
 import { Button, Form } from '@geti/ui';
-import { useConnectSourceToPipeline } from 'src/hooks/use-pipeline.hook';
 
+import { useConnectSourceToPipeline } from '../../../../../hooks/use-pipeline.hook';
 import { useSourceAction } from '../hooks/use-source-action.hook';
 import { SourceConfig } from '../util';
 
--- a/application/ui/src/features/inspect/toolbar/sources/add-source/add-source.test.tsx
+++ b/application/ui/src/features/inspect/toolbar/sources/add-source/add-source.test.tsx
@@ -3,17 +3,17 @@
 
 import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { useConnectSourceToPipeline } from 'src/hooks/use-pipeline.hook';
 import { TestProviders } from 'src/providers';
 
+import { useConnectSourceToPipeline } from '../../../../../hooks/use-pipeline.hook';
 import { useSourceMutation } from '../hooks/use-source-mutation.hook';
 import { IpCameraFields } from '../ip-camera/ip-camera-fields.component';
-import { ipCameraBodyFormatter, ipCameraInitialConfig } from '../ip-camera/utils';
+import { getIpCameraInitialConfig, ipCameraBodyFormatter } from '../ip-camera/utils';
 import { IPCameraSourceConfig } from '../util';
 import { AddSource } from './add-source.component';
 
 vi.mock('../hooks/use-source-mutation.hook');
-vi.mock('src/hooks/use-pipeline.hook');
+vi.mock('../../../../../hooks/use-pipeline.hook');
 
 describe('add-source', () => {
     beforeEach(() => {
@@ -26,6 +26,7 @@ describe('add-source', () => {
         stream_url: 'rtsp://192.168.1.100:554/stream',
         source_type: 'ip_camera',
         auth_required: false,
+        project_id: '123',
     };
 
     it('calls connectToPipelineMutation after successful submit', async () => {
@@ -40,8 +41,8 @@ describe('add-source', () => {
             <TestProviders>
                 <AddSource
                     onSaved={mockOnSaved}
-                    config={ipCameraInitialConfig}
-                    componentFields={(state: IPCameraSourceConfig) => <IpCameraFields state={state} />}
+                    config={getIpCameraInitialConfig(newConfig.project_id)}
+                    componentFields={(state: IPCameraSourceConfig) => <IpCameraFields defaultState={state} />}
                     bodyFormatter={ipCameraBodyFormatter}
                 />
             </TestProviders>
--- a/application/ui/src/features/inspect/toolbar/sources/edit-source/edit-source.component.tsx
+++ b/application/ui/src/features/inspect/toolbar/sources/edit-source/edit-source.component.tsx
@@ -4,8 +4,8 @@
 import { ReactNode, useRef } from 'react';
 
 import { Button, ButtonGroup, Divider, Form, View } from '@geti/ui';
-import { useConnectSourceToPipeline } from 'src/hooks/use-pipeline.hook';
 
+import { useConnectSourceToPipeline } from '../../../../../hooks/use-pipeline.hook';
 import { useSourceAction } from '../hooks/use-source-action.hook';
 import { SourceConfig } from '../util';
 
--- a/application/ui/src/features/inspect/toolbar/sources/edit-source/edit-source.test.tsx
+++ b/application/ui/src/features/inspect/toolbar/sources/edit-source/edit-source.test.tsx
@@ -3,17 +3,17 @@
 
 import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { useConnectSourceToPipeline } from 'src/hooks/use-pipeline.hook';
 import { TestProviders } from 'src/providers';
 
+import { useConnectSourceToPipeline } from '../../../../../hooks/use-pipeline.hook';
 import { useSourceMutation } from '../hooks/use-source-mutation.hook';
 import { IpCameraFields } from '../ip-camera/ip-camera-fields.component';
-import { ipCameraBodyFormatter, ipCameraInitialConfig } from '../ip-camera/utils';
+import { getIpCameraInitialConfig, ipCameraBodyFormatter } from '../ip-camera/utils';
 import { IPCameraSourceConfig } from '../util';
 import { EditSource } from './edit-source.component';
 
 vi.mock('../hooks/use-source-mutation.hook');
-vi.mock('src/hooks/use-pipeline.hook');
+vi.mock('../../../../../hooks/use-pipeline.hook');
 
 describe('EditIpCamera', () => {
     beforeEach(() => {
@@ -23,6 +23,7 @@ describe('EditIpCamera', () => {
     const updatedConfig: IPCameraSourceConfig = {
         id: 'existing-source-id',
         name: 'Updated Camera',
+        project_id: '123',
         stream_url: 'rtsp://192.168.1.201:554/stream',
         source_type: 'ip_camera',
         auth_required: true,
@@ -32,9 +33,9 @@ describe('EditIpCamera', () => {
         render(
             <TestProviders>
                 <EditSource
-                    config={ipCameraInitialConfig}
+                    config={getIpCameraInitialConfig(updatedConfig.project_id)}
                     onSaved={mockOnSaved}
-                    componentFields={(state: IPCameraSourceConfig) => <IpCameraFields state={state} />}
+                    componentFields={(state: IPCameraSourceConfig) => <IpCameraFields defaultState={state} />}
                     bodyFormatter={ipCameraBodyFormatter}
                 />
             </TestProviders>
--- a/application/ui/src/features/inspect/toolbar/sources/hooks/use-source-mutation.hook.tsx
+++ b/application/ui/src/features/inspect/toolbar/sources/hooks/use-source-mutation.hook.tsx
@@ -27,17 +27,15 @@ export const useSourceMutation = (isNewSource: boolean) => {
 
     return async (body: SourceConfig) => {
         if (isNewSource) {
-            const sourcePayload = {
-                ...body,
-                id: uuid(),
-            };
+            const id = uuid();
+            const sourcePayload = { ...body, id };
 
-            const response = await addSource.mutateAsync({
+            await addSource.mutateAsync({
                 body: sourcePayload,
                 params: { path: { project_id: projectId } },
             });
 
-            return String(response.id);
+            return id;
         }
 
         const response = await updateSource.mutateAsync({
--- a/application/ui/src/features/inspect/toolbar/sources/image-folder/image-folder-fields.component.tsx
+++ b/application/ui/src/features/inspect/toolbar/sources/image-folder/image-folder-fields.component.tsx
@@ -9,22 +9,22 @@ import { ImagesFolderSourceConfig } from '../util';
 import classes from './image-folder-fields.module.scss';
 
 type ImageFolderFieldsProps = {
-    state: ImagesFolderSourceConfig;
+    defaultState: ImagesFolderSourceConfig;
 };
 
-export const ImageFolderFields = ({ state }: ImageFolderFieldsProps) => {
+export const ImageFolderFields = ({ defaultState }: ImageFolderFieldsProps) => {
     return (
         <Flex direction='column' gap='size-200'>
-            <TextField isHidden label='id' name='id' defaultValue={state.id} />
-            <TextField isHidden label='project_id' name='project_id' defaultValue={state.project_id} />
-            <TextField width={'100%'} label='Name' name='name' defaultValue={state.name} />
+            <TextField isHidden label='id' name='id' defaultValue={defaultState?.id} />
+            <TextField isHidden label='project_id' name='project_id' defaultValue={defaultState.project_id} />
+            <TextField width={'100%'} label='Name' name='name' defaultValue={defaultState.name} />
 
             <Flex direction='row' gap='size-200'>
                 <TextField
                     flex='1'
                     label='Images folder path'
                     name='images_folder_path'
-                    defaultValue={state.images_folder_path}
+                    defaultValue={defaultState.images_folder_path}
                 />
 
                 <Flex
@@ -41,8 +41,8 @@ export const ImageFolderFields = ({ state }: ImageFolderFieldsProps) => {
             <Switch
                 aria-label='ignore existing images'
                 name='ignore_existing_images'
-                defaultSelected={state.ignore_existing_images}
-                key={state.ignore_existing_images ? 'true' : 'false'}
+                defaultSelected={defaultState.ignore_existing_images}
+                key={defaultState.ignore_existing_images ? 'true' : 'false'}
             >
                 Ignore existing images
             </Switch>
--- a/application/ui/src/features/inspect/toolbar/sources/ip-camera/ip-camera-fields.component.tsx
+++ b/application/ui/src/features/inspect/toolbar/sources/ip-camera/ip-camera-fields.component.tsx
@@ -6,21 +6,21 @@ import { Flex, Switch, TextField } from '@geti/ui';
 import { IPCameraSourceConfig } from '../util';
 
 type IpCameraFieldsProps = {
-    state: IPCameraSourceConfig;
+    defaultState: IPCameraSourceConfig;
 };
 
-export const IpCameraFields = ({ state }: IpCameraFieldsProps) => {
+export const IpCameraFields = ({ defaultState }: IpCameraFieldsProps) => {
     return (
         <Flex direction='column' gap='size-200'>
-            <TextField isHidden label='id' name='id' defaultValue={state.id} />
-            <TextField isHidden label='project_id' name='project_id' defaultValue={state.project_id} />
-            <TextField width={'100%'} label='Name' name='name' defaultValue={state.name} />
-            <TextField width={'100%'} label='Stream Url:' name='stream_url' defaultValue={state.stream_url} />
+            <TextField isHidden label='id' name='id' defaultValue={defaultState?.id} />
+            <TextField isHidden label='project_id' name='project_id' defaultValue={defaultState.project_id} />
+            <TextField width={'100%'} label='Name' name='name' defaultValue={defaultState.name} />
+            <TextField width={'100%'} label='Stream Url:' name='stream_url' defaultValue={defaultState.stream_url} />
             <Switch
                 name='auth_required'
                 aria-label='Require Authentication'
-                defaultSelected={state.auth_required}
-                key={state.auth_required ? 'true' : 'false'}
+                defaultSelected={defaultState.auth_required}
+                key={defaultState.auth_required ? 'true' : 'false'}
             >
                 Require Authentication
             </Switch>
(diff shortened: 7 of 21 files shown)


application/ui/src/features/inspect/toolbar/sources/add-source/add-source.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+        await userEvent.clear(nameInput);
+        await userEvent.type(nameInput, newConfig.name);
+        await userEvent.clear(streamUrlInput);
+        await userEvent.type(streamUrlInput, newConfig.stream_url);
+        await userEvent.click(screen.getByRole('button', { name: /Add & Connect/i }));
+
+        await waitFor(() => {
+            expect(mockSourceMutation).toHaveBeenCalledWith(expect.objectContaining({ ...newConfig, id: '' }));
+            expect(mockConnectToPipeline).toHaveBeenCalledWith(newConfig);
```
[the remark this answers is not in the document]
camiloHimura: updated

application/ui/src/features/inspect/toolbar/sources/source-list/settings-list/settings-list.component.tsx
```diff
@@ -0,0 +1,45 @@
+import { SourceConfig } from '../../util';
+
+import classes from './settings-list.module.scss';
+
+interface SettingsListPops {
+    source: SourceConfig;
+}
+
+export const SettingsList = ({ source }: SettingsListPops) => {
```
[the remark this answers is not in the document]
camiloHimura: updated

application/ui/src/features/inspect/toolbar/sources/source-menu/source-menu.component.tsx
```diff
@@ -0,0 +1,85 @@
...
+                body: { source_id: null },
+            });
+        }
+
+        await removeSource.mutateAsync({ params: { path: { source_id: id } } });
+
+        toast({
+            type: 'success',
+            message: `${name} has been removed successfully!`,
+        });
```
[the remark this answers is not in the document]
camiloHimura: updated

application/ui/src/features/inspect/toolbar/sources/source-list/utils.ts
```diff
@@ -0,0 +1,3 @@
+export const removeUnderscore = (text: string) => {
```
ashwinvaidya17, line 1: Maybe nitpicky but is this needed? Maybe we can make it inline

application/ui/src/features/inspect/toolbar/sources/image-folder/image-folder-fields.component.tsx
```diff
@@ -0,0 +1,50 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Flex, Switch, TextField } from '@geti/ui';
+
+import { ReactComponent as Folder } from '../../../../../assets/icons/folder.svg';
```
MarkRedeman: Maybe update `assets/icons/index.tsx` to import it from there?

application/ui/src/assets/icons/index.ts
MarkRedeman, line 1: Note you only added `PipelineLink` here, not the `Folder`

application/ui/src/features/inspect/toolbar/sources/add-source/add-source.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+interface AddSourceProps<T> {
+    config: Awaited<T>;
+    onSaved: () => void;
+    componentFields: (state: Awaited<T>) => ReactNode;
+    bodyFormatter: (formData: FormData) => T;
+}
+
+export const AddSource = <T extends SourceConfig>({
+    config,
```
MarkRedeman, line 20: Why do we need to use generics here? Can we not type `SourceConfig` directly?

application/ui/src/features/inspect/toolbar/sources/util.ts
```diff
@@ -0,0 +1,41 @@
...
+export type SourceTypes = SourceConfig['source_type'];
+
+export const isOnlyDigits = (str: string): boolean => {
+    return /^\d+$/.test(str);
+};
+
+export const getImageFolderData = <T extends { source_type: string }>(sources: T[]) => {
+    return sources
+        .filter(({ source_type }) => source_type === 'images_folder')
+        .at(0) as unknown as ImagesFolderSourceConfig;
```
MarkRedeman: Generally: I'm wondering if we can't avoid this type assertion. Would it not be possible to do something like,
```
export const getImageFolderData = (sources: Array<SourceConfig>) => {
    return sources
        .find(({ source_type }): source is ImagesFolderSourceConfig => source_type === 'images_folder')
};
```

application/ui/src/features/inspect/toolbar/sources/util.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export const getIpCameraData = <T extends { source_type: string }>(sources: T[]) => {
+    return sources.filter(({ source_type }) => source_type === 'ip_camera').at(0) as unknown as IPCameraSourceConfig;
```
MarkRedeman: Note: each of these function do `.filter().at(0)`. In this case I think it's cleaner to do `.find`.

Though I'm wondering if this is correct: with the latest UX designs we should allow users to have multiple sources of the same type.
For instance I can have a folder source for `/data/cats` and another one for `/data/dogs`

application/ui/src/features/inspect/toolbar/sources/add-source/add-source.component.tsx
MarkRedeman: On second look, I think this will then constrain `AddSource` so that the `config`, `bodyFormatter` and `componentFields` are all compatible?
I think that is a good design so ignore my initial comment. (I'll leave it here just in case my assumption isn't correct)

The one thing that does confuse me though is `Awaited<T>`, why is that needed?

application/ui/src/features/inspect/toolbar/sources/edit-source/edit-source.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { TestProviders } from 'src/providers';
+
+import { useSourceMutation } from '../hooks/use-source-mutation.hook';
+import { IpCameraFields } from '../ip-camera/ip-camera-fields.component';
+import { ipCameraBodyFormatter, ipCameraInitialConfig } from '../ip-camera/utils';
+import { IPCameraSourceConfig } from '../util';
+import { EditSource } from './edit-source.component';
+
+vi.mock('../hooks/use-source-mutation.hook');
+vi.mock('src/hooks/use-pipeline.hook');
```
MarkRedeman: I'd prefer to use relative paths here, or at least be consistent among both imports

application/ui/src/features/inspect/toolbar/sources/hooks/use-source-mutation.hook.tsx
```diff
@@ -0,0 +1,41 @@
...
+        meta: {
+            invalidates: [['get', '/api/sources']],
+        },
+    });
+
+    return async (body: SourceConfig) => {
+        if (isNewSource) {
+            const sourcePayload = {
+                ...body,
+                id: uuid(),
```
MarkRedeman: I would consider adding the `id` to this function's arguments. (which reading below should already be possible via `body.id`?)
That way you can do optimistic updates outside the scope of this function.

Alternatively assign it to a local variable `const id = uuid()` and then return it here instead of parse it out of the response (`String(response.id)`).

application/ui/src/features/inspect/toolbar/sources/hooks/use-source-mutation.hook.tsx
```diff
@@ -0,0 +1,41 @@
...
+            };
+
+            const response = await addSource.mutateAsync({ body: sourcePayload });
+
+            return String(response.id);
+        }
+
+        const response = await updateSource.mutateAsync({
+            params: { path: { source_id: String(body.id) } },
+            body: omit(body, 'source_type'),
```
MarkRedeman, line 43: Why is `source_type` being omitted?

application/ui/src/features/inspect/toolbar/sources/ip-camera/ip-camera-fields.component.tsx
```diff
@@ -0,0 +1,28 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Flex, Switch, TextField } from '@geti/ui';
+
+import { IPCameraSourceConfig } from '../util';
+
+type IpCameraFieldsProps = {
+    state: IPCameraSourceConfig;
```
MarkRedeman: Perhaps call this `defaultState` to better imply that it is only being used for `defaultValue` (i.e. we use uncontrolled form elements and no controlled ones)

application/ui/src/features/inspect/toolbar/sources/source-list/source-icon/source-icon.component.tsx
```diff
@@ -0,0 +1,25 @@
+import { ImagesFolder, IpCamera, VideoFile, Webcam } from '@geti-inspect/icons';
+
+interface SourceIconProps {
+    type: string;
+}
+
+export const SourceIcon = ({ type }: SourceIconProps) => {
+    if (type === 'webcam') {
```
MarkRedeman, line 8: General: you could also use a `switch` in these type of cases.

I don't have a strong opinion on this though, so feel free to keep it as is.

application/ui/src/features/inspect/toolbar/sources/source-list/source-list.component.tsx
```diff
@@ -0,0 +1,71 @@
...
+    const currentSource = pipeline.data.source?.id;
+
+    return (
+        <Flex gap={'size-200'} direction={'column'}>
+            <Button variant='secondary' UNSAFE_className={classes.addSource} onPress={onAddSource}>
+                <AddIcon /> Add new source
+            </Button>
+
+            {sources.map((source) => {
+                const isConnected = isEqual(currentSource, source.id);
```
MarkRedeman: Possibly extract this into a (local) `SourceListItem` to make it easier for the react compiler to optimize this.

Specifically I think it could optimize the usage of `onEdit` here. (I'm also wondering if doing this refactor would allow you to simplify this a bit more - the name `-onEditSourceFactory` feels a bit odd to me.

application/ui/src/features/inspect/toolbar/sources/source-menu/source-menu.component.tsx
```diff
@@ -0,0 +1,97 @@
...
+            invalidates: [
+                ['get', '/api/sources'],
+                ['get', '/api/projects/{project_id}/pipeline', { params: { path: { project_id: projectId } } }],
+            ],
+        },
+    });
+
+    const removeSource = $api.useMutation('delete', '/api/sources/{source_id}', {
+        meta: { invalidates: [['get', '/api/sources']] },
+    });
```
MarkRedeman: Minor thing: code structure wise I would move these below `handleOnAction`. (i.e. declare variables as close to the place where they will be used)

This to me would make it a bit easier to understand where `updatePipeline` will be used. Since before I was wondering if it would also be used in the `onEdit` case.

application/ui/src/features/inspect/toolbar/sources/edit-source/edit-source.component.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+import { useSourceAction } from '../hooks/use-source-action.hook';
+import { SourceConfig } from '../util';
+
+import classes from './edit-source.module.scss';
+
+interface EditSourceProps<T> {
+    config: Awaited<T>;
+    onSaved: () => void;
+    componentFields: (state: Awaited<T>) => ReactNode;
```
MarkRedeman, line 17: I'm wondering if we can't make this a react node directly. I.e.

```
<EditSource
    config={folderSource}
    onSaved={onSaved}
    componentFields={<EditFolderSource state={state} />}
/>
```

Perhaps as well we could get rid of `EditSourceFactory` and use `EditSource` directly? (generally I'm not a huge fan of having components with a `Factory` suffix, I often feel like if we could find a design that avoids this then that would be better)

Though I guess in this case you don't have access to `state`? Maybe this could be shared via context?

application/ui/src/features/inspect/toolbar/sources/edit-source-factory.component.tsx
```diff
@@ -0,0 +1,67 @@
...
+    config: SourceConfig;
+    onSaved: () => void;
+}
+
+export const EditSourceFactory = ({ config, onSaved }: EditSourceFactoryProps) => {
+    if (config.source_type === 'webcam') {
+        return (
+            <EditSource
+                onSaved={onSaved}
+                config={getWebcamData([config])}
```
MarkRedeman: In all of these we always pass `[config]` as an array with a single element, why not pass `config` directly? (and then throw an error if the config is not what we expect - though do be sure to add an error boundary in that case)

application/ui/src/features/inspect/toolbar/sources/source-actions.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { $api } from '../../../../api/client';
+import { EditSourceFactory } from './edit-source-factory.component';
+import { SourcesList } from './source-list/source-list.component';
+import { SourceOptions } from './source-options.component';
+import { SourceConfig } from './util';
+
+export const SourceFactory = () => {
+    const sourcesQuery = $api.useSuspenseQuery('get', '/api/sources');
+    const sources = sourcesQuery.data ?? [];
+    const [view, setView] = useState<'list' | 'options' | 'edit'>(isEmpty(sources) ? 'options' : 'list');
```
MarkRedeman, line 22: Should `options` be part of this state? I.e. can it not always be derived from `sources` being empty?

application/ui/src/features/inspect/toolbar/sources/source-actions.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+        return (
+            <SourcesList
+                sources={sources}
+                onAddSource={handleAddSource}
+                onEditSourceFactory={handleEditSourceFactory}
+            />
+        );
+    }
+
+    return <SourceOptions onSaved={() => setView('list')} />;
```
MarkRedeman, line 42: I think this is mainly used as a way to immediately show a form for saving a new source when there's no other source, right? Maybe rename it to `NewSourceForm`?

application/ui/src/features/inspect/toolbar/sources/source-factory.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+        setView('options');
+    };
+
+    const handleEditSourceFactory = (source: SourceConfig) => {
+        setView('edit');
+        setCurrentSource(source);
+    };
+
+    if (view === 'edit' && !isEmpty(currentSource)) {
+        return <EditSourceFactory config={currentSource} onSaved={() => setView('list')} />;
```
MarkRedeman: Naming wise, maybe rename to `EditSourceForm`

MarkRedeman: Naming wise, maybe rename to `EditSourceForm`

application/ui/vitest.config.ts
```diff
@@ -1,9 +1,11 @@
 import react from '@vitejs/plugin-react';
 import svgr from 'vite-plugin-svgr';
+import tsconfigPaths from 'vite-tsconfig-paths';
 import { defineConfig } from 'vitest/config';
 
 export default defineConfig({
     plugins: [
+        tsconfigPaths(),
```
MarkRedeman, line 9: Is this because we have those path aliases that we use in some places?

Imho we should just remove them since it adds more complexity without a big benefit. Perhpas instead we can make use of `package.json`'s "imports" feature, which is natively supported by both node and typescript (i.e. doesn't need any extra plugins)

application/ui/src/providers.tsx
```diff
@@ -31,6 +32,7 @@ export const TestProviders = ({ children, routerProps }: { children: ReactNode;
                 <Router {...routerProps}>
                     <WebRTCConnectionProvider>{children}</WebRTCConnectionProvider>
                 </Router>
+                <Toast />
```
MarkRedeman:
```suggestion
                    <WebRTCConnectionProvider>{children}</WebRTCConnectionProvider>
                    <Toast />
                </Router>
```

Move the `Toast` inside of `Router` so that if a toast uses a custom component that relies on `useParams` (i.e. because it might use `useProjectIdentifier`) then it won't break.

application/ui/src/features/inspect/toolbar/sources/add-source/add-source.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { ReactNode } from 'react';
+
+import { Button, Form } from '@geti/ui';
+import { useConnectSourceToPipeline } from 'src/hooks/use-pipeline.hook';
+
+import { useSourceAction } from '../hooks/use-source-action.hook';
+import { SourceConfig } from '../util';
+
+interface AddSourceProps<T> {
+    config: Awaited<T>;
```
ActiveChooN: +1 to awaited question

application/ui/src/features/inspect/toolbar/sources/image-folder/image-folder-fields.component.tsx
camiloHimura: update to `import { Folder } from '@geti-inspect/icons'`

application/ui/src/features/inspect/toolbar/sources/add-source/add-source.component.tsx
camiloHimura: You’re right, the idea was to add a flexible constraint. As for the `Awaited<T>`, that’s the type returned by `useSourceAction`; it uses `useActionState` under the hood.

application/ui/src/features/inspect/toolbar/sources/util.ts
camiloHimura: it was part of my initial implementation, but it's not needed any more, removed

application/ui/src/features/inspect/toolbar/sources/util.ts
camiloHimura: removed

application/ui/src/features/inspect/toolbar/sources/edit-source/edit-source.test.tsx
camiloHimura: update to relative paths

application/ui/src/features/inspect/toolbar/sources/hooks/use-source-mutation.hook.tsx
camiloHimura: updated

application/ui/src/features/inspect/toolbar/sources/hooks/use-source-mutation.hook.tsx
camiloHimura: when sending the `source_type` even without changes the endpoint returns an error `The 'source_type' field cannot be changed`

application/ui/src/features/inspect/toolbar/sources/ip-camera/ip-camera-fields.component.tsx
camiloHimura: updated

application/ui/src/features/inspect/toolbar/sources/source-list/source-icon/source-icon.component.tsx
camiloHimura: I'll keep it for now an revisit after all `sink` changes get merged

application/ui/src/features/inspect/toolbar/sources/source-list/source-list.component.tsx
camiloHimura: agree, I'll rename it to `onEditSource`

application/ui/src/features/inspect/toolbar/sources/source-menu/source-menu.component.tsx
camiloHimura: updated

application/ui/src/features/inspect/toolbar/sources/edit-source/edit-source.component.tsx
camiloHimura: The main reason is to pass the new state managed by `useSourceAction`. I’d like to keep it this way for now since it feels more straightforward than using a context. I’ll take another look when I move on to `sinks`. I think I can reuse this code there, and maybe a context could simplify things

application/ui/src/features/inspect/toolbar/sources/edit-source-factory.component.tsx
camiloHimura: agree, updated

application/ui/src/features/inspect/toolbar/sources/source-actions.component.tsx
camiloHimura: Yeah, but users have the option to go to the edit view and then return to the list. In that scenario, we still need to preserve the options state

application/ui/src/features/inspect/toolbar/sources/source-actions.component.tsx
camiloHimura: It’s not visible yet (it’ll come in the next PR), but we’ll also display those options when adding a new element, along with a header to make them easier for users to recognize

application/ui/src/features/inspect/toolbar/sources/source-factory.component.tsx
camiloHimura: updated

camiloHimura: updated

application/ui/vitest.config.ts
camiloHimura: Added a comment to revisit later

application/ui/src/features/inspect/toolbar/sources/source-list/utils.ts
camiloHimura: It’s possible, but I’ll be reusing it in an upcoming PR, so it’s useful to keep it as a reusable function

application/ui/src/providers.tsx
camiloHimura: updated

MarkRedeman (approved):

PR #3122, opened 2025-11-17
feat(inspect): Fix backend job cancellation

ActiveChooN:
## 📝 Description

Improved training job cancellation handling

- 🛠️ Fixes # (issue number)

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [ ] 🔒 Security update

## ✅ Checklist

### DIFF
--- a/application/backend/src/services/training_service.py
+++ b/application/backend/src/services/training_service.py
@@ -239,6 +239,7 @@ class TrainingService:
             service = model_service or ModelService()
             await service.delete_model(project_id=job.project_id, model_id=model.id)
 
+    @staticmethod
     def _compute_export_size(path: str | None) -> int | None:
         if path is None:
             return None


application/backend/src/services/training_service.py
```diff
@@ -90,13 +94,28 @@ async def _run_training_job(cls, job: Job, job_service: JobService) -> Model:
...
+        except TrainingCancelledError:
+            logger.info("Training job `%s` cancelled by user", job.id)
+            await job_service.update_job_status(
+                job_id=job.id,
+                status=JobStatus.CANCELED,
+                message="Training cancelled by user",
+            )
+            if model.export_path:
+                model_binary_repo = ModelBinaryRepository(project_id=project_id, model_id=model.id)
+                await model_binary_repo.delete_model_folder()
```
maxxgx: This is almost the same as in L124::L128. You can refactor the cleanup code into a separate function, e.g. `def _cleanup_partial_model`

application/backend/src/services/training_service.py
```diff
@@ -21,6 +21,10 @@
 from utils.experiment_loggers import TrackioLogger
 
 
+class TrainingCancelledError(Exception):
+    """Raised when a training job is cancelled by the user."""
```
maxxgx: A job cancellation event should not be an exception since it's expected and valid event triggered by the user. Please remove this class

application/backend/src/services/training_service.py
```diff
@@ -90,13 +94,28 @@ async def _run_training_job(cls, job: Job, job_service: JobService) -> Model:
                 device=device,
                 synchronization_parameters=synchronization_parameters,
             )
+
+            if synchronization_parameters.cancel_training_event.is_set():
+                raise TrainingCancelledError
```
maxxgx: As mentioned above, do not use an exception for this as it's semantically wrong for this scenario.
You can instead call a "on_job_cancelled" function that handles the cancellation process.
After that you can return `None`.

application/backend/src/services/training_service.py
```diff
@@ -172,12 +191,23 @@ def _train_model(
...
 
         # Capture pytorch stdout logs into logger
-        with redirect_stdout(LoggerStdoutWriter()):  # type: ignore[type-var]
-            engine.fit(model=anomalib_model, datamodule=datamodule)
+        try:
+            with redirect_stdout(LoggerStdoutWriter()):  # type: ignore[type-var]
+                engine.fit(model=anomalib_model, datamodule=datamodule)
+        except Exception as exc:
+            if synchronization_parameters.cancel_training_event.is_set():
+                raise TrainingCancelledError from exc
```
maxxgx: As above, please remove raise and return None instead.

application/backend/src/services/training_service.py
```diff
@@ -172,12 +191,23 @@ def _train_model(
...
+            raise
+
+        if synchronization_parameters.cancel_training_event.is_set():
+            raise TrainingCancelledError
```
maxxgx: Unnecessary

ActiveChooN: Should we check for cancellations just before exporting the model? I mean, if the model training is cancelled right before it finishes, it can produce a job with a status of 'cancelled' and a model with a status of 'ready', which makes no sense.

application/backend/src/services/training_service.py
```diff
@@ -192,7 +181,7 @@ def _train_model(
         )
 
         if synchronization_parameters.cancel_training_event.is_set():
-            raise TrainingCancelledError
+            return None
```
maxxgx: Checking this event 3 times in the same function is too excessive. The trainer is aware of this event and will stop execution, so you only need to check once at the end before the export.

application/backend/src/services/training_service.py
```diff
@@ -203,18 +192,20 @@ def _train_model(
                 engine.fit(model=anomalib_model, datamodule=datamodule)
         except Exception as exc:
             if synchronization_parameters.cancel_training_event.is_set():
-                raise TrainingCancelledError from exc
+                logger.info("Training cancelled during execution: %s", exc)
+                return None
```
maxxgx: Please remove this exception guard and cancellation checking. If the trainer raises an error, the job should fail even if cancellation was triggered during at the same time (unlikely) of an error.

application/backend/src/services/training_service.py
```diff
@@ -203,18 +192,20 @@ def _train_model(
...
-
         # Find and set threshold metric
         for callback in engine.trainer.callbacks:  # type: ignore[attr-defined]
             if threshold := getattr(callback, "normalized_pixel_threshold", None):
                 logger.debug(f"Found pixel threshold set to: {threshold}")
                 model.threshold = threshold.item()
                 break
+
+        if synchronization_parameters.cancel_training_event.is_set():
+            return None
```
maxxgx, line 199: You can remove the above checks and keep this one.

application/backend/tests/unit/services/test_training_service.py
```diff
@@ -6,11 +6,10 @@
 
 import pytest
 
-from pydantic_models import JobStatus
-from repositories.binary_repo import ImageBinaryRepository, ModelBinaryRepository
+from pydantic_models import JobStatus  # type: ignore[import-not-found]
```
maxxgx: Why do you need the ignore flag? I ran ruff on the same file and it does not give any error without the ignore

ActiveChooN: Hmm, looks like that's some problems with my local Pylance setup. Please, disregard

PR #3133, opened 2025-11-20
feat(inspect): Capture frame

camiloHimura:
## 📝 Description

https://github.com/user-attachments/assets/54e40685-cba1-4227-be6f-52eb173294e5

- 🛠️ Fixes # (issue number)

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [ ] 🔒 Security update

## ✅ Checklist

### DIFF
--- a/application/ui/src/features/inspect/dataset/dataset-item/dataset-item.component.tsx
+++ b/application/ui/src/features/inspect/dataset/dataset-item/dataset-item.component.tsx
@@ -1,3 +1,5 @@
+import { useEffect } from 'react';
+
 import { Skeleton } from '@geti/ui';
 import { useQuery } from '@tanstack/react-query';
 import { clsx } from 'clsx';
@@ -27,11 +29,17 @@ export const DatasetItem = ({ mediaItem }: DatasetItemProps) => {
             if (!response.ok) {
                 throw new Error('Network response was not ok');
             }
-            return response.blob();
+            return URL.createObjectURL(await response.blob());
         },
         retry: 3,
-        retryDelay: (attemptIndex: number) => Math.min(1000 * 2 ** attemptIndex, 100000),
+        retryDelay: (attemptIndex: number) => Math.min(1000 * 2 ** attemptIndex, 10000),
     });
+
+    useEffect(() => {
+        return () => {
+            thumbnailBlob && URL.revokeObjectURL(thumbnailBlob);
+        };
+    }, [thumbnailBlob]);
 
     const handleClick = async () => {
         const selection = mediaItem.id === selectedMediaItem?.id ? undefined : mediaItem;
@@ -46,7 +54,7 @@ export const DatasetItem = ({ mediaItem }: DatasetItemProps) => {
                 <Skeleton width={'100%'} height={'100%'} />
             ) : (
                 <>
-                    <img src={URL.createObjectURL(thumbnailBlob)} alt={mediaItem.filename} />
+                    <img src={thumbnailBlob} alt={mediaItem.filename} />
                     <div className={clsx(styles.floatingContainer, styles.rightTopElement)}>
                         <DeleteMediaItem
                             itemsIds={[String(mediaItem.id)]}
--- a/application/ui/src/features/inspect/stream/util.tsx
+++ b/application/ui/src/features/inspect/stream/util.tsx
@@ -28,9 +28,9 @@ export const captureVideoFrame = (videoRef: RefObject<HTMLVideoElement | null>) 
             return;
         }
 
-        const file = new File([blob], 'frame.jpg', { type: blob.type });
+        const file = new File([blob], 'frame.jpg', { type: 'image/jpeg' });
         resolve(file);
-    }, 'image/jpg');
+    }, 'image/jpeg');
 
     return promise;
 };


application/ui/src/features/inspect/stream/util.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+    canvas.toBlob((blob) => {
+        if (!blob) {
+            reject('Failed to capture frame');
+            return;
+        }
+
+        const file = new File([blob], 'frame.jpg', { type: blob.type });
+        resolve(file);
+    }, 'image/jpg');
```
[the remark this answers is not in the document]
camiloHimura: updated

application/ui/src/features/inspect/dataset/dataset-item/dataset-item.component.tsx
```diff
@@ -42,25 +41,22 @@ const DatasetItem = ({ mediaItem }: DatasetItemProps) => {
...
-            })}
-            onClick={handleClick}
-        >
-            <img src={mediaUrl} alt={mediaItem.filename} />
+        <div className={clsx(styles.datasetItem, { [styles.datasetItemSelected]: isSelected })} onClick={handleClick}>
+            {isLoading || !thumbnailBlob ? (
+                <Skeleton width={'100%'} height={'100%'} />
+            ) : (
+                <>
+                    <img src={URL.createObjectURL(thumbnailBlob)} alt={mediaItem.filename} />
```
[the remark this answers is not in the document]
camiloHimura: updated

application/ui/src/features/inspect/dataset/dataset-item/dataset-item.component.tsx
```diff
@@ -1,38 +1,37 @@
...
+        queryFn: async () => {
+            const response = await fetch(`/api/projects/${mediaItem.project_id}/images/${mediaItem.id}/thumbnail`);
+
+            if (!response.ok) {
+                throw new Error('Network response was not ok');
+            }
+            return response.blob();
+        },
+        retry: 3,
+        retryDelay: (attemptIndex: number) => Math.min(1000 * 2 ** attemptIndex, 100000),
```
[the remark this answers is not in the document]
camiloHimura: updated

application/ui/src/features/inspect/stream/util.tsx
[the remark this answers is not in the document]
camiloHimura: updated

application/ui/src/features/inspect/dataset/dataset-item/dataset-item.component.tsx
```diff
@@ -1,38 +1,45 @@
...
     const { selectedMediaItem, onSetSelectedMediaItem } = useSelectedMediaItem();
     const { onInference, selectedModelId } = useInference();
 
     const isSelected = selectedMediaItem?.id === mediaItem.id;
 
-    const mediaUrl = `/api/projects/${mediaItem.project_id}/images/${mediaItem.id}/thumbnail`;
+    const { data: thumbnailBlob, isLoading } = useQuery({
+        queryKey: ['media', mediaItem.id],
+        queryFn: async () => {
+            const response = await fetch(`/api/projects/${mediaItem.project_id}/images/${mediaItem.id}/thumbnail`);
```
MarkRedeman, line 27: Why not use `$api.useQuery` or `fetchClient.fetch()` here?

application/ui/src/features/inspect/dataset/dataset-item/dataset-item.component.tsx
```diff
@@ -42,25 +49,22 @@ const DatasetItem = ({ mediaItem }: DatasetItemProps) => {
...
-            })}
-            onClick={handleClick}
-        >
-            <img src={mediaUrl} alt={mediaItem.filename} />
+        <div className={clsx(styles.datasetItem, { [styles.datasetItemSelected]: isSelected })} onClick={handleClick}>
+            {isLoading || !thumbnailBlob ? (
+                <Skeleton width={'100%'} height={'100%'} />
+            ) : (
+                <>
+                    <img src={thumbnailBlob} alt={mediaItem.filename} />
```
MarkRedeman, line 57: Would it not be possible to add error handling on the image element itself (and use the actual `/thumbnail` url?
Using a blob for an image that does have an actual endpoint that expects to return an image feels a bit redundant to me. (mainly because now we'd load the image twice in memory - not a huge deal if we don't have a lot of images but could be problematic long term)

application/ui/src/features/inspect/dataset/delete-dataset-item/alert-dialog-content.component.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { useEventListener } from '../../../../hooks/event-listener/event-listener.hook';
+
+type AlertDialogContentProps = {
+    itemsIds: string[];
+    onPrimaryAction: () => void;
+};
+
+export const AlertDialogContent = ({ itemsIds, onPrimaryAction }: AlertDialogContentProps) => {
+    useEventListener('keydown', (event) => {
+        if (event.key === 'Enter') {
```
MarkRedeman, line 12: Would it be possible to auto focus the confirmation button? That way we would be following "the platform" approach in that the enter button would trigger the confirmation button, while also keeping the option that the user can shfit-tab to focus the cancel button and press enter to cancel.

Not a huge blocker or anything, but would be nice if we can avoid this.

application/ui/src/features/inspect/dataset/delete-dataset-item/delete-dataset-item.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+                type: 'error',
+                message: `Failed to delete, ${error?.detail}`,
+            });
+        },
+    });
+
+    const handleRemoveItems = async () => {
+        alertDialogState.close();
+
+        toast({ id: 'deleting-notification', type: 'info', message: `Deleting items...` });
```
MarkRedeman, line 42: I don't think we'd really need a toast for this.

What do you think of changing the opacity of the item-to-be-deleted to 50% or so while a request is being made to delete it?
And when the delete request is done the image will be removed from the list (via the cache invalidation) so we don't necessarily need a success toast here as well.

To avoid having to prop drill the delete mutation or pass a `setIsDeleting` prop, perhaps we could render a `<div style={{ position: 'aboslute', inset: 0, backgroundColor: 'rgba(255, 255, 255, 0.5)` }} />` div that is positioned on top of the image?

application/ui/src/features/inspect/stream/stream.tsx
```diff
@@ -87,25 +95,73 @@ const useStreamToVideo = () => {
...
+
+    videoRef?.current.classList.add(classes.takeFlash);
+    videoRef?.current.classList.add(classes.takeOldCamera);
+};
+
+const removeAnimationClasses = (videoRef: RefObject<HTMLVideoElement | null>) => {
+    if (!videoRef.current) return;
+
+    videoRef?.current.classList.remove(classes.takeFlash);
+    videoRef?.current.classList.remove(classes.takeOldCamera);
```
MarkRedeman, line 109: Do we have to add / remove these classes via the ref? Why not add them declaratively in `<video className={clsx({ [takeFlas]: classes.takeFlash, [takeOldCamera]: classes.takeOldCamera })} `?

application/ui/src/features/inspect/dataset/dataset-item/dataset-item.component.tsx
camiloHimura: `$api` is configured to expect JSON responses, which always results in an error in this case. I looked into it and it would require a custom fetch with some conditions. After trying a few approaches and not getting it to work, I went with the simpler solution.

application/ui/src/features/inspect/stream/stream.tsx
camiloHimura: It was originally taken from `Geti Classic`, where we didn’t have direct access to the video tag, so using refs was the only option. I’ll update it in my next PR.

PR #3136, opened 2025-11-21
📦 feat(inspect): Add docker build

ashwinvaidya17:
## 📝 Description

- [x] Add CPU build
- [x] Add GPU build
- [x] Add XPU build
- Add XPU related changes from https://github.com/open-edge-platform/anomalib/pull/3210
- Remove trackio as it does not play well with multi-threading when Evaluator is present (needed for early stopping)

## ✨ Changes

- [ ] 📚 Documentation update
- [x] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [ ] 🔒 Security update

## ✅ Checklist

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -5764,7 +5764,7 @@ wheels = [
 
 [[package]]
 name = "onnx-ir"
-version = "0.1.12"
+version = "0.1.14"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "ml-dtypes" },
@@ -5772,14 +5772,14 @@ dependencies = [
     { name = "onnx" },
     { name = "typing-extensions" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/6c/1a/2a94112a39d01a9d1490f5ef3c205d8a17fe1ca27f307b026c40d62d8e9f/onnx_ir-0.1.12.tar.gz", hash = "sha256:742e0bff875d0547724187560b3f441833191c8aa939c05f14176f4892784deb", size = 112699, upload-time = "2025-10-28T23:43:54.129Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/a5/5b/ebd083a5c3d25ce9f95b34a11b3a492cdcf7831bf127c0f64429a4e83961/onnx_ir-0.1.14.tar.gz", hash = "sha256:bd69e3b5821046d5d7c9d0fdd023f8e1d0cc9a62cbee986fa0e5ab2b1602d7ae", size = 120732, upload-time = "2026-01-07T01:19:47.777Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/c8/36/c4df116f5dcaa82ec7944e5d25624a3811f6603fd190660b0b079ea759fb/onnx_ir-0.1.12-py3-none-any.whl", hash = "sha256:17f86faf8a53b979430bde1bc6022c7a162b0d1534550ddb17a1d37eb993e765", size = 129277, upload-time = "2025-10-28T23:43:52.493Z" },
+    { url = "https://files.pythonhosted.org/packages/53/d1/bd9a5007448b4599a80143b0b5ccc78e9c46176e5e1bee81f6d3da68d217/onnx_ir-0.1.14-py3-none-any.whl", hash = "sha256:89b212fa7840981c5db5dc478190f1b7369536297c3c6eae68fb1c2237dd2554", size = 139128, upload-time = "2026-01-07T01:19:46.403Z" },
 ]
 
 [[package]]
 name = "onnxscript"
-version = "0.5.6"
+version = "0.5.7"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "ml-dtypes" },
@@ -5789,9 +5789,9 @@ dependencies = [
     { name = "packaging" },
     { name = "typing-extensions" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/fb/4b/eed2199327bbf12c3443d7835893e3c4c23b1c1a4aa13efe0f7fbe0a6bf9/onnxscript-0.5.6.tar.gz", hash = "sha256:cc3338b2976daffd2af0bb6ac4866a4dca76aefface1666a0d7bc65ad9850822", size = 587017, upload-time = "2025-10-31T03:50:38.656Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/2c/f8/358a7d982ea51bc1b0c264f29c08adf096c62ba9f258ba13c954b41c46f5/onnxscript-0.5.7.tar.gz", hash = "sha256:480d572451bc233ed7f742b5005cb0c899594b2fdc28e15167dab26f7fd777ad", size = 596306, upload-time = "2025-12-16T20:47:15.762Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/fd/1e/a5462bfe28a2add00dc0abec7dd9b742ac3207b73e5c97bde9747b503971/onnxscript-0.5.6-py3-none-any.whl", hash = "sha256:b0c3355fea3eecab8ca291da8b77afddcaacd3ada5ee59294390a049ea123938", size = 683045, upload-time = "2025-10-31T03:50:41.15Z" },
+    { url = "https://files.pythonhosted.org/packages/f6/ec/1656ea93be1e50baf429c20603dce249fa3571f3a180407cee79b1afa013/onnxscript-0.5.7-py3-none-any.whl", hash = "sha256:f94a66059c56d13b44908e9b7fd9dae4b4faa6681c784f3fd4c29cfa863e454e", size = 693353, upload-time = "2025-12-16T20:47:17.897Z" },
 ]
 
 [[package]]
--- a/application/backend/src/pydantic_models/pipeline.py
+++ b/application/backend/src/pydantic_models/pipeline.py
@@ -88,9 +88,9 @@ class Pipeline(BaseModel):
         if self.is_running:
             self.status = PipelineStatus.RUNNING
         if self.status == PipelineStatus.RUNNING and any(
-            x is None for x in (self.source_id or self.source, self.sink_id or self.sink, self.model_id or self.model)
+            x is None for x in (self.source_id or self.source, self.model_id or self.model)
         ):
-            raise ValueError("Pipeline cannot be in 'running' status when source, sink, or model is not configured.")
+            raise ValueError("Pipeline cannot be in 'running' status when source, or model is not configured.")
         return self
 
     @cached_property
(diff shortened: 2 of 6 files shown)


application/backend/src/main.py
```diff
@@ -36,11 +36,7 @@
 # Enable CORS for local test UI
 app.add_middleware(
     CORSMiddleware,
-    allow_origins=[
-        "http://localhost:3000",
-        "http://localhost:9000",
-        "http://127.0.0.1:9000",
-    ],
+    allow_origins=["*"],
```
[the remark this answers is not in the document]
ashwinvaidya17: Matching https://github.com/open-edge-platform/training_extensions/blob/4d9a2b88f82680f869fe79fc3a85b5b0ec0f38eb/application/backend/app/main.py#L42

application/.packaging/docker/Dockerfile
```diff
@@ -0,0 +1,154 @@
...
+
+#------------------------------------------
+# Geti Inspect CPU version
+#------------------------------------------
+FROM geti-inspect-base AS geti-inspect-cpu
+USER non-root
+WORKDIR /app
+
+RUN --mount=type=cache,target=/home/non-root/.cache/uv,uid=10001,gid=10001 \
+    uv sync --frozen --no-dev --extra cpu --extra application
```
maxxgx: Why run uv sync from the root anomalib folder instead of application?

Installing inspect as an extra might be useful during development but I don't see the value in production environments

application/backend/src/api/endpoints/media_endpoints.py
maxxgx, line 1: Which pre-commit config did you run to result in all these style fixes? Using prek, ruff does not apply these changes.

application/backend/pyproject.toml
```diff
@@ -23,6 +27,7 @@ dependencies = [
     "uvloop==0.21.0",
     "loguru==0.7.3",
     "trackio~=0.6.0",
+    "sse-starlette",
```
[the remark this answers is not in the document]
maxxgx: Agree, but should pin a higher version. Eg. `"sse-starlette~=3.0"`

application/backend/pyproject.toml
```diff
@@ -50,9 +55,19 @@ mqtt = [
...
+
 [tool.uv.sources]
 anomalib = { path = "../../", editable = true }
 
+[tool.hatch.build.targets.wheel]
+packages = ["src"]
+only-include = ["src"]
+
+[tool.hatch.build.targets.wheel.sources]
+"src" = ""
```
maxxgx: What does `"src" = ""` do?

application/backend/src/main.py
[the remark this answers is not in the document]
warrkan: I'd rather implement it via env variable: https://github.com/open-edge-platform/geti-prompt/blob/main/application/backend/app/main.py#L73

application/backend/src/settings.py
```diff
@@ -36,8 +39,8 @@ class Settings(BaseSettings):
     db_echo: bool = Field(default=False, alias="DB_ECHO")
 
     # Alembic
-    alembic_config_path: str = "src/alembic.ini"
-    alembic_script_location: str = "src/alembic"
+    alembic_config_path: str = str(_MODULE_DIR / "alembic.ini")
```
warrkan, line 43: Should be adjusted further when we introduce Pyinstaller

application/docker/nginx.conf
```diff
@@ -0,0 +1,52 @@
+# PID file in a location non-root user can write to
```
warrkan: Can we omit having nginx by serving UI static using FastAPI?

application/docker/docker-compose.yml
maxxgx, line 1: Just a question: why put everything in a parent folder `.packaging/`?

application/.packaging/docker/Dockerfile
```diff
@@ -0,0 +1,154 @@
...
+FROM geti-inspect-base AS geti-inspect-cpu
+USER non-root
+WORKDIR /app/application/backend
+
+RUN --mount=type=cache,target=/home/non-root/.cache/uv,uid=10001,gid=10001 \
+    uv sync --frozen --no-dev --extra cpu
+
+EXPOSE 80
+
+CMD ["sh", "-c", "nginx && exec uv run src/main.py"]
```
maxxgx: Should we use `backend/run.sh`?

application/docker/README.md
maxxgx, line 1: Could you add example of running the container with camera passthrough? I guess would be a common usecase

application/docker/Dockerfile
maxxgx, line 1: I've tried to build it and it works. However, couldn't get camera passthrough to work on mac. Have you tested it on linux?

application/docker/docker-compose.yml
MarkRedeman: Could we move it into `application/docker` (without `.packaging`?) no need to hide it imho. This would also align well with the other applications.

application/backend/src/api/endpoints/webui_endpoints.py
```diff
@@ -0,0 +1,20 @@
...
+from fastapi import APIRouter, HTTPException
+from fastapi.responses import FileResponse
+
+from settings import get_settings
+
+settings = get_settings()
+webui_router = APIRouter(tags=["Webui"])
+
+
+@webui_router.get("/", include_in_schema=False)
```
MarkRedeman, line 15:
```suggestion
@webui_router.get("/", include_in_schema=False)
@webui_router.get("/{full_path:path}", include_in_schema=False)
```

With the previous implementation if the user is on a route (say `/projects`) and then refresh the browser page then they would see a 404 page. 
This is because the UI uses client side routing, so what we need the server to do is serve our `index.html` file for _any_ page that the user might access.

application/docker/docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+      dockerfile: application/docker/Dockerfile
+      target: geti-inspect-${AI_DEVICE:-cpu}
+      <<: *proxies
+    working_dir: /app/application/backend
+    container_name: geti-inspect-${AI_DEVICE:-cpu}
+    # Uncomment the following lines to enable torch XPU support
+    # For more info, see README.md
+    group_add:
+    #   - "${RENDER_GROUP_ID}" # This is needed to allow access to the 'render' group for torch XPU support
+      - "${VIDEO_GROUP_ID}"  # This is needed to allow access to video devices (webcams)
```
MarkRedeman: I didn't need this setting while testing on a lunar lake laptop. Though I guess this could differ based on how docker was installed?

application/docker/docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+  args: *proxy-list
+
+services:
+  geti-inspect:
+    image: geti-inspect-${AI_DEVICE:-cpu}:${TAG:-latest}
+    environment: *proxy-list
+    restart: unless-stopped
+    build:
+      # Use root directory as context to build the image
+      context: ../..
```
MarkRedeman: Should this not be `../`? (i.e. the application being the context excluding any library files?)

application/docker/Dockerfile
```diff
@@ -0,0 +1,146 @@
...
+COPY --link application/ui/packages ./packages
+RUN npm ci --audit=false --ignore-scripts
+
+
+COPY --link application/ui/tsconfig.json ./tsconfig.json
+COPY --link application/ui/rsbuild.config.ts ./rsbuild.config.ts
+COPY --link application/ui/src ./src
+
+ARG PUBLIC_API_BASE_URL=""
+ENV PUBLIC_API_BASE_URL=${PUBLIC_API_BASE_URL}
```
MarkRedeman: I think we could omit this, or set `PUBLIC_API_BASE_URL=""`.
When deploying Inspect the UI and backend should be expected to be served on the same hosts, this allows us to avoid security policies like CORS and (later?) have a better CSP configuration.

application/docker/Dockerfile
```diff
@@ -0,0 +1,146 @@
+#------------------------------------------
+# Web UI Build Stage
+#------------------------------------------
+FROM node:24-alpine3.22@sha256:3e843c608bb5232f39ecb2b25e41214b958b0795914707374c8acc28487dea17 AS web-ui-base
+
+WORKDIR /home/app/web_ui/
+
+COPY --link application/ui/package.json ./package.json
+COPY --link application/ui/package-lock.json ./package-lock.json
+COPY --link application/ui/packages ./packages
```
MarkRedeman: I think this should be omitted. The npm install script should clone and install the geti packages with the `npm ci` command.
If we copy from the host then we risk the docker image containing outdated packages (or at least packages that are not in sync with what's in the anomalib repo).

application/docker/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+## To create XPU build
+
+> [!NOTE]
+> You need to first set the `RENDER_GROUP_ID` environment variable to match your host system's render group ID.
+
+```bash
+cd application/docker
+RENDER_GROUP_ID=$(getent group render | cut -d: -f3) AI_DEVICE=xpu docker compose up
```
MarkRedeman: For me just running,
```
AI_DEVICE=xpu docker compose up
```

was working. (same for build command)

application/docker/README.md
```diff
@@ -0,0 +1,28 @@
+# Docker distribution for Geti Inspect
+
+## To create CPU build
```
MarkRedeman, line 3: The docs here mention creating builds, but the commands show `up` being used, is that intended?

application/docker/README.md
MarkRedeman, line 1: Perhaps one other thing we could look into is having separate `docker-compose.xpu.yaml`, `docker-compsoe.nvidia.yaml` files and tell the user to use these (instead of them uncommenting stuff).

See for instance https://github.com/open-webui/open-webui as an example of a repo using this.

.dockerignore
MarkRedeman, line 1: I would move this file into the `application` folder, since I think we shouldn't expect users to use docker for the library?

ashwinvaidya17: Since the context is the root of anomalib, it needs .dockerignore in the same folder

uv.lock
MarkRedeman, line 1: Why did this lockfile change? I do'nt see changes related to anomalib's pyproject file

.github/workflows/geti-inspect.yaml
```diff
@@ -48,7 +48,7 @@ jobs:
         working-directory: application/backend
         run: |
           uv lock --check
-          uv sync --frozen --all-extras
+          uv sync --frozen --extra cpu --extra mqtt
```
warrkan, line 51: Shouldn't we introduce `--extra cpu` as an input parameter?

application/backend/src/main.py
```diff
@@ -39,11 +42,7 @@
 # Enable CORS for local test UI
 app.add_middleware(
     CORSMiddleware,
-    allow_origins=[
-        "http://localhost:3000",
-        "http://localhost:9000",
-        "http://127.0.0.1:9000",
-    ],
+    allow_origins=["http://localhost:3000", "http://localhost:9000", "http://127.0.0.1:9000"],
```
warrkan, line 45: We need to make it configurable via env variables

application/backend/src/api/endpoints/webui_endpoints.py
```diff
@@ -0,0 +1,21 @@
...
```
warrkan, line 15: Isn't it worth it to put it directly into `main.py` module? There we also have a check for `settings.static_files_dir`

PR #3145, opened 2025-11-25
feat(inspect): use dataset snapshot for training

maxxgx:
## 📝 Description

Training with dataset snapshot with Parquet

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [ ] 🔒 Security update

## ✅ Checklist

### DIFF
--- a/application/backend/src/services/training_service.py
+++ b/application/backend/src/services/training_service.py
@@ -72,6 +72,8 @@ class TrainingService:
             raise ValueError(f"Job {job.id} payload must contain 'model_name'")
 
         synchronization_task: asyncio.Task[None] | None = None
+        model: Model | None = None  # Initialize model to None
+
         try:
             model_service = ModelService()
             snapshot = await DatasetSnapshotService.get_or_create_snapshot(
@@ -119,7 +121,7 @@ class TrainingService:
             await job_service.update_job_status(
                 job_id=job.id, status=JobStatus.FAILED, message=f"Failed with exception: {str(e)}"
             )
-            if model.export_path:
+            if model and model.export_path:
                 logger.warning(f"Deleting partially created model with id: {model.id}")
                 await model_service.delete_model(project_id=project_id, model_id=model.id)
             raise e
--- a/application/backend/tests/unit/services/test_training_service.py
+++ b/application/backend/tests/unit/services/test_training_service.py
@@ -31,6 +31,7 @@ def fxt_mock_job_service():
     mock_job_service = MagicMock()
     mock_job_service.get_pending_train_job = AsyncMock()
     mock_job_service.update_job_status = AsyncMock()
+    mock_job_service.get_job_by_id = AsyncMock()
     return mock_job_service
 
 
@@ -107,6 +108,8 @@ def fxt_mock_dataset_snapshot_service():
         mock_snapshot = MagicMock()
         mock_snapshot.id = uuid4()
         mock_service.create_snapshot = AsyncMock(return_value=mock_snapshot)
+        # Fix: get_or_create_snapshot needs to be awaited
+        mock_service.get_or_create_snapshot = AsyncMock(return_value=mock_snapshot)
         mock_service.delete_snapshot_if_unused = AsyncMock()
 
         # Setup context manager for use_snapshot_as_folder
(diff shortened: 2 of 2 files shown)


rajeshgangireddy (approved):

application/backend/src/services/dataset_snapshot_service.py
```diff
@@ -0,0 +1,200 @@
...
+        """Create a new immutable dataset snapshot from current media."""
+        snapshot_id = uuid4()
+        filename = f"{snapshot_id}.parquet"
+
+        logger.info(f"Creating dataset snapshot {snapshot_id} for project {project_id}")
+
+        async with get_async_db_session_ctx() as session:
+            # 1. Fetch all media
+            media_repo = MediaRepository(session, project_id=project_id)
+            all_media: list[Media] = await media_repo.get_all()
```
MarkRedeman: Would it be possible to design this so it would work with a iterator or generator?
I.e. making it so that we won't possibly load a lot of media in memory at the same time, and instead take a paginated approach?

application/backend/src/services/model_service.py
```diff
@@ -84,16 +85,19 @@ async def get_model_by_id(project_id: UUID, model_id: UUID) -> Model | None:
...
+        if not model:
+            raise ResourceNotFoundError(resource_id=str(model_id), resource_type=ResourceType.MODEL)
+
+        model_binary_repo = ModelBinaryRepository(project_id=project_id, model_id=model_id)
+        try:
+            await model_binary_repo.delete_model_folder()
+        except FileNotFoundError:
+            logger.warning(f"Model artifacts already absent on disk for model {model_id} in project {project_id}")
+        ds_snapshot_id = model.dataset_snapshot_id
+        await DatasetSnapshotService.delete_snapshot_if_unused(snapshot_id=ds_snapshot_id, project_id=project_id)
```
MarkRedeman, line 100: Maybe for later, we could consider a more message oriented design where the model service dispatches a `model_was_removed` event to an event bus.
Then we could have a `DatasetSnapshotSync` object/worker listen to that specific event (and maybe more later) that would delete the snapshot.

In my experience this type of message oriented design can make it easier to keep services small. Another thing that can help is to very explicitly define all events that might happen in the app by defining them via an `EVENT_TYPE` enum (`MODEL_WAS_TRAINED`, `MODEL_WAS_REMOVED`, `MODEL_WAS_ACTIVATED`, `INFERENCE_PIPELINE_CRASHED` etc), I guess like the kafka topics we use in Geti? But it is an extra layer of abstraction so might not be worth it yet to add it now.

application/backend/src/services/training_service.py
```diff
@@ -64,17 +65,30 @@ async def _run_training_job(cls, job: Job, job_service: JobService) -> Model | N
...
 
         model_service = ModelService()
+
+        # 1. Snapshot Resolution
+        if snapshot_id:
+            logger.info(f"Using existing snapshot {snapshot_id} for training")
+            snapshot_id = UUID(str(snapshot_id))
+        else:
+            logger.info("No snapshot provided, creating new snapshot")
+            snapshot = await DatasetSnapshotService.create_snapshot(project_id=project_id)
```
MarkRedeman: Qq: would this also create a new snapshot if you train two models successively? I.e. you first train PADIM and then Patchcore without changing the datasets, would that produce two snapshots or one?

MarkRedeman (approved):

application/backend/src/services/training_service.py
maxxgx: It would create two snapshots, we don't have a mechanism for determining if a dataset is unchanged since last snapshot. I think we'd need to keep track of a `last_modified_timestamp` for the dataset in order and use an existing snapshot which was created after that timestamp. I think we could possibly add this timestamp in the project table.
Having an event bus would probably help here as well.

~At the moment, a snapshot is only reused if an ID is passed to the job payload.~

application/backend/src/services/dataset_snapshot_service.py
maxxgx: I changed the implementation to be compatible with the generator. Since `get_all` is implemented in the base repository class, I didn't modify it in this PR. But I added a TODO comment as a reminder to address it later.

application/backend/src/services/model_service.py
maxxgx: Yeah that'd be nice to have. I'm personally looking into ØMQ, I've seen a few projects using it even for dispatching frames

application/backend/src/services/training_service.py
maxxgx: Implemented the above timestamp based snapshot creation condition.

New snapshot will be created only if the dataset has been modified (upload or delete media) since last snapshot

PR #3201, opened 2025-12-08
refactor(inspect): Improve pipeline visualization and add max_epochs

maxxgx:
## 📝 Description

- added `max_epochs` to job payload
- added option to turn anomaly map overlay on/off via PATCH pipeline

https://github.com/user-attachments/assets/00a869c3-a77b-49e2-8f93-f1249cb42771

### Overlay ON/OFF

https://github.com/user-attachments/assets/4cae1032-a00f-49e4-9055-83df3240a167

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [ ] 🔒 Security update

## ✅ Checklist

### DIFF
--- a/application/ui/src/features/inspect/toolbar/pipeline-switch/pipeline-switch.component.tsx
+++ b/application/ui/src/features/inspect/toolbar/pipeline-switch/pipeline-switch.component.tsx
@@ -23,7 +23,7 @@ export const PipelineSwitch = () => {
                 isSelected={hasOverlay}
                 isDisabled={patchPipeline.isPending || isPipelineStopped}
             >
-                Enabled
+                Overlay
             </Switch>
         </Flex>
     );


application/backend/src/services/model_service.py
```diff
@@ -264,6 +265,7 @@ async def predict_image(
             image_bytes: Raw image bytes from uploaded file
             cached_models: Optional dict to cache loaded models (for performance)
             device: Optional string indicating the device to use for inference
+            is_bgr: Whether the image is in BGR format
```
ashwinvaidya17: Maybe we should stick to a single format and require each service to only return in that format?

application/backend/src/pydantic_models/job.py
```diff
@@ -64,3 +64,4 @@ class TrainJobPayload(BaseModel):
     model_name: str
     device: str | None = Field(default=None)
     dataset_snapshot_id: str | None = Field(default=None)  # used because UUID is not JSON serializable
+    max_epochs: int | None = Field(default=None, ge=1)
```
ashwinvaidya17, line 67: How are we planning on dealing with rest of the configurable parameters?

application/backend/src/utils/visualization.py
```diff
@@ -46,7 +46,7 @@ def overlay_predictions(
     def overlay_anomaly_heatmap(
```
ashwinvaidya17, line 46: Something for the future but maybe we should consolidate the visualization methods here and the ones in anomalib. This way we won't have to maintain two separate visualizers

application/backend/src/pydantic_models/job.py
maxxgx: We don't for the time being. We're likely going to remove this once we have a proper design for configurable parameters.

application/backend/src/services/model_service.py
maxxgx: Fixed, I'm converting to RGB from the opencv stream

application/backend/src/services/training_service.py
```diff
@@ -105,6 +106,7 @@ async def _run_training_job(cls, job: Job, job_service: JobService) -> Model | N
                     model=model,
                     device=device,
                     synchronization_parameters=synchronization_parameters,
+                    max_epochs=max_epochs,
```
ActiveChooN, line 109: Should we add this to the model training dialog? cc @ashwinvaidya17, @MarkRedeman

MarkRedeman: It would be nice if we can already expose this (perhaps next to the picker of the training device from the new designs)

However there are some edge cases we would need to deal with. Some models will overwrite the max epochs to be 1 since they only need to do a single pass.
We could hardcode the UI to not show the input for those models, but that might get messy.

ashwinvaidya17: Let's expose this and just have a note saying something like, "some models only extract features so max_epochs will be overridden for those models".

PR #3210, opened 2025-12-11
fix(inspect): Enable training on XPU using SingleXPUStrategy

MarkRedeman:
Setting the SingleXPUStrategy fixes the error message,
```
Failed to train pending training job: Device should be xpu, got cpu
instead
```

After this another error would show up,
```
Cannot re-initialize XPU in forked subproces. To use XPU with
multiprocessing you must use the `spawn` method
```

This requires us to set the start method of our `main.py` file.

---

I've only verified these fixes on a lunar lake device on which I could train with both its CPU and iGPU.

### DIFF
--- a/application/backend/src/main.py
+++ b/application/backend/src/main.py
@@ -2,10 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import multiprocessing as mp
-
-if mp.get_start_method(allow_none=True) != "spawn":
-    mp.set_start_method("spawn", force=True)
-
 import os
 
 import uvicorn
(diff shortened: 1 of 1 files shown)


ashwinvaidya17 (approved):

application/backend/src/main.py
```diff
@@ -1,6 +1,11 @@
 # Copyright (C) 2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
+import multiprocessing as mp
+
+if mp.get_start_method(allow_none=True) != "spawn":
+    mp.set_start_method("spawn", force=True)
```
maxxgx: According to the docs, `set_start_method` should be called inside of the `__main__` guard: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method

If called from a sub process, it might result in a run time error.

PR #3224, opened 2025-12-16
fix(inspect): Remove alpha channel before inference

MarkRedeman:
## 📝 Description

When running inference on png images the `:predict` endpoint would return a 500 internal server error. The server logs include the following message,
```
Can't set the input tensor with index: 0, because the model input (shape=[?,3,?,?]) and the tensor (shape=(1.1085.1905.4)) are incompatible
```

here my image was a 1905 x 1085 png image.
This error gets fixed by removing the alpha channel from the image.

To me ideally this should be done in Anomalib's preprocessors. However I'm not sure where to do this.

## ✨ Changes

### DIFF
--- a/application/backend/src/services/model_service.py
+++ b/application/backend/src/services/model_service.py
@@ -335,6 +335,10 @@ class ModelService:
         if numpy_image is None:
             raise ValueError("Failed to decode image")
 
+        # Remove alpha channel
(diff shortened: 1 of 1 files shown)


ashwinvaidya17 (approved):
Thanks for the fix

PR #3322, opened 2026-02-06
fix(inspect): fix inference devices

maxxgx:
## 📝 Description

- Inference devices differ from training devices (torch vs openvino)
- OV do not support cuda, therefore we need to exclude nvidia gpus from inference
- Added experimental support for NPU (not yet tested)
- Use relative path in the UI

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [ ] 🔒 Security update

## ✅ Checklist

### DIFF
--- a/application/ui/src/api/utils.ts
+++ b/application/ui/src/api/utils.ts
@@ -6,9 +6,9 @@ import spec from './openapi-spec.json' with { type: 'json' };
 
 const handlers = await fromOpenApi(JSON.stringify(spec).replace(/}:/g, '}//:'));
 
-const getOpenApiHttp = (): OpenApiHttpHandlers<paths> => {
+const getOpenApiHttp = (baseUrl?: string): OpenApiHttpHandlers<paths> => {
     const http = createOpenApiHttp<paths>({
-        baseUrl: process.env.PUBLIC_API_BASE_URL ?? 'http://localhost:8000',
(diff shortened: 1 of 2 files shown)


application/ui/rsbuild.config.ts
```diff
@@ -22,9 +22,9 @@ export default defineConfig({
         define: {
             ...publicVars,
             'import.meta.env.PUBLIC_API_BASE_URL':
-                publicVars['import.meta.env.PUBLIC_API_BASE_URL'] ?? '"http://localhost:8000"',
+                publicVars['import.meta.env.PUBLIC_API_BASE_URL'] ?? '""',
```
ashwinvaidya17, line 25: why was `http://localhost:8000` removed? I thought this part defaults to 8000 if the env variables aren't set

maxxgx: This makes the UI use relative URLs instead of localhost. This solves the issue where if you run the app on a remote machine with SSH, the UI uses localhost, which points to your local machine instead of `<remote-machine-ip>/api/...`. 

I think this is why @rajeshgangireddy had the app stuck at the loading screen. The app was sending requests to the wrong place. With this change, I don't think we need SSH forwarding anymore (which could be a bottleneck for video streaming bandwidth).

ashwinvaidya17: Ah thanks for the explanation

ashwinvaidya17 (approved):

Issue #2054, opened 2024-05-14
[Task]: Throughput in benchmarking


ashwinvaidya17:
### What is the motivation for this task?

Match previous design

### Describe the solution you'd like

Restore torch, lightning, OpenVINO throughput in the new benchmarking design.

ashwinvaidya17:
Saving for later: https://pytorch.org/tnt/stable/utils/generated/torchtnt.utils.flops.FlopTensorDispatchMode.html#torchtnt.utils.flops.FlopTensorDispatchMode

PR #3520, opened 2026-04-09
feat(benchmark): restore throughput and add inference microbench

vnk8071:
## 📝 Description
Benchmark throughput reporting was previously missing/incorrect in the benchmark pipeline (see #2054).
In this PR, I:
- Restore test throughput metrics in benchmark outputs (`test_num_images`, `test_batch_size`, `test_throughput_fps`).
- Add an optional torch-only inference microbenchmark (disabled by default) to measure forward-pass throughput on the test dataloader.
- Update example + integration benchmark configs to include the new inference benchmark options.
- Adjust integration test mocking to keep the benchmark pipeline test stable.

Pasted Graphic

## ✨ Changes
- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ♻️ Refactor (non-breaking change which refactors the code base)
- [x] ⚡ Performance improvements
- [x] 🧪 Tests (adding/modifying tests)
- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🏗️ CI/CD configuration
- [ ] 🔒 Security update
## ✅ Checklist
- [x] I have made the necessary updates to the documentation (if applicable).
- [x] I have written tests that support my changes and prove that my fix is effective or my feature works (if applicable).
- [x] My PR title follows conventional commit format.


vnk8071:
@ashwinvaidya17 Could you please review this benchmark? And let me know if any changes are needed

tools/experimental/benchmarking/sample.yaml
```diff
@@ -1,15 +1,22 @@
 # sample script to show grid search for two categories
 accelerator:
   - cuda
+  - mps
```
ashwinvaidya17, line 4: I agree with the copilot review. You need to update `pipeline.py` as well in order to support mps otherwise you will get an error.

src/anomalib/pipelines/benchmark/job.py
```diff
@@ -152,17 +189,160 @@ def run(
...
+        Args:
+            warmup_runs: Number of warmup iterations.
+            runs: Number of measured iterations.
+            batch_size: Optional override for dataloader batch size.
+
+        Returns:
+            dict[str, Any]: Benchmark metrics. Returns an empty dict if unavailable.
+        """
+        try:
+            import torch
```
ashwinvaidya17: try to avoid inline imports

src/anomalib/pipelines/benchmark/job.py
```diff
@@ -152,17 +189,160 @@ def run(
...
+        num_images: int | None
+        batch_size: int | None
+        try:
+            num_images = len(dataloader.dataset)  # type: ignore[attr-defined]
+        except Exception:  # noqa: BLE001 - best-effort only
+            num_images = None
+        batch_size = getattr(dataloader, "batch_size", None)
+        return {"num_images": num_images, "batch_size": batch_size}
+
+    def _benchmark_inference_throughput(  # noqa: C901
```
ashwinvaidya17, line 319: I feel this entire function can be refactored into smaller components. Feel free to create another file if needed (like utils.py). The `TiledEnsemble` pipeline is a great example

ashwinvaidya17 (commented):
Thanks for the efforts. I have a few concerns with the implementation.
The original code computed throughput on OpenVINO, lightning and torch models (essentially the three modes supported by our inferencers). Ideally all these should be restored. Additionally, can we avoid inline function declarations?
The comment in the PR also mentions `FlopTensorDispatchMode`. Did you explore it? If so, is there a reason it was not used?

vnk8071:
@ashwinvaidya17, can you take a look at this PR?

PR #3546, opened 2026-04-17
fix(sampling): handle coreset_size == 0 in KCenterGreedy

### LINKED ISSUE #3544: 🐞 fix(sampling): KCenterGreedy returns 1 sample when coreset_size == 0
## Describe the bug

`KCenterGreedy.select_coreset_idxs()` violates its size contract when
`int(n_observations * sampling_ratio) == 0`. The caller requests zero samples
but receives one.

PR #3509 added the random seed index to the returned coreset and paired it
with `range(self.coreset_size - 1)`. When `coreset_size` is 0 the range
becomes `range(-1)` (empty), so the seed index is appended and never popped.
The returned list has length 1 instead of 0.

## Steps to reproduce

```python
import torch
from anomalib.models.components.sampling import KCenterGreedy

sampler = KCenterGreedy(embedding=torch.randn(10, 4), sampling_ratio=0.01)
idxs = sampler.select_coreset_idxs()

assert sampler.coreset_size == 0     # passes
assert len(idxs) == 0                 # fails — len(idxs) == 1
```

`sampling_ratio=0.01` together with a 10-row embedding (or any small ratio /
small embedding combination) triggers the path.

## Expected behavior

When `coreset_size == 0`, return an empty list. Silently seeding with a
single observation distorts the downstream distance metric for callers that
rely on the documented size contract.

## OS information

- Python: 3.10+
- anomalib: `main` @ 91d16ef0
- Torch: 2.6+

## Additional context

This is a direct regression introduced by PR #3509. Fix is a single-line
guard; PR with regression test ready to submit:

```python
if self.coreset_size == 0:
    return []
```

shaun0927:
## Summary

Fixes #3544.

`KCenterGreedy.select_coreset_idxs()` returned a coreset of length 1 when
the caller asked for 0. With `int(n * sampling_ratio) == 0`, the greedy
loop `range(coreset_size - 1)` became `range(-1)` (empty) but the seed
index had already been appended, violating the size contract.

**Fix:** early-return an empty list when `coreset_size == 0`.

### Changed files

| File | Change |
|------|--------|
| `src/anomalib/models/components/sampling/k_center_greedy.py` | Return `[]` when `coreset_size == 0` |
| `tests/unit/models/components/sampling/test_k_center_greedy.py` | Add `test_coreset_size_zero` regression test |

<details>
<summary>Before (current main)</summary>

```
>>> sampler = KCenterGreedy(torch.randn(10, 4), sampling_ratio=0.01)
>>> sampler.coreset_size
0
>>> len(sampler.select_coreset_idxs())
1   # contract violated
```
</details>

<details>
<summary>After (this PR)</summary>

```
>>> sampler = KCenterGreedy(torch.randn(10, 4), sampling_ratio=0.01)
>>> sampler.coreset_size
0
>>> sampler.select_coreset_idxs()
[]
```
</details>

<details>
<summary>Unit tests</summary>

```
tests/unit/models/components/sampling/test_k_center_greedy.py::TestKCenterGreedy::test_coreset_size PASSED
tests/unit/models/components/sampling/test_k_center_greedy.py::TestKCenterGreedy::test_initial_point_included PASSED
tests/unit/models/components/sampling/test_k_center_greedy.py::TestKCenterGreedy::test_no_duplicate_indices PASSED
tests/unit/models/components/sampling/test_k_center_greedy.py::TestKCenterGreedy::test_sample_coreset_shape PASSED
tests/unit/models/components/sampling/test_k_center_greedy.py::TestKCenterGreedy::test_coreset_size_one PASSED
tests/unit/models/components/sampling/test_k_center_greedy.py::TestKCenterGreedy::test_indices_within_bounds PASSED
tests/unit/models/components/sampling/test_k_center_greedy.py::TestKCenterGreedy::test_coreset_size_zero PASSED
======================== 7 passed =========================
```
</details>


ashwinvaidya17 (changes requested):
Great catch! One minor comment

shaun0927:
Thanks for taking a look @ashwinvaidya17! I can see the `CHANGES_REQUESTED` review but I don't think the inline comment came through on my end — the review body just says *"Great catch! One minor comment"* with no attached file/line comment visible in the Files tab.

Could you re-post the suggestion (line reference would be great)? Happy to turn it around quickly once I know which spot you meant.

src/anomalib/models/components/sampling/k_center_greedy.py
```diff
@@ -118,6 +118,13 @@ def select_coreset_idxs(self) -> list[int]:
             self.features = self.embedding.reshape(self.embedding.shape[0], -1)
             self.reset_distances()
 
+        # Degenerate case: a caller that asked for zero samples should get an
+        # empty coreset, not the seed index. ``range(coreset_size - 1)`` would
+        # become ``range(-1)`` and silently skip the loop, leaving the seed
+        # behind and violating the size contract.
+        if self.coreset_size == 0:
```
ashwinvaidya17: Maybe we can add a validation in `__init__` instead? If `coreset_size` is 0, `sample_coreset` will return an empty tensor.

shaun0927:
Thanks for the suggestion, @ashwinvaidya17!

Moved the guard into `__init__`: after computing `coreset_size = int(n * sampling_ratio)`, if it is `<= 0` a `ValueError` is raised immediately with the offending values (`sampling_ratio`, `n`) in the message. The stale early-return + comment block in `select_coreset_idxs` are gone since that path is now unreachable.

The regression test (`test_coreset_size_zero`) has been rewritten as `test_coreset_size_zero_raises` and asserts the exception at construction time via `pytest.raises(ValueError)`.

Landed in 03a48a1cdd75c4f9d12b8814d6580e8727b5f682.

ashwinvaidya17 (approved):
Thanks

PR #3547, opened 2026-04-17
fix(data): reject str size in ExportableCenterCrop

shaun0927:
## Summary

Fixes #3545.

PR #3539 broadened `ExportableCenterCrop.__init__` from
`isinstance(size, tuple)` to `isinstance(size, Sequence)`. Because `str`
is a `Sequence`, passing a string such as `"224"` silently became the
3-element list `["2", "2", "4"]` and blew up later inside
`_center_crop_parse_output_size` with an opaque error.

**Fix:** route `int` through `[size, size]`, explicitly exclude
`str`/`bytes`/`bytearray` from the sequence branch, and raise a clear
`TypeError` otherwise — matching the torchvision `CenterCrop` contract.

### Changed files

| File | Change |
|------|--------|
| `src/anomalib/data/transforms/center_crop.py` | Split `int` / `Sequence` / else branches, reject str-like sequences |
| `tests/unit/data/transforms/__init__.py` | **New** — test package init |
| `tests/unit/data/transforms/test_center_crop.py` | **New** — 5 regression tests (int, tuple, list, str, bytes, transform end-to-end) |

<details>
<summary>Before (current main)</summary>

```
>>> from anomalib.data.transforms.center_crop import ExportableCenterCrop
>>> ExportableCenterCrop(size="224").size
["2", "2", "4"]
```
</details>

<details>
<summary>After (this PR)</summary>

```
>>> ExportableCenterCrop(size=224).size
[224, 224]
>>> ExportableCenterCrop(size=(224, 256)).size
[224, 256]
>>> ExportableCenterCrop(size="224")
  ...
TypeError: size must be int or Sequence[int], got str
```
</details>

<details>
<summary>Unit tests</summary>

```
tests/unit/data/transforms/test_center_crop.py::TestExportableCenterCropSize::test_int_size_is_squared PASSED
tests/unit/data/transforms/test_center_crop.py::TestExportableCenterCropSize::test_sequence_size_is_copied PASSED
tests/unit/data/transforms/test_center_crop.py::TestExportableCenterCropSize::test_str_size_is_rejected PASSED
tests/unit/data/transforms/test_center_crop.py::TestExportableCenterCropSize::test_bytes_size_is_rejected PASSED
tests/unit/data/transforms/test_center_crop.py::TestExportableCenterCropSize::test_transform_still_works_for_int_size PASSED
======================== 5 passed =========================
```
</details>


src/anomalib/data/transforms/center_crop.py
```diff
@@ -129,7 +129,20 @@ class ExportableCenterCrop(Transform):
 
     def __init__(self, size: int | Sequence[int]) -> None:
         super().__init__()
-        self.size = list(size) if isinstance(size, Sequence) else [size, size]
+        # ``str`` and ``bytes`` are ``Sequence`` instances in Python, so a plain
```
ashwinvaidya17: Thanks for spotting this. I think we can just check `if isinstance(size, int)` to match the typing. That way we don't have to compare with the rest

ashwinvaidya17 (changes requested):
Thanks for spotting and fixing. Can you have a look at at prek as well?

shaun0927:
Thanks for the review @ashwinvaidya17! Simplified the dispatch to a single `isinstance(size, int)` check as you suggested — the `else` branch now trusts `Sequence[int]` per the declared typing, and the explicit `TypeError` / str / bytes rejection tests are gone (71229ef).

Re: `prek` — the failure on the previous run was a transient HTTP 502 Bad Gateway from PyPI while `uv` tried to build `hadolint-py` (log: `× Failed to build hadolint-py ... HTTP Error 502: Bad Gateway or Proxy Error`). Nothing in the PR touches the hook config, so the re-run triggered by this push should pick it up cleanly. I'll keep an eye on it and push a follow-up if a real prek finding surfaces.

src/anomalib/data/transforms/center_crop.py
```diff
@@ -129,7 +129,13 @@ class ExportableCenterCrop(Transform):
...
     def __init__(self, size: int | Sequence[int]) -> None:
         super().__init__()
-        self.size = list(size) if isinstance(size, Sequence) else [size, size]
+        # Match the declared ``int | Sequence[int]`` typing: route ``int``
+        # through ``[size, size]`` (mirroring torchvision ``CenterCrop``) and
+        # copy any other sequence as-is. The previous ``isinstance(size, tuple)``
+        # check misclassified plain ``list`` inputs, and broadening it to
+        # ``Sequence`` without a narrower ``int`` guard silently split ``str``
+        # inputs like ``"224"`` into ``["2", "2", "4"]``.
+        self.size = [size, size] if isinstance(size, int) else list(size)
```
ashwinvaidya17: Wouldn't `list(size)` give the same issue as before? Maybe we can use `_setup_size` from torchvision's transforms https://docs.pytorch.org/vision/main/_modules/torchvision/transforms/transforms.html. The base classes has
```python
class CenterCrop(torch.nn.Module):
    ...

    def __init__(self, size):
        super().__init__()
        _log_api_usage_once(self)
        self.size = _setup_size(size, error_msg="Please provide only two dimensions (h, w) for size.")
```
So we can just import `_setup_size` from transforms and call `self.size = _setup_size(...`

shaun0927:
Good catch, thanks! You were right — `list("224")` would still silently produce `['2', '2', '4']` since the `isinstance(size, int)` guard only fires for plain ints, and a bare string falls through to `list(size)` which iterates its characters.

Switched to `torchvision.transforms.transforms._setup_size` as you suggested (78d4f6d8ba9580b582034db7e46d775a30aea686). `ExportableCenterCrop` now matches upstream `CenterCrop` exactly — int normalisation to `[size, size]`, len==2 enforcement for sequences, and str rejection all come from torchvision. `_setup_size` raises `ValueError` for str input, so the regression test was updated to match that.

ashwinvaidya17 (approved):
Thanks

PR #3572, opened 2026-04-30
fix: resolve PIL Image type annotation causing NameError

Midoriya-w:
### Fix: Image type annotation causing NameError

This PR fixes a NameError caused by incorrect type annotations using Image.Image.

###Changes
- Updated type annotations to correctly reference PIL Image
- Ensured Image is properly imported in test files
- Fixed failing visualization unit tests

###Result
- All tests in tests/unit/visualization/image/ now pass successfully


Midoriya-w:
Happy to make any changes if needed. Thanks for reviewing!

ashwinvaidya17:
Can you give more context on this fix? Which version of python are you using? What is the command that you are using to run visualization unit tests?
Would be useful to have the steps from env creation to the actual reproducer that shows the failing type check

Midoriya-w:
Hi @ashwinvaidya17! I was using Python 3.10. The error occurred because PIL.Image.Image was being used as a type annotation directly in function signatures without from  future import annotations, causing a NameError at runtime when the type was evaluated eagerly.
To reproduce:
bash
conda create -n anomalib-env python=3.10
conda activate anomalib- env
pip install -e ".[dev]"
pytest tests/unit/visualization/image/ -v

The tests fail with NameError: name 'Image' is not defined before the fix.
After the fix, all 6 tests in that directory pass. Happy to add a minimal reproducer in the PR description if that would help!

ashwinvaidya17:
Which version of python 3.10 are you using? I have `3.10.18` on my system.
Maybe it is a conda issue?
These are the steps I followed
```bash
uv venv .venv --python=3.10
source .venv/bin/activate
uv pip install -e ".[dev]"
pytest tests/unit/visualization/image/ -v
```

The tests pass on my end

Midoriya-w:
"Thanks for the guidance. I reproduced using uv with Python 3.10.20 and the tests pass without the fix. The issue appears to have been specific to my conda environment. I'll close this PR."

PR #3634, opened 2026-06-11
fix(data): use split.value for pandas comparison in dataset loaders

### LINKED ISSUE #3581: 🐞 Default MVTecAD fails to build datasets
### Describe the bug

When using the MVTecAD dataset I simply load it up, use the prepare_data and setup and when I call the train_dataloader on top it seems like the data is empty. This is due to a bug related to the interaction between the Split Enum and Pandas.

## Error statement
```
Zero subset length encountered during splitting. This means one of your subsets
            might be empty or devoid of either normal or anomalous images.
  File "/Users/amirgheser/SSA-IAD/bug.py", line 9, in <module>
    for i, data in enumerate(datamodule.train_dataloader()):
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/amirgheser/SSA-IAD/.venv/lib/python3.12/site-packages/anomalib/data/datamodules/base/image.py", line 373, in train_dataloader
    return DataLoader(
           ^^^^^^^^^^^
  File "/Users/amirgheser/SSA-IAD/.venv/lib/python3.12/site-packages/torch/utils/data/dataloader.py", line 394, in __init__
    sampler = RandomSampler(dataset, generator=generator)  # type: ignore[arg-type]
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/amirgheser/SSA-IAD/.venv/lib/python3.12/site-packages/torch/utils/data/sampler.py", line 149, in __init__
    raise ValueError(
```

## Proposed fix
In the `make_mvtec_ad_dataset` function simply add this line or anything with a similar logic.
```
    split = split.value if isinstance(split, Split) else split
```
Basically the dataframe is being fully and correctly handled until we reach the last if statement `if split: `
where the dataframe is finally being emptied because the equality between enum and string is always False.

### Dataset

MVTecAD

### Model

N/A

Executing this simple script to check that the dataloader is working fails
```python
if __name__ == "__main__":
    from anomalib.data import MVTecAD
    
    datamodule = MVTecAD(category="bottle")
    datamodule.prepare_data()
    datamodule.setup()
    
    for i, data in enumerate(datamodule.train_dataloader()):
        print(data.keys())
        print(data["image"].shape)
        break
```

### OS information

OS information:
- OS: [e.g. Ubuntu 20.04]
- Python version: [e.g. 3.10.0]
- Anomalib version: [e.g. 0.3.6]
- PyTorch version: [e.g. 1.9.0]
- CUDA/cuDNN version: [e.g. 11.1]
- GPU models and configuration: [e.g. 2x GeForce RTX 3090]
- Any other relevant information: [e.g. I'm using a custom dataset]

### Expected behavior

No error message :)

### Screenshots

_No response_

### Pip/GitHub

pip

2.4.0

### Configuration YAML

```yaml
.
```

### Logs

```shell
.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

won-seoop:
## 📝 Description

In **pandas 3.x**, comparing a DataFrame column of string values against a `str`-based Enum instance (e.g. `Split.TRAIN`) returns an **empty result**, even though `Split` inherits from `str`. This silently breaks all dataset loaders that filter rows with `samples[samples.split == split]` when a `Split` enum is passed, causing the dataloader to produce zero samples.

**Reproduction (pandas 3.x):**
```python
import pandas as pd
from enum import Enum

class Split(str, Enum):
    TRAIN = "train"

df = pd.DataFrame({"split": ["train", "train", "test"]})
print(len(df[df["split"] == Split.TRAIN]))   # 0 — wrong!
print(len(df[df["split"] == Split.TRAIN.value]))  # 2 — correct
```

The fix normalises the split argument to its plain string `.value` before the comparison, following the pattern already established in `make_mvtec2_dataset` (see `mvtecad2.py` line 292–293).

🛠️ Fixes #3581

## ✨ Changes

**Affected files:**
- `src/anomalib/data/datasets/image/mvtecad.py`
- `src/anomalib/data/datasets/image/mvtec_loco.py`
- `src/anomalib/data/datasets/image/btech.py`
- `src/anomalib/data/datasets/image/kolektor.py`
- `src/anomalib/data/datasets/depth/mvtec_3d.py`

## ✅ Checklist

- [x] 🧪 The fix is consistent with the pattern already used in `mvtecad2.py`.

**Note:** Full unit tests (which require dataset downloads) were not run locally. The fix was verified with a minimal pandas 3.x reproduction script confirming the root cause and the correctness of the `.value`-based comparison.


samet-akcay (changes requested):
Thanks for creating this pr @won-seoop 

I see the same changes might be needed in the following files? 

| File | Line |
|---|---|
| `src/anomalib/data/datasets/image/mpdd.py` | 208 |
| `src/anomalib/data/datasets/image/bmad.py` | 223 |
| `src/anomalib/data/datasets/image/vad.py` | 202 |
| `src/anomalib/data/datasets/image/kaputt.py` | 403 |
| `src/anomalib/data/datasets/image/datumaro.py` | 105 |
| `src/anomalib/data/datasets/video/ucsd_ped.py` | 251 |
| `src/anomalib/data/datasets/video/avenue.py` | 174 |
| `src/anomalib/data/datasets/video/shanghaitech.py` | 290 |
| `src/anomalib/data/datasets/depth/adam_3d.py` | 224 |

any reason why these are not modified ?

PR #3680, opened 2026-07-20
🐞 fix(data): distribute split remainder over subsets to avoid ZeroDivisionError

### LINKED ISSUE #3679: 🐞 random_split raises ZeroDivisionError when a label group is too small for the split ratios (label-aware splitting)
### Describe the bug

`random_split` in `src/anomalib/data/utils/split.py` distributes remainder samples with:

```python
subset_idx = i % sum(subset_lengths)
```

When a label group is so small that `floor(len * ratio)` is 0 for every ratio (e.g. a **1-sample group** with the default 50/50 val split), `sum(subset_lengths)` is 0 and the modulo raises `ZeroDivisionError`.

This is easy to hit in practice: a `Folder` datamodule with a small dataset and no `normal_test_dir` puts a single normal image into the test set; the default `ValSplitMode.FROM_TEST` (ratio 0.5, `label_aware=True`) then crashes in `datamodule.setup()` before training starts, with no hint that the dataset is simply small. The function even has a dedicated warning for this exact situation ("Zero subset length encountered...") a few lines below, but the crash fires first.

### Steps to reproduce

Minimal mechanism:

```python
import math
samples, split_ratio = 1, [0.5, 0.5]
subset_lengths = [math.floor(samples * r) for r in split_ratio]  # [0, 0]
i = 0
i % sum(subset_lengths)  # ZeroDivisionError
```

End-to-end (anomalib 2.5.1): put 8 normal images in `Temp/ok` and 4 anomalous in `Temp/nok`, then:

```python
from anomalib.data import Folder

dm = Folder(name="Temp", normal_dir="Temp/ok", abnormal_dir="Temp/nok",
            seed=42, train_batch_size=1, num_workers=0)
dm.setup()  # ZeroDivisionError: integer division or modulo by zero
```

```
  File ".../anomalib/data/datamodules/base/image.py", line 159, in setup
    self._create_val_split()
  File ".../anomalib/data/datamodules/base/image.py", line 346, in _create_val_split
    self.test_data, self.val_data = random_split(
  File ".../anomalib/data/utils/split.py", line 168, in random_split
    subset_idx = i % sum(subset_lengths)
ZeroDivisionError: integer division or modulo by zero
```

### Expected behavior

The split should not crash. The remainder should be distributed over the subsets (`i % len(subset_lengths)`, as in `torch.utils.data.random_split`), letting the existing "Zero subset length" warning inform the user that a subset may be empty.

### Environment

- anomalib 2.5.1 (pip); the same code is present on current `main`
- PyTorch 2.11.0+cpu, Python 3.10, Windows 11

I have a one-line fix plus a regression test ready (the test fails with the exact `ZeroDivisionError` on current code and passes with the fix) — opening a PR shortly.

prawnsgupta:
## 📝 Description

Fixes #3679.

`random_split` distributes remainder samples with `subset_idx = i % sum(subset_lengths)`. When a label group is too small for the split ratios — e.g. a **single sample** with the default 50/50 label-aware val split — every `floor(len * ratio)` is 0, the sum is 0, and the modulo raises `ZeroDivisionError` from `datamodule.setup()` before training starts. Easy to hit with a small custom `Folder` dataset and no `normal_test_dir` (one normal image lands in the test set, and the default `ValSplitMode.FROM_TEST` crashes).

**Fix (one word):** distribute the remainder round-robin over the **subsets** — `i % len(subset_lengths)` — matching `torch.utils.data.random_split`. In the previously-working cases the chosen index is unchanged; in the crash case the sample goes to the first subset and the existing "Zero subset length" warning (which sits three lines below but could never fire) now fires as intended.

## ✨ Changes

- `src/anomalib/data/utils/split.py`: `sum(subset_lengths)` → `len(subset_lengths)` in the remainder-distribution loop.
- `tests/unit/data/utils/test_split.py` (new — there were no `random_split` tests): total-length preservation, the 1-sample-label-group regression test, label representation in label-aware splits, and the invalid-ratio guard. Tests build a real `FolderDataset` over `tmp_path`-generated images, no fixtures needed.

## ✅ Verification

- Regression test fails on the previous code with exactly `split.py:168: ZeroDivisionError` and passes with the fix.
- End-to-end: the `Folder.setup()` reproducer from #3679 now completes and logs the intended zero-subset-length warning.
- `tests/unit/data/utils`: 98 passed; remaining failures are environment-specific (missing `pytest_mock`, Windows permission-semantics tests) and unrelated to this change.

## 📋 Checklist

- [x] I have added tests that prove my fix is effective

### DIFF
--- a/src/anomalib/data/utils/split.py
+++ b/src/anomalib/data/utils/split.py
@@ -165,7 +165,7 @@ def random_split(
         # get subset lengths
         subset_lengths = [math.floor(len(label_dataset.samples) * ratio) for ratio in split_ratio]
         for i in range(len(label_dataset.samples) - sum(subset_lengths)):
-            subset_idx = i % sum(subset_lengths)
+            subset_idx = i % len(subset_lengths)
             subset_lengths[subset_idx] += 1
         if 0 in subset_lengths:
             msg = """Zero subset length encountered during splitting. This means one of your subsets
--- a/tests/unit/data/utils/test_split.py
+++ b/tests/unit/data/utils/test_split.py
@@ -1,0 +1,75 @@
+# Copyright (C) 2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for dataset split utils."""
+
+from pathlib import Path
+
+import numpy as np
+import pytest
+from PIL import Image
+
+from anomalib.data.datasets.image.folder import FolderDataset
+from anomalib.data.utils.split import random_split
+
+
+def _make_image(path: Path) -> None:
+    """Write a small dummy RGB image to ``path``."""
+    rng = np.random.default_rng(seed=len(str(path)))
+    image = rng.integers(0, 255, size=(32, 32, 3), dtype=np.uint8)
+    path.parent.mkdir(parents=True, exist_ok=True)
+    Image.fromarray(image).save(path)
+
+
+def _make_folder_dataset(root: Path, num_normal: int, num_abnormal: int) -> FolderDataset:
+    """Create a ``FolderDataset`` backed by generated dummy images."""
+    for i in range(num_normal):
+        _make_image(root / "good" / f"{i:03d}.png")
+    for i in range(num_abnormal):
+        _make_image(root / "bad" / f"{i:03d}.png")
+    return FolderDataset(
+        name="dummy",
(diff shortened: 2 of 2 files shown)


ashwinvaidya17 (approved):
Thanks for identifying and fixing this bug

PR #3729, opened 2026-08-23
fix(metrics): use tensor boolean reduction in F1AdaptiveThreshold sample validation

Xayar145:
### Description

Fixes sample presence checks in `_F1AdaptiveThreshold`:
- In `_has_anomalous_samples` and `_has_normal_samples`, replace Python `1 in batch` / `0 in batch` membership tests with vectorized PyTorch boolean reductions `(batch == 1).any().item()` and `(batch == 0).any().item()`.
- Ensures reliable detection of missing anomalous/normal classes and correct fallback to `_get_max_pred()` or `_get_min_pred()`.
- Adds unit tests in `tests/unit/metrics/threshold/test_threshold.py` covering normal-only and anomalous-only validation batches.

### DIFF
--- a/tests/unit/metrics/threshold/test_threshold.py
+++ b/tests/unit/metrics/threshold/test_threshold.py
@@ -49,3 +49,34 @@ class TestBaseThreshold:
 
         with pytest.raises(NotImplementedError, match=r"Subclass of Threshold must implement the update method"):
             base_threshold.update()
+
+
+class TestF1AdaptiveThreshold:
+    """Test cases for F1AdaptiveThreshold class."""
+
+    @staticmethod
+    def test_normal_samples_only() -> None:
+        """Test threshold computation when only normal samples are present."""
+        import torch
+        from anomalib.metrics.threshold.f1_adaptive_threshold import _F1AdaptiveThreshold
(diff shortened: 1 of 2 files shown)


tests/unit/metrics/threshold/test_threshold.py
```diff
@@ -49,3 +49,34 @@ def test_base_threshold_abstract_methods() -> None:
...
+
+
+class TestF1AdaptiveThreshold:
+    """Test cases for F1AdaptiveThreshold class."""
+
+    @staticmethod
+    def test_normal_samples_only() -> None:
+        """Test threshold computation when only normal samples are present."""
+        import torch
+        from anomalib.metrics.threshold.f1_adaptive_threshold import _F1AdaptiveThreshold
```
ashwinvaidya17, line 61: Our standard practice is to avoid inline imports if possible in favour of file-level imports.

ashwinvaidya17 (changes requested):
Thanks for the PR but do you have a reproducer where this issue is visible? Are you using torch.compile in your usecase? This does not seem to be a problem in torch's eager mode.

PR #3738, opened 2026-08-25
fix(data): treat seed=0 as a valid seed in random_split

VenishPaneliya:
`random_split` builds its generator like this:

```python
random_state = torch.Generator().manual_seed(seed) if seed else None
```

so the falsy `0` takes the same branch as `None` and the split is drawn from an unseeded generator. The signature is `seed: int | None = None` and the docstring describes it as "Random seed for reproducibility", so `0` is meant to be an ordinary seed rather than a sentinel.

Calling it three times with the same seed:

```python
for _ in range(3):
    splits = random_split(dataset, [0.5, 0.5], seed=0)
```

On `main` the three splits differ from each other, while `seed=42` is stable:

```
seed=0    3 repeated splits identical? False
             ['img_00.png', 'img_01.png', 'img_02.png', 'img_04.png', 'img_07.png']
             ['img_00.png', 'img_05.png', 'img_06.png', 'img_08.png', 'img_11.png']
             ['img_00.png', 'img_01.png', 'img_03.png', 'img_04.png', 'img_06.png']
seed=42   3 repeated splits identical? True
```

With `is not None` both are stable, and every non-zero seed produces byte-identical output to before, so only the `0` case changes.

This is reachable from `AnomalibDataModule(seed=0)`, which stores the value and passes it into the test-split, `FROM_TRAIN`, `FROM_TEST` and `SYNTHETIC` paths, so a run configured with `seed: 0` silently gets non-deterministic splits.

`random_split` is the only place in `src/` using this pattern, so there is no sibling to change.

Added `test_seed_zero_is_reproducible`, which fails on `main` and passes here.


ashwinvaidya17 (approved):
Thanks for the fix

PR #3739, opened 2026-08-26
chore(typing): remove type-only references to wandb's removed RunDisabled class

dmitryduev:
wandb deprecated `RunDisabled` in v0.17.6 (wandb/wandb#8064): since then, `wandb.init(mode="disabled")` returns a regular no-op `wandb.Run`, and `RunDisabled` has been an empty compatibility shim that emits a deprecation warning on attribute access. The shim is being removed in an upcoming wandb release (wandb/wandb#12586), after which importing it raises `ImportError`.

This PR drops the remaining references so the integration keeps working with upcoming wandb releases, while remaining compatible with older ones.

Details: the references here are `TYPE_CHECKING`-only, so there is no runtime break — but mypy and docs builds would fail once the name disappears from wandb. Annotations narrowed to `Run | None`.


ashwinvaidya17:
@dmitryduev thanks for the PR. Do we need to pin the lowest version for wandb for this change?

ashwinvaidya17 (approved):

PR #3771, opened 2026-09-03
feat(data): Use temporary workspace folder for anomalib artifacts

warrkan:


application/backend/src/services/model_service.py
```diff
@@ -178,25 +178,24 @@ async def export_model(self, project_id: UUID, model_id: UUID, export_parameters
...
+                / "model.ckpt"
+            )
+            # Resolve 'latest' to actual version dir to avoid traversing junction (e.g. WinError 448 on Windows)
+            ckpt_path = resolve_versioned_path(ckpt_path)
             if not ckpt_path.exists():
-                raise FileNotFoundError(f"Model checkpoint not found at {ckpt_path}")
+                # Try alternative path for older structure or if title case isn't used
+                ckpt_path = Path(model_binary_repo.model_folder_path) / "weights" / "lightning" / "model.ckpt"
+                if not ckpt_path.exists():
+                    raise FileNotFoundError(f"Model checkpoint not found at {ckpt_path}")
```
ashwinvaidya17: Do you think this is relevant if the model checkpoints are copied successfully? In case the copy or save fails, it will fail at the end of the training so maybe when we start the training it should not look at the Anomalib Engine's defined path.

application/backend/src/services/training_service.py
```diff
@@ -251,47 +267,65 @@ def _train_model(
...
-            logger=[tensorboard],
-            devices=[device_index] if device_index is not None else 1,  # Single GPU training for now
-            max_epochs=max_epochs,
-            callbacks=[
-                AnomalibStudioProgressCallback(synchronization_parameters),
-                EarlyStopping(monitor="pixel_AUROC", mode="max", patience=5),
-            ],
-            **kwargs,
-        )
+        work_dir = settings.data_dir / f"train-workspace-{model.id}"
```
ashwinvaidya17, line 270: My concern is that if data dir is nested in a long path like `c:/users/name/AnomalibStudio/...` we might still hit the max path length limit on windows due to UUID before we move the model checkpoints. But I may be wrong.

ashwinvaidya17 (commented):
Thanks for the PR. Has this been tested on MSIX and UWP build? I remember I was facing long path issue on of of them

application/backend/src/services/training_service.py
warrkan: The longest path in this case will be (for example):
`C:\Users\<username>\AppData\Local\Intel\AnomalibStudio\train-workspace-102dad00-e1ae-498d-a94c-de99d438fc84\Padim\3832b055-8e02-412c-b11c-791f777b15d4-padim\v0\weights\lightning\model.ckpt`
which still gives around 80 characters space for `<username>`

application/backend/src/services/model_service.py
warrkan: I can remove all the fallbacks since we haven't release Anomalib Studio yet

ashwinvaidya17 (dismissed):

application/backend/src/services/model_service.py
ashwinvaidya17: Would be great

ashwinvaidya17: But maybe we can do it in a different PR?

application/backend/src/services/model_service.py
```diff
@@ -178,25 +177,9 @@ async def export_model(self, project_id: UUID, model_id: UUID, export_parameters
...
-        # Resolve 'latest' to actual version dir to avoid traversing junction (e.g. WinError 448 on Windows)
-        ckpt_path = resolve_versioned_path(ckpt_path)
-
+        ckpt_path = Path(model_binary_repo.model_folder_path) / "checkpoint" / "model.ckpt"
         if not ckpt_path.exists():
-            # Try alternative path for older structure or if title case isn't used
-            ckpt_path = Path(model_binary_repo.model_folder_path) / "weights" / "lightning" / "model.ckpt"
-            if not ckpt_path.exists():
-                raise FileNotFoundError(f"Model checkpoint not found at {ckpt_path}")
+            raise FileNotFoundError(f"Model checkpoint not found at {ckpt_path}")
```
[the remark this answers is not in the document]
warrkan: Product has not been release yet, so it's acceptable

ashwinvaidya17 (approved):
Thanks for the PR

PR #3774, opened 2026-09-04
fix(CI): gate issue triage on maintainer-applied triage label

ashwinvaidya17:
## Description

Follow-up to #3770, addressing https://github.com/open-edge-platform/anomalib/pull/3770#discussion_r3924239917 (thanks @AlexanderBarabanov).

That PR replaced the `author_association` check with a two-stage `analyze` (read-only LLM) → `apply` (deterministic, allowlist-validated) split. That closes the *mutation* vector, but the workflow still triggered on `issues: opened`, so any GitHub user could schedule an up-to-20-minute LLM run on the self-hosted GPU runner just by filing an issue.

Triage is now **opt-in**: a maintainer requests it by applying the `triage` label.

### Changes

- **`issues: [labeled]` trigger**, gated on `github.event.label.name == 'triage'`. No longer runs automatically on newly opened issues.
- **New `gate` job** — hosted runner, `issues: read`, no GPU. Verifies the labeler has write/maintain/admin access and that the issue has not been edited since the label was applied. Emits `approved` / `labeled_at`; `analyze` only runs on `approved == 'true'`.
- **TOCTOU re-check in `apply`** — the gate check alone is insufficient, since `analyze` can run for 20 minutes after labeling. `apply-issue-triage.sh` re-verifies before it writes anything, so an edit landing mid-run cannot swap out the content the maintainer approved.
- **New `reject` job** — removes the label and posts a fixed, hard-coded string when a request is rejected as stale. Kept separate from `gate` so the decision stays read-only, and separate from `apply` so it never touches model output.
- **Replay prevention** — the `triage` label is removed once the pipeline finishes with an issue, and a suggestion can never re-apply it.
- New `triage-lib.sh` holds the shared timestamp helpers used by both the gate and apply paths.

### Notes for reviewers

- **`lastEditedAt`, not `updated_at`.** The review comment suggested comparing against the issue's last-modified timestamp, but `updated_at` is bumped by comments, label changes, assignment and milestones, so it would abort valid runs constantly. This uses GraphQL `lastEditedAt` (body edits) combined with `renamed` timeline events (title edits), taking the later of the two. Both are fixed-width UTC, so lexicographic comparison is chronological.
- **Label timestamp source.** The `labeled` webhook payload does not carry the label application time, so it is recovered from the issue timeline.
- **Fail-closed on a missing label timestamp**, treated as stale.
- **Permission check degrades gracefully.** If the collaborator-permission endpoint is unreachable with this workflow's minimal token scopes, the run logs a warning and falls back to GitHub's built-in requirement that only users with triage-or-higher access can apply labels. It never widens access. Happy to make this a hard failure instead if you'd prefer.
- The job-level `if` uses the literal string `'triage'` because the workflow-level `env` is not available in job-level `if` expressions.

### Prerequisite :warning:

**The `triage` label must be created in this repository before this merges.** There is no labels-as-code file in `.github/`, so this is a one-time manual step. Until the label exists the workflow is simply a no-op.

## Checks

- `actionlint` clean
- `shellcheck -x` clean on all three scripts
- `pre-commit` (incl. `zizmor`, `prettier`, `check-yaml`) passing
- Timestamp comparison logic exercised locally against stubbed API responses: edit-before-label, edit-after-label, never-edited, missing-label-timestamp, same-second, and cross-year cases all behave as intended

### DIFF
--- a/.github/scripts/apply-issue-triage.sh
+++ b/.github/scripts/apply-issue-triage.sh
@@ -9,19 +9,47 @@
 # reach it except as inert data inside a JSON field, which is validated
 # against strict allowlists/limits below before any `gh` write command runs.
 #
-# Required env vars: GH_TOKEN, GITHUB_REPOSITORY, ISSUE_NUMBER, SUGGESTION
+# Required env vars: GH_TOKEN, GITHUB_REPOSITORY, ISSUE_NUMBER, SUGGESTION,
+# LABELED_AT
 
 set -euo pipefail
 
 : "${GITHUB_REPOSITORY:?}"
 : "${ISSUE_NUMBER:?}"
+: "${LABELED_AT:?}"
 SUGGESTION="${SUGGESTION:-}"
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=.github/scripts/triage-lib.sh
+source "${SCRIPT_DIR}/triage-lib.sh"
 
 MAX_LABELS=3
 MAX_COMMENT_CHARS=2000
 
+# Drop the request label once the pipeline is done with the issue, so a
+# completed (or abandoned) run can't be replayed and re-triage always
+# requires a maintainer to deliberately re-apply the label.
+remove_triage_label() {
+  gh issue edit "$ISSUE_NUMBER" --repo "$GITHUB_REPOSITORY" \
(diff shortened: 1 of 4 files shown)


AlexanderBarabanov:
Thanks for this proposal! @ashwinvaidya17 
As this workflow has been [removed](https://github.com/open-edge-platform/anomalib/pull/3802), I think we may close this PR.

PR #3793, opened 2026-09-14
feat(model): Add FoundAD few-shot anomaly detection model

rajeshgangireddy:
Adds **FoundAD** ([paper](https://arxiv.org/abs/2510.01934), [official repo](https://github.com/ymxlzgy/FoundAD)): freezes a DINOv2/DINOv3 encoder and trains a lightweight manifold projector on CutPaste anomalies, enabling few-shot (1/2/4-shot) anomaly detection.

Reproduced the paper's exact multi-class-one-model few-shot protocol; verified export (TORCH/ONNX/OPENVINO incl. FP16/INT8) and integration tests (8/8 passing), lint/format clean.

**I-AUROC, ours vs. paper (Table 1):**

| Dataset | 1-shot | 2-shot | 4-shot |
| --- | --- | --- | --- |
| MVTec-AD | 94.5 / 96.1 | 95.5 / 96.8 | 96.7 / 97.1 |
| VisA | 92.2 / 92.6 | 93.5 / 93.5 | 93.6 / 94.4 |

Residual gap vs. paper is explained by fewer training epochs and single-seed sampling (see README).


src/anomalib/models/image/foundad/components/augmentation.py
```diff
@@ -0,0 +1,142 @@
...
+    def __init__(
+        self,
+        p_orient: float = 0.3,
+        p_appear: float = 0.3,
+        blur_kernel_size: int = 23,
+    ) -> None:
+        self.p_orient = p_orient
+        self.p_appear = p_appear
+        self._orient_ops: list[Callable[[torch.Tensor], torch.Tensor]] = [
+            _horizontal_flip,
```
ashwinvaidya17: can't we use torchvision's horizontal and vertical flip? Also, can we move random rotate to anomalib's transforms folder? This way we can reuse it elsewhere

src/anomalib/models/image/foundad/components/cutpaste.py
```diff
@@ -0,0 +1,252 @@
...
+from collections.abc import Callable
+
+import torch
+import torchvision.transforms.functional as TF  # noqa: N812
+from torchvision import transforms
+
+# ImageNet normalization stats used by FoundAD's default pre-processor. Needed to
+# temporarily map patches back to real [0, 1] pixel space before color jitter.
+_IMAGENET_MEAN = (0.485, 0.456, 0.406)
+_IMAGENET_STD = (0.229, 0.224, 0.225)
```
ashwinvaidya17: if this is only used in _apply_in_pixel_space, maybe we can just move it in the function?

src/anomalib/models/image/foundad/torch_model.py
```diff
@@ -0,0 +1,290 @@
...
+DINOV2_ARCHITECTURES: dict[str, dict[str, int]] = {
+    "small": {"embed_dim": 384, "num_heads": 6},
+    "base": {"embed_dim": 768, "num_heads": 12},
+    "large": {"embed_dim": 1024, "num_heads": 16},
+}
+
+# Inference constants
+DEFAULT_GAUSSIAN_KERNEL_SIZE = 5
+DEFAULT_GAUSSIAN_SIGMA = 4.0
+
```
ashwinvaidya17: Same. Do we need these? Maybe we can just move the values to GaussianKernel2d. If people want to change those values, they can always do it there. Not sure what is the best practice there?

ashwinvaidya17 (commented):
Thanks for the efforts! Just a few minor comments.

src/anomalib/models/image/foundad/components/augmentation.py
rajeshgangireddy: Sure. Can do both.

src/anomalib/models/image/foundad/components/cutpaste.py
rajeshgangireddy: agreed.

src/anomalib/models/image/foundad/torch_model.py
rajeshgangireddy: I am a bit hesitant  to move the values to GaussianKernel2d. I see that other models also specify their guassian sigma constants in their own torch_model.py (for example Dinomaly and iNP Former has DEFAULT_GAUSSIAN_SIGMA).
So, I would just leave these as is for noww.

PR #3801, opened 2026-09-17
fix(weights): change how models are serialized and deserialized

ashwinvaidya17:
## 📝 Description

- Addresses PTK0009203
- Currently anomalib engine loads checkpoints using `weights_only=False`. This will end up executing malicious coded if the checkpoint comes from untrusted sources. This PR proposes changes to `AnomalibModule` to mitigate this.
- An alternative to this design is to use `TRUST_REMOTE_CODE=True` but this might make the users set this is their default env variable which is risky.

## Design

This design introduces a **scoped allowlist** used at both checkpoint load seams:

1. Shared first-party enums (`ANOMALIB_SAFE_GLOBALS`)
2. Shared numpy leaf types (`NUMPY_SAFE_GLOBALS`)
3. Optional per-model extras via `AnomalibModule.checkpoint_safe_globals()`

### In Anomalib Module

Override

```python
@classmethod
 def load_from_checkpoint(
      ...
  ) -> dict[str, Any]:
    with anomalib_safe_globals(extra=cls.checkpoint_safe_globals()):
            return super().load_from_checkpoint(
```

Introduce

```python
@classmethod
def checkpoint_safe_globals(cls) -> Sequence[Any]:
    """Model-specific types persisted in ``hyper_parameters``.
    """
    return ()
```

Model specific globals live here.
    
Add new plugin for model checkpoint `AnomalibCheckpointIO`

### Single context manager, layered allowlist

[`anomalib.utils.serialization.anomalib_safe_globals`](src/anomalib/utils/serialization.py)
merges three layers for the duration of a load:

```text
ANOMALIB_SAFE_GLOBALS  +  NUMPY_SAFE_GLOBALS  +  extra (optional)
        │                         │                      │
        ▼                         ▼                      ▼
  PrecisionType            numpy scalar/ndarray     model enums
                           (scheduler / opt state)
```

```python
with anomalib_safe_globals(extra=cls.checkpoint_safe_globals()):
    torch.load(..., weights_only=True)  # via Lightning / CheckpointIO
```

Rules for what may be added:

- **Shared anomalib list:** first-party enums reused across models.
- **Numpy list:** inert reconstruction helpers / dtypes required for scheduler and
  optimizer state under NumPy 2.x (`numpy._core`). Both `numpy.core` and
  `numpy._core` pickle path aliases are registered so older pickles still resolve.
  The project’s dependency floor is NumPy ≥ 2 (via `opencv-python-headless`), so
  importing `numpy._core.multiarray` is intentional.
- **Model extras:** small first-party types persisted only by that model’s
  constructor hyperparameters. Not shared third-party types, not modules, not
  `pathlib.Path`.

### Dual load seams

Checkpoint restore happens in two places; both must use the same allowlist policy.

```text
AnomalibModule.load_from_checkpoint
        │
        └─ anomalib_safe_globals(extra=cls.checkpoint_safe_globals())

Engine → AnomalibCheckpointIO.load_checkpoint
        │
        └─ anomalib_safe_globals(extra=self.extra_safe_globals)
              ▲
              │ Engine._ensure_checkpoint_io_plugin(model)
              │ copies model.checkpoint_safe_globals()
```

1. **Direct API / export:** `AnomalibModule.load_from_checkpoint` wraps Lightning’s
   loader with `extra=cls.checkpoint_safe_globals()`.
2. **Trainer `ckpt_path`:** `AnomalibCheckpointIO` wraps `TorchCheckpointIO`. The
   Engine installs it when the user has not supplied a `CheckpointIO`, and refreshes
   `extra_safe_globals` from the active model before fit / validate / test / predict.

Caller code stays unchanged:

```python
model = VlmAd.load_from_checkpoint("model.ckpt")
engine.test(VlmAd(), ckpt_path="model.ckpt", datamodule=datamodule)
```

### Per-model extension point

```python
class AnomalibModule(...):
    @classmethod
    def checkpoint_safe_globals(cls) -> Sequence[Any]:
        return ()

class VlmAd(AnomalibModule):
    @classmethod
    def checkpoint_safe_globals(cls) -> Sequence[Any]:
        return (ModelName,)
```

## ✨ Changes

- [ ] 📚 Documentation update
- [ ] 📦 Build system changes
- [ ] 🚧 CI/CD configuration
- [x] 🔒 Security update

## ✅ Checklist


samet-akcay (approved):

src/anomalib/engine/plugins/checkpoint_io.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, extra_safe_globals: Sequence[Any] | None = None) -> None:
+        super().__init__()
+        self.extra_safe_globals = list(extra_safe_globals or ())
+
+    @override
+    def load_checkpoint(
+        self,
+        path: _PATH,
+        map_location: Callable | None = lambda storage, _loc: storage,
+        weights_only: bool | None = None,
```
rajeshgangireddy, line 38: similar to the copilot's comment. 
I think it might be better to explicitly enforce weights_only=True by default 
here and in 
anomalib/src/anomalib/models/components/base/anomalib_module.py

Lightning might change None to False for HTTP/HTTPS checkpoint URLS? 

https://docs.pytorch.org/docs/main/notes/serialization.html#id3

rajeshgangireddy (commented):
Minor comments. 
- Can you also make sure tiling checkpoints are also loaded okay with this this new changes. `get_ensemble_model()` Tiling loading passes a preprocessor and i feel this might fail 

- Some files might need year update in copyright headers. 
- Do we need a similar weight_only plugin for TorchInferencer as well ? It still uses `torch.load(path, map_location=self.device, weights_only=False)  # nosec B614`

